PR #26, opened 2015-12-12
Add ember-simple-auth

JoshSmith:
This adds `ember-simple-auth` to the project.

We need to:

- [x] Write feature tests for logging in
- [x] Write feature tests for logging out
- [ ] Find a way to wrap API errors and provide more user-friendly errors
- [ ] Write unit tests for error wrapping


JoshSmith:
I added a little bit for us to get started: a barebones login and failure test, set up mirage, `ember-simple-auth`. We need to nail down our style guidelines for these tests now and figure out the right way to do these things early.

JoshSmith:
This also adds ember deploy scripts. We still need to get this working fully on Codeship for uploading all our assets to S3. It uses the lightning deploy strategy.

begedin:
For API error wrapping, some research notes:

Ember-Data expects the typical JSON API error format, but there is a difference in how it works with validation, and other errors.

* If the response code is 422 Unprocessable entity, ember data will initialize a [`DS.InvalidError`](http://emberjs.com/api/data/classes/DS.InvalidError.html).
* For any other code, the initialized error will be [`DS.AdapterError`](http://emberjs.com/api/data/classes/DS.AdapterError.html)

## `DS.InvalidError`

`DS.InvalidError` expects the error data to be in the following format (the listed properties are the minimum it needs to have):

```JavaScript
{
  errors: [
  {
    detail: "Detailed message"
    source: { pointer: "data/adapter/dasherized-field-name" }
  }
  ...
  ]
}

```

If the errors are in this format, ember-data's serialization layer will parse them and the model will have its `model.errors` property set. Note that the `model.isError` will not be set to true in this case. The `model.isError` property is only set to true if, during a `model.save`, the server **returns an error that is not a validation error**.

The default field name format is dasherized, but my guess is, if we override our serializer to expect underscored property names by default, that setting will also apply to errors.

My proposal is that, for validation errors, we do a PR on the API side which changes how we serialize those errors so the format is valid for ember.

Then, we handle such errors by catching a failed save on the model. We will have access to properties such as `model.get('errors.username');` and it will be easy to display them.

Fetching the property in the way I describe will give us back an object:

```JavaScript
[
   {
      "attribute": "username",
      "message": "The username is too long"
    }
]
```

Basically, `detail` is mapped to `message` and `pointer` is mapped (with modification) to `attribute`.

## `DS.AdapterError`

This instance is used for any other error, including model save errors that aren't validation errors. This type of error expects the format:

```JavaScript
errors = [
  {
    title: 'Adapter Error',
    detail: message
  }
]

```

If the API doesn't return a body with an error response, a default error object will be generated, with a single error in it, with `detail` set to `'Adapter operation failed'`.

There are two additional error classes which are sometimes used - `DS.AbortError` and `DS.TimeoutError`. They inherit from `DS.AdapterError` by initializing it with custom messages of their own.

Basically, I don't really think we should handle errors in any custom way.  Ember-data already wraps errors on it's own well enough.

At most, we might want to override the messages ember uses, so we could do a wrapper for that, which checks the error type and returns a different message based on that.

Something like

```JavaScript

function overrideErrorMessage(error)
  if error.typeOf(TimeoutError) { 
    error.set('message', 'Our custom message for timeout error');
  } else if (error.typeOf(AbortError) {
    error.set('message', 'Our custom message for abort error');
  }
}

```

We would likely override the above two, since as far as I can tell, they are raised when the operation fails, but the server doesn't respond with anything.

For an `AdapterError`, the server did respond with something, so we should leave the message the server provided.

PR #31, opened 2015-12-15
Add routes

JoshSmith:


begedin:
_Copied from other PR:_

# Implementation

## Nested API routing

This was and still is a problem.

Ember or ember-data support no such thing out of the box. In order to make the resource URL become 

```
"#{host}/:member_slug/:project_slug
```

I had to resort to basically hacking at it. In the `member.project` route, I use `this.paramsFor('member')` in order to fetch the `member_slug` param. I then pass that in along with `projectSlug` into `this.store.queryRecord('project', query)`. 

The `project` adapter then builds the full url using those parameters.

For the parent, `member` route, it's a bit easier, since we only need the `memberSlug`.

For both the project and the member adapter, we also use the `sortQueryParams` method to clean up the `slug` and `memberSlug` respectively. Otherwise, the adapter will request something like

```
GET /:member_slug?slug=member_slug
GET /:member_slug/:project_slug?slug=project_slug&member_slug=member_slug
```

# Note on routing to users and organizations

I deleted the `user` and `organization` routes. The general idea I had is to just use the `member` route. Based on the underlying type the member's `model` property has, we can either render a `user-details` or an `organization-details` component.

It just feels like the least hacky/simplest way to do it.

begedin:
As far as i can tell, I misunderstood the general layout of the site, so I possibly should not have removed some of the routes.

I think we need to lay them out here, though, so I could use some input on how these would look like. The way I see it, we

* need the `member` route, with the url at `/member_slug`
  * there's a `member.user` route showing detailed user information, if the member is a user
  * same goes for a `member.organization` route
  * ideally, the url for both of these routes should remain `/member_slug`, if that's at all possible
  * there's a `member.projects` route, showing a list of projects for that member, url is `/member_slug/projects`
  * there's a `member.project` route, showing a single project's information, url is `/member_slug/project_slug`
    * each project has a list of posts at `member.posts.index`, url is `/member_slug/project_slug/posts`
    * each post's details are at `member.posts.post`, url is `/member_slug/project_slug/posts/id`

Would that make sense?

How about things like user editing (adding skills, etc.)?
 
Would that go under `member.user`, with extra controls for stuff like that shown if the user is the current user, or would it go under something like `profile.edit`, as in a route of it's own? I'm leaning towards the second choice.

image

From the image above, it looks like we should also have a general `projects.index` route, with the url `projects` listing (and filtering) projects across all organizations.

app/adapters/project.js
```diff
@@ -0,0 +1,41 @@
...
+    query = query || {};
+    if (query.slug && query.memberSlug) {
+      delete query.slug;
+      delete query.memberSlug;
+      return query;
+    } else {
+      return this._super.apply(arguments);
+    }
+  }
+});
```
begedin: It looks like there is code duplication in these two, but I think it's preferable to somehow bunching the code together, due to the way it works. It's similar for the two adapters, but not quite the same.

begedin:
@JoshSmith I added the missing acceptance test, but I had a lot of difficulty getting mirage to work and the result is hacky.

To fetch our record on the member route, we use `store.queryRecord`, which then goes through the adapter's `urlForQueryRecord` to build the url our API expects.

Our API includes the model, which is either a user or an organization with the member record. However, `ember-cli-mirage` does not currently support relationships in its factories at all, so we have to build our response almost completely manually, which is the big and ugly chunk of code in `app/mirage/config.js`

I looked into the alternative of `ember-data-factory-guy`, but that library doesn't support mocking the `store.queryRecord` request, so it's not working nicer. Additionally, `ember-data-factory-guy` completely ignores our custom serializers and adapters, so it's basically useless for us until it adds support for that.

Sadly, until mirage 0.2 is out, it seems we will have to mock our requests this way which is just a short step away from specifying requests manually using pretender, as we did in some of our older projects. It's still a significant improvement, but not as much as it could've been.

package.json
```diff
@@ -28,6 +28,7 @@
     "ember-cli-deploy": "0.5.1",
     "ember-cli-deploy-lightning-pack": "0.2.0",
     "ember-cli-dotenv": "1.0.4",
+    "ember-cli-es5-shim": "^0.1.1",
```
begedin: Needed for mirage. It uses `.bind()` internally, which PhantomJs doesn't support out of the box. This shim adds that support.

PR #32, opened 2015-12-15
Add routes

begedin:
Didn't have read access, so I made a fork.

As far as I remember, this was the original intent with code-corps anyway, so I thought I might as well get used to it.

The problem with this approach, though, is that CI doesn't work on the base repo. The fork needs to have it's own CI set up.

In any case, I got some of the stuff working, but i'll probably end up pushing the changes to `origin/add-routes`, since we already have a PR there.

# Implementation

## Nested API routing

This was and still is a problem.

Ember or ember-data support no such thing out of the box. In order to make the resource URL become 

```
"#{host}/:member_slug/:project_slug
```

I had to resort to basically hacking at it. In the `member.project` route, I use `this.paramsFor('member')` in order to fetch the `member_slug` param. I then pass that in along with `projectSlug` into `this.store.queryRecord('project', query)`. 

The `project` adapter then builds the full url using those parameters.

For the parent, `member` route, it's a bit easier, since we only need the `memberSlug`.

For both the project and the member adapter, we also use the `sortQueryParams` method to clean up the `slug` and `memberSlug` respectively. Otherwise, the adapter will request something like

```
GET /:member_slug?slug=member_slug
GET /:member_slug/:project_slug?slug=project_slug&member_slug=member_slug
```


begedin:
# Note on routing to users and organizations

I deleted the `user` and `organization` routes. The general idea I had is to just use the `member` route. Based on the underlying type the member's `model` property has, we can either render a `user-details` or an `organization-details` component.

It just feels like the least hacky/simplest way to do it.

Issue #22, opened 2015-07-14
Create profile edit page


JoshSmith:
Implement the profile edit page designed in #21.

PR #33, opened 2015-12-16
Profile route

### LINKED ISSUE #20: Create profile page
We need to implement the profile page designed in #19.

begedin:
Closes #20, #22 when done

This PR, in the current state, mostly servers as a proof of concept for using integration tests for components and ember-mirage in conjunction with acceptance tests.

After my experiences today, I'm leaning against using ember-mirage, at least no in the current, stable 0.1 release. At some point in the future, it will be great, but I don't think it is right now, without a lot of overhead we need to handle.

* server.create('type') doesn't create a DS.model, just a plain object payload (something the server would return as a response), so we can't use moel instance directly (for the purposes of unit/integration testing, for example).
* json api isn't supported, so we need to wrap all of our requests in proper formats. this means we can't use any of the verb shorthands.
* request wrapping means we need to basically manually implement a patch method since it needs to
  * find and update the affected record in mirage's database
  * return a proper response
* it's a pain to use mirage outside of acceptance tests in general, since we need to manually load mirage test helpers

A good alternative would be ember-data-factory-guy. It's not without it's issues, but the server requests/responses, at least, are handled automagically.

In my opinion, at a later point in development, mirage will be the way to go, but right now, ember-data-factory-guy, in conjunction with, perhaps, ember-cli-faker, would be the simpler approach.


begedin:
@JoshSmith I'm still working on wrapping this up, but I could use your thoughts on the mirage vs factory guy issue.

begedin:
Just to recap

* Advantages of `ember-cli-mirage`
  * Extremely good once it gets further ahead up the development tree
  * Allows us to use fake data in development in addition to testing

* Disadvantages of `ember-cli-mirage`
  * Doesn't support JSON API in current version (0.1), but will support it in 0.2
    * Requires wrapping requests into proper format as a workaround
    * Makes us unable to use shorthands for route definitions
    * Requires quite a bit of work for POST and PATCH requests (need to create/update specified record manually, relate to other records, etc.)
  * `create` and `createList` commands return attribute hashes instead of instances of `DS.Model`
 
* Advantages of `ember-data-factory-guy`
  * We just use factory syntax similar to ruby `factory-girl`. The actual requests are handled automagically.
  * There are `create` and `build` commands to create either a  `DS.Model` or an attribute hash
  * Easy to define relationships, etc.

* Disadvantages of `ember-data-factory-guy`
  * Harder to simulate error responses. Methods exist, but overall, mirage is simpler and gives more control. We'll probably have to use mockjax directly on occasion. Luckily, factory guy already uses it internally, so we have access to it.

I think our options are
* use factory guy - relatively easy to use now. not much expected to change in the future
* use mirage - more initial overhead, probably just as good in the future (distance uncertain)

begedin:
@JoshSmith This still isn't ready for merge, but it's much better than it was. With the mirage update, mocking the `GET /users/:id` request works just fine.

This comment should serve both as a checklist for myself and as something we can base our discussion on.

First of all, is this a general user's profile page, or the current user's profile page? Right now, I'm approaching it as the current user's profile page, which can be edited if the the current user is the user being viewed, but that behavior isn't really implemented yet.

* [ ] Make it so that the profile is editable only if the current user is the user being viewed. Take into account the possibility of this power being extended to admins by making it some sort of `canEdit` property.


Based on #20 - Profile page should include
* Basic information 
  * [ ] name
  * [x] username
* [ ] Photo
* [ ] Skills - this one probably won't be added in this PR, since we don't have skills yet
* [x] Website
* [x] Twitter
* [x] Biography
* [ ] Contributions - again, probably not added with this PR

Based on #22, we should also add editing capabilities. This is what can currently be edited
* username
* website
* twitter
* biography

I'm not sure skill should be editable from the profile, or a separate page, but that's for a different PR, not this one.

* [ ] Should the username be editable? I'm actually thinking no, at least for now, because I'm not sure if that will complicate things too much on the API side.
* [ ] I still need to add in the option to edit the user photo, once it's shown here, but considering we'll probably need some library for that, it might be a good idea to put it in a separate PR. We can display the photo in this one, but make it editable in another one.

begedin:
# New update

* Profile page is working. It's used for editing the current user. I don't think we should worry about admins being able to edit other users yet, but if we do go with that, a simple check at the route level would suffice to enable this feature. The check would be
  * Is there an id parameter
  * Is the current user an admin
* We can edit website, biography, twitter
* We cannot edit username. I decided it might cause too much trouble at the moment. We can tackle that at a later time if necessary
* We cannot edit user photo, so we do not display it yet either. We need to handle code-corps/code-corps-api#187 first, and then complete this feature in #44
* We do not list skills yet. I'm not sure what the prerequisites are, but the feature can be done in #46 once we figure them out
* We do not list contributions yet. Same as with skills, the associated task is #45

With all that, I believe this is in a mergeable state.

begedin:
@JoshSmith I need a review of this in order to merge.

JoshSmith:
This is not reviewable right now because the API has changed considerably. There is no longer a `Member` but we're relying it on for routing.

JoshSmith:
The user's profile should not be hidden behind authentication.

JoshSmith:
I see, the profile is the user's own view of themselves. This is a little confusing naming-wise.

JoshSmith:
Changing the path to `/settings/profile` and removing the `editing` nature of the component. I think it should always be in editing mode.

Will need to:

- [ ] Fix tests
- [ ] Add a flash message for success

JoshSmith:
The only failing test right now is rollback. Doesn't work for some reason.

JoshSmith:
The only remaining issue is with `ember-cli-flash` and the messages. Seems like an injection issue but can't seem to figure it out.

@begedin if you can solve this, I think this should be mergeable after pending your review.

begedin:
@JoshSmith 

In order to get our test to work, I had to make a few changes, which I'd like to explain

### Made injection of `service:flash-messages` explicit for the component we use it in

The initializer provided by the addon first registers `service:flash-messages` and then automatically injects the service into the list of factories we specify with the `injectionFactories` property in our `config/environment.js`. For us, that was `['component', 'route', 'view', 'controller']`

This is the old Ember way of doing things. The new way is to just register the service and then use `Ember.inject.service()` in places where we need access to it, in order to inject it explicitly in only those places.

```JavaScript
  //...

  flashMessages: Ember.inject.service(),
  
  //...
});
```

### Mocked out `service:flash-messages` in our `profile-details` component integration test

Acceptance tests initialize our entire application, so `Acceptance: Profile` had no issues with flash messages

However, integration tests only start up a container for our component, instead of initializing the entire app, so the initializer provided by the flash messages addon (which registers its service), never runs, and even if we run it manually, it uses features of the application instance which our minimal component container instance does not have.

Instead of that, I decided to define a stub object which mocks our flash messages service and register it as `service: flash-messages` so the component has access to it. This also allows us to make an additional test that the service is being called.

PR #38, opened 2016-01-11
Login form

### LINKED ISSUE #3: Create login form
We need a login form for email/password, add a link for "sign up" to login form, and after signing in, redirect to projects page.

begedin:
Closes #3 

While we already do have login code in the controller, there was a task to create a login-form, so using the signup form PR #34 as a template, I've moved the login form logic into a component of it's own. 

ember will eventually have routeable components, so we should strive to have most things organized this way wherever possible.  For the most part, our controller scripts should be undefined and our route templates should just render a component template, without any other content.

This should make the switch to routable components, once it's out, much easier.


begedin:
@JoshSmith I'm unable to access the repo on codeship, so I can't see why the tests are failing. They all pass on my machine.

JoshSmith:
Looks good! I simply added some styling here. I don't know if routable components will move to pods structure eventually, but putting components under `styles/components` should make it dead simple to do the transition if that happens.

PR #39, opened 2016-01-13
Project page

### LINKED ISSUE #11: Implement project page
Implement the design finished in #10.

begedin:
Resolves #11 

# Further difficulties with ember-cli-mirage

As I already mentioned before, `ember-cli-mirage` does not support relationships at the moment. The documentation states that, for now, we can get around that by specifying the foreign key of the relationship when creating an object (ex, 
```JavaScript
let post = server.create('post', { title: 'Some title', project_id: ˙some id˙});
```

However, this does not work with JSON API, since the format there is different and relationships are a separate object.

Right now, we're hacking our way through GET requests for members and projects due to this issue, but it's getting increasingly difficult to do so and will just become a mess eventually.

On the other hand, the alternative, `ember-data-factory-guy` supports relationships quite well, even polymorphic, but it has another issue.

In order to fetch our record by slug, we use `store.queryRecord` and then override `urlForQueryRecord` inside our adapters, to build urls in the format `/:memberSlug/:projectSlug`. `queryRecord` doesn't look at the store at all and outright fetches the record in question from the API side, basically forcing a reload. That and the fact that the url building for it is separate from other things in the adapter made it seem like a great candidate for fetching members and projects.

The problem is, `ember-data-factory-guy` works in one of two ways.

1. We can call `TestHelper.handleFind/handleFindAll/etc.` to mock a request to the server and provide it data to return In our case, we can't use this approach, since there is no `handleQueryRecord`
2. We can call `FactoryGuy.make/` to create a payload and push it to the store.In that case, a `store.find` will just fetch a record from the store, without making an API request. We can't do this either as it is right now, because `store.queryRecord` always makes an API request.

## My suggestion

After putting a lot of thought into this, and wasting a lot of time trying to work around mirage features that still aren't developed, I think we should do the following.

1. Switch to `ember-data-factory-guy`
2. Switch to use `store.findBy` in our member and project routes
3. Switch our adapters to override `buildURL` instead of `urlForQueryRecord`. `buildURL` will, in the case when the query object has a slug parameter, crate the URL we need (`/:memberSlug/:projectSlug) and just call super otherwise, which is similar to what `urlForQueryRecord` does.

## Optional - Don't rely on including records on the API so much

Starting at 2.0, Ember will default to `async: true` for relationships on models. It looks like this is the recommended behavior to use in general, instead of pulling a bunch of included information via a single request.

We already have most endpoints for fetching records directly anyway, and I really feel it would make things simpler on the ember side if we didn't include things, both in tests and in the actual app.

## Optional - Don't nest routes as much or at all on the API side.

Due to the way the interface is built, nesting makes a lot of sense on the ember side of things. 

However, the default ember adapter behavior doesn't nest at all and instead just uses regular queries for get requests, implying that this way of structuring the API is preferable to nesting.

What is the reason we went with nesting in the first place? It does seem like it makes handling some parameters easier, but if we work with slugs and ids, I don't really think we need it. 

Sure, it fits with the way things are structured in the database, but the routes we use on the API side are never visible to the user, so I don't see an advantage there.

The biggest advantage I'm seeing is with `GET /:member_slug/:project_slug` where it allows us to define the same project slug for different members, but even that can be done with a regular

```
GET /projects?member_slug=x&slug=y
```

Incidentally, this is the URL which ember would build by default for 

```Javascript
store.queryRecord('project', {memberSlug: x, slug: y})
```

Or, assuming ,`member.id` is `1`

```Javascript
store.findBy('project', {member: member, slug: y});
```
Would build a request along the lines of 

```
GET /projects?member_id=1&slug=y
```


JoshSmith:
I think I agree with this assessment completely. Do you want to open some issues on https://github.com/code-corps/code-corps-api to take care of the API issues?

JoshSmith:
After talking to someone about mirage, they recommended we work off master. Have you tried that?

begedin:
I could give master a go, but considering some of the open issues, I'm weary of it.

A quick glance at a documentation tells me master should support JSON API, as well as associations, but I'm not sure how polymorphic associations will work. I'll give it a try and continue with factory-guy if it seems like too much trouble.

Ideally, I would prefer mirage because it really seems like the more feature-complete library, once it's actually feature-complete.

begedin:
@JoshSmith I got things working with the master branch of mirage

It's till not perfect, since even the master branch doesn't support polymorphic relationships, but I managed to sort of monkeypatch those in, so both projects (for polymorphic "owners") and members (for polymorphic "model") now work and we can create records.

Some of the fixes and changes got applied to PR #31, so my suggestion is the following

* Merge #31 
* At that point, the changeset here should be clearer. I can add comments to further clear it up and then merge this as well.
* Once mirage supports polymorphism, we can get rid of our monkeypatches and do it properly. For now, I think it should be good enough.

Note that most of the hacking is within 
* mirage/models/member
* mirage/models/project
* mirage/serializers/member
* mirage/serializers/project

A few examples of how mirage can be used to create models with relations is in 
* tests/acceptance/project-test
* tests/acceptance/member-test

PR #42, opened 2016-01-15
Posts paging

### LINKED ISSUE #41: Implement paging for project post listing

begedin:
Closes #41

Nowhere near done yet, but I need a place to put my findings as I'm researching into this.

Based on what I have right now, it would be extremely difficult to implement paging on a hasMany relationship directly.

Instead, it would be simpler to have the component assigned for post listing to fetch from the /posts API endpoint directly, providing a `projectId`, `postType` and `pageNumber`.


begedin:
I believe i got paging to work, together with filtering by post type, but I need to handle edge cases now, and actually make the interface useful. Right now, I'm just listing post titles.

begedin:
Outside of styling, this can be considered done. However, the changes made in #33 to mirror API changes also need to be applied here.

I think it's faster easier to first merge #33 and then resolve any conflicts here once that's merged, than to apply the same set of changes to this branch/PR.

JoshSmith:
This is technically mergeable, but the tests still reference `member` when the reality is that we need to incorporate organizations and slugged routes instead.

begedin:
I did some further cleanup, so this ought to be mergeable now.

JoshSmith:
You can go ahead and merge this. I'm not sure if the issues I noticed with next, previous buttons are gone, but you should definitely test this yourself to be sure.

Issue #14, opened 2015-07-14
Create add post page


JoshSmith:
We need to implement the page designed in #13.

PR #47, opened 2016-01-19
Add posts

### LINKED ISSUE #16: Create post page
We need to create the post page designed in #15.

JoshSmith:
Closes #16, #14 

- Need to handle publishing posts. Right now a post just goes to published immediately - #52 
- API side we need to not show posts that are in draft mode. Some sort of filtering is needed - code-corps/code-corps-api/issues/196
- [x] Ember app needs to hit posts using their numbering, which means they need to be scoped under a project's slug.
- [x] Fix posts to display html correctly (i.e. we need `{{{post.body}}}` vs `{{post.body}}`. We need a test for this.


begedin:
I made the process work, but there will be conflicts when we merge #33, #42 and #48.

begedin:
Since there's quite a bit of overlap, I also implemented viewing of individual posts ( sans comments) to this PR.

begedin:
@JoshSmith I did what I could here, but I think the proper publishing process should be handled separately. This PR is more for the purposes of getting the infrastructure ready.

I created the tasks that follow and I think this ought to be mergeable, but it will probably have conflicts with other merged PRs. Marking as "awaiting review" right now, with the possibility of some conflict resolution once #42  is merged.

JoshSmith:
Looks good to me.

Issue #49, opened 2016-01-21
Add comments to individual post page


begedin:
Continuation of #16

JoshSmith:
This was closed by #54

PR #54, opened 2016-01-25
Add comments listing and creation

### LINKED ISSUE #17: Create comments form
We need to create the front-end for creating a comment. Markdown preview would be nice but not necessary.

begedin:
Closes #17, #49 

Branches off of #47, so the change set will not make much sense until that is merged.

I'm still figuring out what the proper way of implementing these type of "entity-creation" forms would be. Thinking in terms of data-down-actions-up, which is the new Ember mantra, we should handle saving records at the route level, with the component simply triggering the save action and sending the record as a parameter.

The two different approaches are in the comment and post creation components. Post creation handles saving and error display internally, while comment creation does it explicitly. I think there are arguments for both approaches, but we should eventually decide on one of them. Switching everything to that one approach should not be a huge issue.

Based on various ember-related documents, blog posts and tutorials, the way it's handled in comment creation would be the intended way of doing it.


app/routes/project/posts/new.js
```diff
@@ -20,7 +20,7 @@ export default Ember.Route.extend({
 
   actions: {
     viewPost(post) {
-      this.transitionTo('project.posts.post', post);
+      this.transitionTo('project.posts.post', post.get('number'));
```
begedin: The `project.posts.post` route uses a complex set of models which include the specified post, it's comments and a new, unsaved comment, so we need to pass in the post number instead of an already retrieved post model.

The number is all we need, since we already have the slugged route and project slugs from the parent route params.

app/routes/project/posts/post.js
```diff
@@ -1,17 +1,43 @@
...
+
+    return Ember.RSVP.hash({
+      post: this.store.queryRecord('post', queryParams),
+      user: Ember.isPresent(userId) ? this.store.find('user', userId) : null
+    }).then((result) => {
+      return Ember.RSVP.hash({
+        post: result.post,
+        comment: this.store.createRecord('comment', { post: result.post, user: result.user }),
+        comments: this.store.query('comment', { postId: result.post.id })
+      });
```
begedin: Model is returning an `RSVP.hash` because, based on documentation I'm finding, it should make it the simplest for us to switch to routed components once those are out.

PR #59, opened 2016-01-30
User photo

scarfdestroyer:
Added a section in the `/settings/profile` route that enables you to add a photo via the add on ember-image-drop and view the thumbnail of the image on the slug route for the owner.

**Note**: this only works with a fix that takes the headers out of the base64_photo_data that is located in the `user` model on the API side.


JoshSmith:
There appears to be a lot of weird mirage stuff added in here that probably shouldn't be added.

begedin:
Hey.

So earlier in the development of `ember-cli-mirage` the addon would create a `code-corps/app/mirage` folder, where it would put it's stuff.

In later versions, which we are using now, the folder got moved to a level above, so it's now in `code-corps/mirage`. 

Something happened during your merge, so the old folder wasn't removed, which is why you have this bunch of weird mirage stuff in the change list.

You should be save to outright delete the `code-corps/app/mirage` folder completely.

JoshSmith:
I got `ember-image-drop` working after fiddling around with the server issues we had before. But there are a number of issues with the component that lead me to believe we should just write our own component instead.

This should be relatively straightforward to do. There are some tests there and their component itself that we can use as a starting point. Then there are some CSS changes we can make, and some new UI to think about for it.

I can walk through the new UI changes once we get a spike going for this.

JoshSmith:
Closing in favor of doing this in our own component.

PR #60, opened 2016-01-30
Add 404 pages

JoshSmith:
This adds the following:

- An `error-wrapper` component that wraps errors.
- Template that for now just responds to 404 errors.
- Test for the slugged route
- Component test

Not sure if the way I'm doing the computed properties makes sense. Could stand to add some documentation to say what it's doing (looking through the `errors` array and finding statuses.

We need to handle the general error case, as well, where we don't have a status code returned that matches what we're looking for.

- [x] Handle general case


JoshSmith:
I think we'll do just fine for now with a 404 and a general error case.

I'm using a service that's injected into the component to change the background color of the `html`, which is necessary because of CSS issues.

There's some indirect testing that this gets unset in the acceptance level test. This would be nice to test at the unit level if possible.

@begedin would be good to get your thoughts on the infrastructure here and the choices that were made. I'm sure this could use improvement, but seems like a decent spike.

JoshSmith:
Here's what they look like:

screen shot 2016-01-30 at 6 36 27 pm
screen shot 2016-01-30 at 6 41 59 pm

JoshSmith:
Eventually I would like to see a search bar added to the 404 error page, and a report error field added to the general page.

app/components/error-wrapper.js
```diff
@@ -0,0 +1,50 @@
...
+  background: Ember.inject.service(),
+
+  observeErrorClass: Ember.observer('errorClass', function() {
+    this.updateBackground();
+  }),
+
+  updateBackground: function() {
+    this.set('background.class', this.get('errorClass'));
+    this.get('background').updateBackgroundClass();
+  },
```
begedin: Are you sure we need this. At a glance here, it looks like we could just use 

```JavaScript
classNames: ['error-wrapper', 'center-pseudo'],
classNameBindings:  'errorClass']
```

This will dynamically set the components class to the value of the `errorClass` property, in addition to assigning the static `error-wrapper` and `center-pseudo` classes.

begedin:
Hey. Offered my thoughts. Basically, I think it can be simplified and the service removed completely, but I'm not sure if I'm missing something.

JoshSmith:
@begedin I think you're missing that the service is setting the class on the `html` element. That's why I can't do simple class bindings here.

begedin:
@JoshSmith You're right, I completely missed that part. Looks good in that case :+1:

PR #64, opened 2016-02-03
Fix posts

### LINKED ISSUE #61: Remove nesting for routes
Unlike on the API side, routing on the Ember side should be mostly unnested. Nesting routes in Ember means nesting of UI, which is mostly not what we want. An individual post should not be nested inside the list of posts, for example.

JoshSmith:
This fixes #61.

This is branched off #60. That will be merged, and if it's changed, there will be merge conflicts.

I managed to figure out how to redo the post routing here for a way that makes sense for the UI.

I'm pretty sure comments are broken with working with the API. We need to get in the habit of QAing these things, and also figure out why there are such inconsistencies between the API and our adapters, and how we can prevent those problems in the future.

I still need to add:

- [x] Integration tests for navigating the `project-menu` and checking for the proper `active` statuses
- [x] Get comments working properly

There are some bugs, too, for which we need tests and then fixes:

- [x] Going from `project.posts` => `login` => `project.posts.new` goes from `/code-corps/code-corps/posts` to `/login` to `/undefined/undefined/new`
- [x] You can't `link-to` some of the models with slugs and expect them to work. Try doing a `link-to` with a `project` or an `organization`, for example, and watch the URL also appear `undefined`. Something is happening to mess up the models/params that are passed around, and I just don't know enough architecturally to fix it.
- [x] Until we deal with proper draft/publish process in #63, we should probably set post.state to "published" upon save. In fact, that will likely be the final behavior anyway.
- [x] When posting a new comment, the comment list should be reloaded, or the comment should be pushed to the list so it's displayed


JoshSmith:
Comments were an API problem.

begedin:
@JoshSmith Since we use {{#link-to}} helpers for the project menu, I had to write an acceptance test, instead of an integration test for that.

Acceptance tests load our entire application, while integration tests do not, so the only other option would be to somehow mock the router in our integration test, I believe. I'm not sure how to do that yet, and I didn't have the time today to figure it out. I'll add a task to improve this - #67

As for the general `link-to` with slugs issue, I believe the problem was with the way I specified the dynamic segments in our router. It seems ember is expecting these to be underscored, not camelized, so it couldn't figure out how to render the `href`. I changed the segments so they're underscored, wrote a few tests and it seems to be working now.

Finally for the bug you mentioned. I wasn't able to reproduce it at all. I'm not really sure what you mean by it. Could you elaborate?

JoshSmith:
I should have been more explicit on steps to reproduce the other bug. If you're logged out and go to the `project.posts` route and click `New Post`, then you get the login screen since that's protected. But then the URL is `undefined` where the slug parts are.

JoshSmith:
Also, I see a few test failures here.

begedin:
So regarding link-tos, It wasn't as easy as I expected. The link-to's in the project-item component still don't work properly, and I'm not really sure how to make them work, without hardcoding the dynamic parameters (which is possible, we could do 

```Handlebars
{{link-to 'projects' sluggedRouteSlug projectSlug}}
```

Except we don't have access to sluggedRouteSlug from within the component. We could always go with an action that propagates up to the route, where a transition is triggered, since the information is available there, but I'm not sure about that either, since we'd lose the href attribute altogether in that case.

The remaining links seem to be working, though.

tests/acceptance/projects-test.js
```diff
@@ -32,5 +32,8 @@ test('It renders all the required ui elements', (assert) => {
   andThen(function() {
     assert.equal(find('.project-list').length, 1, 'project-list component is rendered');
     assert.equal(find('.project-list .project-item').length, 5, 'correct number of project-items is rendered');
+
+    let firstProjectHref = find('.project-list .project-item:eq(0) a').attr('href');
+    assert.ok(firstProjectHref.indexOf('/test_organization/test_project_0') > -1, 'The link to a project is properly rendered');
```
begedin: This is the assertion that's failing, since I didn't manage to fix the link-to issue here yet. If we end up going in a different direction, feel free to remove it.

begedin:
@JoshSmith I got the final slug link to work with minimal issues, by making it a 

```Handlebars
{{link-to 'projects' project.organization.slug project.slug'}}
```

begedin:
@JoshSmith I believe I fixed it. The trick was in the newly added `serialize` method in `routes/project.js`. This method is used to extract query parameters from the model, if it is available. Apparently, the default method was not able to extract them.

It took me A LOT of debugging to figure it out, but I think this will help us greatly in the future.

JoshSmith:
I have borked the tests by changing HTML structure. Not sure.

PR #71, opened 2016-02-11
Replace select view with components - closes #55

### LINKED ISSUE #55: Replace `{{view 'select'}}` with proper component
It's deprecated, so we really ought to replace it ASAP.

We could use a general third party component, or something specific for our needs, depending on judgement.

hbrysiewicz:
Use a select-dropdown and select-option component to render a select instead of deprecated view select


JoshSmith:
Would you mind writing some quick tests for these components?

JoshSmith:
This looks great, btw. :+1:

JoshSmith:
@hbrysiewicz would you be able to merge `develop` in and repush? Also, would you mind doing a `rebase` off of `develop` and squash your commits?

PR #72, opened 2016-02-12
Don't show next page control when no pages - closes #65

### LINKED ISSUE #65: Don't show page controls if there are <= 1
If the number of pages is <= 1, we should not be showing the page controls at all.

hbrysiewicz:


tests/integration/components/pager-control-test.js
```diff
@@ -123,6 +123,19 @@ test('If we are currently on the first page, the previous page button does not r
   assert.equal(this.$('.previous-page').length, 0, 'The button does not render');
 });
 
+test('If we are currently on the first page, the next page button does not render', function(assert) {
+  this.set('options', {
+    pageSize: 5,
+    totalRecords: 0,
+    currentPage: 1,
+    totalPages: 0
+  });
```
JoshSmith: I think this needs to be read:

> If we are currently on the first and only page, the next page button does not render

We could probably also remove all the:

> If we are currently

from all these.

JoshSmith:
:+1: this is great.

JoshSmith:
@hbrysiewicz would you be able to merge `develop` in and repush? Also, would you mind doing a `rebase` off of `develop` and squash your commits?

PR #78, opened 2016-02-16
Replace select view with components

hbrysiewicz:


JoshSmith:
@begedin one of us should help look into this failing test when we have a chance.

PR #91, opened 2016-02-18
Upgraded to ember-cli 2.3.0

### LINKED ISSUE #88: Upgrade to ember-cli 2.3.0
We can finally upgrade to `ember-cli` 2.3.0! :balloon: 

Let's make sure to follow [the changelog](http://ember-cli.com/#changelog) and make this happen!

begedin:
Closes #88 

# Notes

* Changes in flow caused project post list not to work anymore. Since I wanted to make that one a bit more "data-down-action-up", I decided to do it know. I think the new approach is cleaner, less hacky much simpler. It's still doesn't rely on the `queryParams` feature a route would provide, but it's closer to that now than it previously was.
* There were some namespace changes, so I had to change how we inject stub services into our component integration tests.
* There was a change in how adapter errors affect tests, so I had to add an action to the application route to handle errors.
* There was a change in the adapters `sortQueryParameters` hierarchy, causing our `_super` calls to throw errors. I outright removed those calls, since we don't really have a use-case where the `_super` is needed, I outright removed it to simplify the logic.
* I also updated `ember-simple-auth` to `1.1.0-beta.1` thinking the previous version was causing an issue with a test. The addon wasn't the cause of the issue, but I decided to keep the update, since it doesn't break anything.

Overall, this felt like a relatively clean upgrade. Some tests did start to fail, but it forced me to rewrite a few things in a nicer way. I'm happy with the result.


begedin:
Is it just me, or are specs way faster on CircleCI?

PR #96, opened 2016-02-19
User mentions for posts and comments

### LINKED ISSUE #23: Change user mentions in a post to their URL
We'll want to borrow heavily from this: https://github.com/twitter/twitter-text/blob/master/js/twitter-text.js#L643

Closed by #93.

### LINKED ISSUE #92: Render user mentions in posts
Also see #23

Closed by #96.

### LINKED ISSUE #93: Render user mentions in comments
Also see #23

begedin:
Closes #92. Closes #93. Closes #23.

Still a work in progress, but I'm liking how it's progressing, so wanted to put something up for discussion/review.

## Here's what we have

* A `mention-parser` library. It takes a body string and a set of user mentions, replacing each mention with a link to the user.
* An `array` transform, so we can declare the `indices` property on a user-mention model as `DS.attr('array')`
* A `post-user-mention` model (`comment-user-mention` is on the way)
* Specs that test the parser and the transform (they work!).

## Here's what we need

* [x] A `comment-user-mention` model
  * We can probably make a base model which our two models inherit, but I don't think there's a need for that yet.

* The following components should display the parsed body instead of the one we get from the API
  * [x] any component which internally uses the `editor-with-preview` component
    * [x] `create-comment-form`
    * [x] `comment-item`
    * [x] `post-details`
    * [x] `post-new-form`
  * [x] Additionally, two components display the post body directly, editor not included
    * [x] `post-details`
    * [x] `comment-item`

On the API side, I believe all the endpoints which deal with these components should be returning mentions, included in the response.


begedin:
Some notes: 

* The `editor-with-preview` component now yields to a block. This allows us to keep the component completely agnostic of it's environment, It doesn't care if it's being used for post or comment editing. The parent component actually provides the preview, the editor simply displays it when in preview mode.

* `postUserMention` and `commentUserMention` could both inherit from a base `userMention` model, but I don't really see a benefit in it, other than...

* the base `userMention` model we could hypothetically create could define computed properties for `bodyWithMentions` and `bodyPreviewWithMentions`. We could then use these properties everywhere. It would increase code reuse, but I'm still not completely sold on the idea.

* `mirage/config.js` is getting bloated. it will reduce in complexity as mirage grows up, but right now, I'm getting to a point where I'm thinking about writing tests for it. I think I should extract the various handlers into separate files for each route to simplify things. As mirage gets better, those files will mostly go away.

* I've updated mirage to beta.5. Now, `server.create` (which is the only create method that uses factories), returns a mirage-model instance. This allows us to simplify the data-setup for our acceptance tests a bit. I think I can get it to a point where I can define a couple of different setups and extract them into methods or even helper scripts. I'm not sure if I should do it here or as a separate PR, but my vote is a separate PR. As with route  handlers, these will get simpler and mostly disappear with time, I hope.

JoshSmith:
I love that you extracted the generalized `editor-with-preview` to yield. Great work, and also helps clean up design around it.

Hold off on base `userMention` for now. We'll learn over time if it's something that's worthwhile. For now, it'd be over-optimization.

I was thinking the same thing about writing tests for the `mirage/config`. Let's do it as a separate PR/PRs.

Also handle refactoring into methods/helper scripts as a separate PR/PRs.

JoshSmith:
This is pretty great. I have no real feedback on anything I think you can improve, other than to say that I think we need some better documentation.

I believe that Ember uses YUIdoc. Not positive. Here's how that's written: http://yui.github.io/yuidoc/syntax/index.html

It'd be worthwhile to look into this. It can be done in a separate PR, I think, but we should open some issues for it now while we can. The sooner we can document this the better, as it's still fresh.

JoshSmith:
Just pulled this down and tested it. Found a number of bugs or unexpected behavior.

Mentions do not render when previewing markdown.

It looks like if you change text from a user mention to something else, that doesn't reflect unless you reload the page. I'm guessing because there are no mentions to parse? Not sure.

I get a server error page when posting comment with mentions. Warning I get when doing so:

> Encountered a resource object with type "comment_user_mentions", but no model was found for model name "comment-user-mention"

JoshSmith:
For some reason I didn't have latest. One minute while I retry _all_ of that with the changes.

app/models/post.js
```diff
@@ -19,4 +19,6 @@ export default DS.Model.extend({
   project: DS.belongsTo('project', { async: true }),
   user: DS.belongsTo('user', { async: true }),
   comments: DS.hasMany('comment', { async: true }),
+
+  postUserMentions: DS.hasMany('post-user-mention', { asnyc: false })
```
JoshSmith: `asnyc` should be `async`. Also, should this be `true` and not `false`? I know that's not how it is in API now, but should it be?

JoshSmith:
Unrelated to this PR, but I notice usernames are no longer showing up on posts in the post list, unless I click on them. Odd behavior.

app/models/post.js
begedin: I was thinking those specifically make sense to be async, since we're never interested in them outside of their parent (comment or post)

JoshSmith: You mean make sense not to be async?

begedin: Yeah, exactly. Sorry, it's still early :)

begedin:
Regarding usernames not showing up, that also might be an issue with `post.user` being async.

JoshSmith:
Possibly. Make an issue for it?

begedin:
Looking at some of the issues you pointed out, it looks like the API has some problems still. For instance, it doesn't include mentions when responding for a preview, somehow. Possibly. it saves them after responding. I'll look into it.

begedin:
More odd behavior. With each response for a post, new user mentions don't override old ones, so they just get added to the relationship. I need to figure this out ember-side as well.

begedin:
Changes look good, but looking at this right now, is there much of a point to keep `user_serializer` and  `authenticated_user_serializer`, especially considering how Ember behaves.

The differences are that `user_serializer` has `facebook_id` and `facebook_access_token`, while `authenticated_user_serializer` does not.

JoshSmith:
The `post title resets the input element when editing is cancelled and then restarted` test is really finnicky for me. Not sure how I should be handling it.

begedin:
@JoshSmith I'm trying to figure this out but with no luck. The fact that the test is failing 50% of the time exactly (couldn't get it to happen any other way), makes me think there's something more to it.

JoshSmith:
Yeah, this is pretty absurd. I'm fairly certain we're switching between parallel universes with each test run. Not realistically reproducible.

PR #97, opened 2016-02-19
Add users organizations to the profile, no tests yet

### LINKED ISSUE #90: Add tests for profile
There are no acceptance for profile. We should write some.

JoshSmith:
This doesn't have tests.

It needs:

- [x] Tests for user sidebar component
- [x] Tests for user organizations list component
- [x] Acceptance test for user's profile (includes clicking all links)

Fixes #90.


begedin:
@JoshSmith The fix is up. I had to do a customization of the mirage serializer for slugged routes, so it "supports" polymorphic relationships. We had something similar back when projects could be owned by users or organizations.

https://app.asana.com/0/92371261454978/92995962047391

The most important part of the fix is in `mirage/serializers/slugged-route.js`

PR #109, opened 2016-03-11
Add image uploading for users, organizations, projects

### LINKED ISSUE #44: Add photo to user profile
* It should be viewed
* It should be editable
* The API will accept base_64_photo_data, so there is no need to upload. It should suffice to read the photo on the client side and send the bae64 data - blocked by code-corps/code-corps-api#187
* Might be a good idea to limit file size.

JoshSmith:
Fixes #44.

I created two components. One is a `drag-zone` component that actually wraps the `app`. And it has a `drag-state` service, which I inject into the `image-drop` component and bind to the `dragging` property on the service.

You could have any number of components that react to the service.

Here's a demo: http://cl.ly/2Z1J401z1V1g


JoshSmith:
Added component and acceptance tests and is ready for review.

app/services/drag-state.js
```diff
@@ -0,0 +1,13 @@
...
+  isDragging: false,
+
+  dragging() {
+    this.set('isDragging', true);
+  },
+
+  leaving() {
+    this.set('isDragging', false);
+  }
+});
```
begedin: Now we're thinking in services :)

begedin:
Looks good! :+1:

Issue #29, opened 2015-12-14
Handle errors better


JoshSmith:
For API error wrapping, some research notes:

Ember-Data expects the typical JSON API error format, but there is a difference in how it works with validation, and other errors.
- If the response code is 422 Unprocessable entity, ember data will initialize a [`DS.InvalidError`](http://emberjs.com/api/data/classes/DS.InvalidError.html).
- For any other code, the initialized error will be [`DS.AdapterError`](http://emberjs.com/api/data/classes/DS.AdapterError.html)
## `DS.InvalidError`

`DS.InvalidError` expects the error data to be in the following format (the listed properties are the minimum it needs to have):

``` JavaScript
{
  errors: [
  {
    detail: "Detailed message"
    source: { pointer: "data/attributes/dasherized-field-name" }
  }
  ...
  ]
}

```

If the errors are in this format, ember-data's serialization layer will parse them and the model will have its `model.errors` property set. Note that the `model.isError` will not be set to true in this case. The `model.isError` property is only set to true if, during a `model.save`, the server **returns an error that is not a validation error**.

The default field name format is dasherized, but my guess is, if we override our serializer to expect underscored property names by default, that setting will also apply to errors.

My proposal is that, for validation errors, we do a PR on the API side which changes how we serialize those errors so the format is valid for ember.

Then, we handle such errors by catching a failed save on the model. We will have access to properties such as `model.get('errors.username');` and it will be easy to display them.

Fetching the property in the way I describe will give us back an object:

``` JavaScript
[
   {
      "attribute": "username",
      "message": "The username is too long"
    }
]
```

Basically, `detail` is mapped to `message` and `pointer` is mapped (with modification) to `attribute`.
## `DS.AdapterError`

This instance is used for any other error, including model save errors that aren't validation errors. This type of error expects the format:

``` JavaScript
errors = [
  {
    title: 'Adapter Error',
    detail: message
  }
]

```

If the API doesn't return a body with an error response, a default error object will be generated, with a single error in it, with `detail` set to `'Adapter operation failed'`.

There are two additional error classes which are sometimes used - `DS.AbortError` and `DS.TimeoutError`. They inherit from `DS.AdapterError` by initializing it with custom messages of their own.

Basically, I don't really think we should handle errors in any custom way.  Ember-data already wraps errors on it's own well enough.

At most, we might want to override the messages ember uses, so we could do a wrapper for that, which checks the error type and returns a different message based on that.

Something like

``` JavaScript

[...]

Issue #28, opened 2015-12-14
Add a better error message for invalid grant


JoshSmith:
This may need to be done on the API but right now the error messages are very user-unfriendly.

JoshSmith:
Closing this because it has been addressed in the new Phoenix API. 🎈

PR #110, opened 2016-03-11
Add error-formatter component to parse and display adapter errors

### LINKED ISSUE #85: Friendlier errors on post creation
Post creation without a title gives you something like:

> The adapter rejected the commit because it was invalid

Ideally we should try to give the user a better error.

begedin:
This will close #85 and possibly #28 and #29

The phantomjs  installation ember-cli includes in its locally installed npm package somehow got corrupted on my end, causing the browser to crash while running tests.

To get around it, I deleted the `node_modules` folder and re-ran `npm install`. I'm guessing some components got updated this way, due to how we specify our versions, so as a result, the default message an `AdapterError` contains changed, causing tests to fail.

Since we already talked about formatting errors such as that one in a nicer way, I decided to quickly implement an error formatter component, to use in place of these basic errors.

An adapter error looks like something along the lines of

```Javascript
errors: [
  // contains errors returned by the API. In case of validation errors, there will be multiple
  // in case of a generic error, there will usually be just one object
  { 
    id: 'ERROR_ID'
    title: 'Error title',
    detail: 'error detail', },
    status: 404 // status code,
    // this last one is only for validation errors
    source: { pointer: 'data/attributes/field_with_validation_error' }
  }
],
message: 'Some sort of standard error message goes here'
// there are other properties here to, but of no concern to us at the moment
```

For each error in the errors array, the component will output a 

```Handlebars
<p class="error">{{error.title}}: {{error.message}}</p>
```

I think that's sufficient for now, but we will be able to modify this behavior as we please in the future.

### DIFF
--- a/app/components/error-formatter.js
+++ b/app/components/error-formatter.js
@@ -1,0 +1,11 @@
+import Ember from 'ember';
+
+export default Ember.Component.extend({
+  classNames: ['error-formatter'],
+  messages: Ember.computed('error.errors', function() {
+    return (this.get('error.errors') || []).map((e) => {
+      return `${e.title}: ${e.detail}`;
+    });
+  }),
+  defaultMessage: 'An unexpected error has occured'
+});
--- a/app/templates/components/error-formatter.hbs
+++ b/app/templates/components/error-formatter.hbs
@@ -1,0 +1,5 @@
+{{#each messages as |message|}}
+  <p class="error">{{message}}</p>
+{{else}}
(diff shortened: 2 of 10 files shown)


begedin:
@JoshSmith The tests actually failed on CircleCI as well, with the same PhantomJS crashing issue. I had to "rebuild without cache" to get them to run.

We may have a different problem here, but I'm not sure what it could be.

PR #111, opened 2016-03-11
Query params

### LINKED ISSUE #68: Make post filters use url/query parameters
When we apply a filter, it should be reflected in the URL, ex:

* `/posts?type=task`

begedin:
Closes #68 

It works!

My aim was to use the controller to achieve this as little as possible, but I couldn't completely avoid it, reason being, the query parameters only get removed from the URL, when they are set back to the initial values defined on the controller. This means that when `page` is 1, it will not be shown in the url, same going for `type=null`.

I also had to add a type filter for "all" which resets the type to null in order to show all post types. I'm not sure how we missed that. I don't have an appropriate icon for something like that, so I didn't style it, @JoshSmith . I'm hoping you would be able to. - #113

Specs have been added to "Acceptance: Project" and everything seems to be working fine. It's not a completely standard approach, but I couldn't figure out a better way. The bit stuff is done in the post adapter, so I explained everything in detail there, using comments.


begedin:
I was gonna fix this today, but was unable to. I'm almost done, though. Here's a quick recap:

I replaced all page/filter switching actions with `{{#link-to}}`helpers, so we have links with proper `href` attributes. This **completely eliminates the need for the `page-button` component** so I just removed it.

Everything works, but several integration tests for `pager-control` component are failing, for a simple reason - integration tests do not start the whole app, so the router isn't loaded properly, so all link-to's don't have their `href` attribute set at all.

I looked for options to enable this with no luck, so my proposal is to get rid of the tests that fail completely, and write acceptance tests instead, to cover this behavior. I'll quickly do that tomorrow.

begedin:
@JoshSmith I got this working, so I feel it's in a mergeable state now. I added a task to style the "All" filter button separately - #113

begedin:
Note that we should merge #110 first, since this one branches off of it.

begedin:
@JoshSmith The phantomjs crashing issue started occurring again, and this time, deleting `node_modules` and rerunning `npm install` did not help. Some googling led me to think the issue is with ember/ember-cli 2.3.x, so I upgraded to 2.4.2. I also preferred some bower/npm cleanup. Turns out, we were using addons that are part of the old ember-cli addon set, but should not be used anymore.

Hopefully, this fixed it, but I won't close issue #112 until I'm sure.

JoshSmith:
This all looks great! I'm going to go ahead and merge now.

Issue #112, opened 2016-03-17
Builds sometimes start failing for no reason. "Rebuild without cache" fixes them.


begedin:
I'm not sure what's going on, but it started happening to me locally first.

Runin `ember t` would cause phantomjs to crash after the 7th or 8th test module (not sure which at the moment). Deleting `node_modules` and rerunning `npm install` fixes it locally. Pushing to the remote then causes `Circle CI` to fail for the same reason, so a `Rebuild without cache` is needed. After that, everything starts working again.

begedin:
After a bunch of digging, #119 seems to fix this, but I'll wait for a merge and revisit later to double check.

JoshSmith:
@begedin is this fixed? I haven't seen it.

JoshSmith:
@begedin bump.

begedin:
@JoshSmith Last I checked, the problem was us using PhantomJS 2.0 +. I believe we're down to 1.9.something, so it should be good now.

JoshSmith:
Closing then.

PR #119, opened 2016-05-17
Update to ember 2.5.0

begedin:
I had issues with #112 again, so this was an attempt to fix them.

My initial idea was that maybe 2.4.2 uses some sort of weird dependency combination that was causing an issue, so I preformed an update to 2.5.0. 

No good, PhantomJS was still crashing locally for me.

I also tried updating phantomjs, as well as switching to phantomjs-prebuilt, with no luck.

It turns out, the solution is to downgrade phantomjs to 1.9.20.

I'll update the issue, but since I did the work of upgrading here, I decided to make a PR.

There are no deprecations I could find. One test was failing, but I fixed that one. There was also a `{{log}}` helper in the application error template which was outputing a bunch of stuff into the console while running tests so I removed it.


JoshSmith:
👍

PR #120, opened 2016-05-17
Allow users to join projects

### LINKED ISSUE #12: Add ability to join a project
You should be able to click a button in a project to join it. We should flash success (and maybe some sort of popover to really drive home the thanks and give some orientation), and change the button.

### LINKED ISSUE #40: Find a way to get user's permissions for a project
This is problematic because only certain types of users can create certain types of posts. And we'll need to show the types in a dropdown.

We can't show the same list to everyone, since some people can only submit issues, and some people can submit anything.

begedin:
Will close #12

Closes #40.

Working on making all the tests pass at the moment. List of things that need doing but aren't being done yet
* [ ] There is `ember-can` code related to user being able to join organization in the `organization-header` component. There are no tests for this yet, and the `#if` doesn't wrap any additional template code, so it technically does nothing. My guess is, there is supposed to be a join button there.
* [ ] Some of the stuff in templates should be replaced with `ember-can`


app/components/organization-settings.js
```diff
@@ -0,0 +1,21 @@
...
+    }
+  }),
+
+  findMembership: function(item, index, enumerable) {
+    let itemMemberId = item.belongsTo('member').id()
+    let itemOrganizationId = item.belongsTo('organization').id()
+    let currentUserId = this.get('session.currentUser.id');
+    let organizationId = this.get('organization.id');
+    return (itemMemberId === currentUserId) && (itemOrganizationId === organizationId);
+  },
```
JoshSmith: I'm repeating myself a lot with this and would love to find a better way to do this. It seems like the general idea here should be possible to extract out, though not sure how yet. Had enough trouble just getting it to work by tying it with yarn.

app/components/user-dropdown.js
```diff
@@ -3,8 +3,6 @@ import Ember from 'ember';
 export default Ember.Component.extend({
   classNames: ['user-dropdown'],
 
-  session: Ember.inject.service(),
```
JoshSmith: I'm injecting now in an initializer.

app/serializers/application.js
```diff
@@ -5,4 +5,7 @@ export default DS.JSONAPISerializer.extend({
   keyForAttribute(attr) {
     return Ember.String.underscore(attr);
   },
+  keyForRelationship: function(key) {
+    return Ember.String.underscore(key);
+  },
```
JoshSmith: We weren't serializing relationships correctly. Needs tested.

app/models/organization.js
```diff
@@ -11,4 +11,5 @@ export default Owner.extend({
   iconLargeUrl: attr(),
   projects: DS.hasMany('project', { async: true }),
   members: DS.hasMany('user', { async: true }),
+  organizationMemberships: DS.hasMany('organization-membership', { async: true }),
```
JoshSmith: We have to explicitly do this on _all_ relationships like this (where we need extra information on the join) or weirdness will arise.

app/templates/components/project-details.hbs
```diff
@@ -1,8 +1,34 @@
...
-<div class="details-container">
-  <h2 class="title">{{#link-to 'project.index' project}}{{project.title}}{{/link-to}}</h2>
-  <p class="description">{{project.description}}</p>
-  <p class="organization">{{#link-to 'slugged-route' project.organization.slug}}<img class="icon tiny description-icon" src={{project.organization.iconThumbUrl}} />{{project.organization.name}}{{/link-to}}</p>
+<div class="more-details">
+  <div class="right">
+  {{#unless project.organization.isPending}}
+    {{#if session.isAuthenticated}}
+      {{#if membership}}
+        {{#if membership.isPending}}
```
JoshSmith: This syntax is a little confusing since `isPending` is an Ember thing. ¯\_(ツ)_/¯

JoshSmith:
The _behavior_ here for the button is generally how I want it, I think.

The details of the session, etc. I could really use the feedback on. I'm kind of shooting from the hip here and it is probably not suitable beyond being a first pass.

Also, as usual, I broke stuff and didn't clean up the mess I left behind. So there's that.

app/components/user-dropdown.js
begedin: I think that's actually something ember wants to walk away from and the preferred approach is to explicitly inject services this way, not through initializers.

app/serializers/application.js
begedin: Yeah, this is something we already had with attributes, as you can see. The default for rails is underscore, while the default for ember-data is a dash, I believe. It may have changed recently on one side, though, so I'm not sure anymore. Maybe we should look into it?

app/models/organization.js
begedin: Do you mean the `async` part? Isn't `true` the default for that?

I agree with going the async approach in most cases either way. While there are more required endpoints on the API that way, as well as more network requests, I believe it simplifies things on both ends overall.

app/templates/components/project-details.hbs
begedin: That's going to be fun :)

app/components/organization-profile.js
```diff
@@ -2,4 +2,21 @@ import Ember from 'ember';
...
+    }
+  }),
+
+  findMembership: function(item) {
+    let itemMemberId = item.belongsTo('member').id();
+    let itemOrganizationId = item.belongsTo('organization').id();
+    let currentUserId = this.get('session.currentUser.id');
+    let organizationId = this.get('organization.id');
+    return (itemMemberId === currentUserId) && (itemOrganizationId === organizationId);
+  },
```
begedin: We now at least know what we need. As you said, good for the first pass, but next, my idea is to have some sort of credential service. When we're working on any route that deals with an organization, we can provide the current user and the organization to that service, and then the service can figure these things out. We would have it all in one place that way.

app/serializers/application.js
JoshSmith: I think that's still the case on both sides, but would be nice to see if there's an option for it on the JSON API side.

app/models/organization.js
JoshSmith: No, I mean that we need the explicit `organization-membership` everywhere, even if it might not necessarily be used in that context. Omitting a relationship in one place can cause a lot of weirdness elsewhere in the app (objects missing when you think they should be accessible).

app/templates/components/project-details.hbs
JoshSmith: I don't know if there's a better name here. Can you think of something?

app/components/organization-profile.js
JoshSmith: @begedin At some point we're going to move to project-based permissioning, too, so things will only grow more and more complex. It will be nice to consolidate that logic. As you can see, there's a lot of repetition in my approach.

Can we have the second pass happen in the context of this PR? I don't really want to merge this as-is (even if I get the tests passing). I'd like for you to take point on it and I can work on some other things.

begedin:
@JoshSmith I got  second pass up, but this will need a third pass to make it look nice.

Right now, the logic is in a service (name is up for debate, as are most other things, really), an is being used by the `project-details` component. I will clean it up and restore the functionality of being taken to signup when the button is clicked and the user is signed in as soon as I can. 

The added functionality of storing credentials for multiple organizations (for lists) can be added to this service relatively easily.

JoshSmith:
@begedin this is really fantastic. Better architecture than I was coming up with, although similar. I like that you found a way to set the organization in the `afterModel` hook. That was the "a-ha" moment for me here that I was missing.

I wonder if the name shouldn't be more indicative, like this is really an `organizationMemberships` service. It's a bit harder on the tongue, but I can imagine a similar `userSkills` service, or `userRoles` service, etc.

app/serializers/application.js
begedin: Switching it on the API side would be problematic, due to the way we write our specs and code in general. `:some_symbol` or `object.some_property` work fine in Ruby. `:some-symbol` and `object.some-property` not so much.

JoshSmith:
I’m not sure it’s worth the effort to test the `credentials` service directly, but rather to let the acceptance tests test it indrectly. I gave up on `user-categories` unit testing for the same reason. We can circle back later if we like, but right now it seems like a rabbit hole that provides us little in the way of additional security blanket.

JoshSmith:
This is basically ready. Not super satisfied that `credentials.userMembershipIsPending` _and_ `can "join organization"` both are used in templates. Can't figure out a way to rethink it as a `can` in any reasonable way.

JoshSmith:
The `credentials` service is not tested correctly and not sure I'm a huge fan of the name. Looked at this PR too long, though, so should be re-reviewed by you @begedin and we can decide what's next.

app/abilities/organization.js
```diff
@@ -0,0 +1,17 @@
...
+
+export default Ability.extend({
+  isAtLeastAdmin: Ember.computed.or('membership.isAdmin', 'membership.isOwner'),
+  userCanJoinOrganization: Ember.computed.empty('membership'),
+  userCanLeaveOrganization: Ember.computed.or('membership.isContributor', 'membership.isAdmin'),
+  userIsMemberInOrganization: Ember.computed.notEmpty('membership'),
+
+  canJoin: Ember.computed('membership', 'membership.role', 'model', function() {
+    return this.get('userCanJoinOrganization');
+  }),
```
begedin: This should probably just be an alias, as in `canJoin: Ember.computed.alias('userCanJoinOrganization')`. If I'm understanding my ember correctly, everything should be have the same.

package.json
```diff
@@ -39,7 +40,7 @@
     "ember-cli-ic-ajax": "0.2.4",
     "ember-cli-inject-live-reload": "^1.4.0",
     "ember-cli-jshint": "^1.0.0",
-    "ember-cli-mirage": "0.2.0",
+    "ember-cli-mirage": "^0.2.0",
```
begedin: If we want to merge this, we will have to link to my PR for now. Otherwise, tests will fail.

tests/acceptance/organization-test.js
```diff
@@ -20,23 +20,24 @@ test("it displays the organization's details", (assert) => {
...
   }, 'Organization');
+
   sluggedRoute.save();
 
   for (let i = 1; i <= 3; i++) {
-    organization.createProject({
-      slug: `test_project_${i}`,
-      title: `Test project ${i}`
-    });
+    server.schema.projects.create({ organization: organization });
```
begedin: Just an FYI, for future needs, `server.schema.modelName.create` doesn't use factories, while `server.create('model-name, attrs)` does, in case we want some data to be auto-generated.

begedin:
Looks good and I'm OK with merging, but we do need to change the mirage version, in that case, so it temporarily uses my PR until that, itself, is merged.

Also, there's that `Ember.computed.alias` think we may want to look into. Good to merge after that 👍

JoshSmith:
Marking as blocked because we need `ember-cli-mirage` to update.

begedin:
@JoshSmith I did what we discussed and basically monkeypatched the `hasMany` behavior into our own, customized `JsonApiSerializer`. Once samselikoff/ember-cli-mirage#758 is removed, we can just have our `organization` and `user` serializers extend from the `MirageApplicationSerializer` instead of `ManyToManySerializer` and that should be it.

The tests pass, so this should be good to review, and hopefully merge. Once merged, I would create an issue to go through the app and switch to `ember-can` where applicable, since this PR adds that into the app, I believe.

begedin:
I was celebrating too early. There's something wrong with our dependencies, so it won't build at all right now.

begedin:
Build is fixed. Either we accidentaly removed the predenter bower dependency (required by mirage), or mirage started requiring it in the latest version.

JoshSmith:
👍

PR #124, opened 2016-05-21
Add project grid and categories

JoshSmith:
Still more to do here. At minimum:

- [x] Write test for categories list component
- [x] Write test for category item component
- [x] Write test for project categories list component
- [x] Write test for project category item component
- [x] Write test for skill list item component
- [x] Write test for project grid item skills list component
- [x] Write test for project grid item component


begedin:
What I'm seeing right now looks pretty good.

What I would possibly try and do differently is the following:

1. Make the visibility of the "show more" button a computed property on the component, which returns true if any of the children are obscured, instead of manually triggering an action from the individual children.

2. If that doesn't work due to the data-down-actions-up behavior (I'm really not sure if it would right now), then try and make use of the mut, or some other of the available helpers, since that seems like the more conventional way to do it.

If neither works, though,your implementation also seems fine.

JoshSmith:
@begedin I'm not sure exactly how you would do the first approach since only the children know if they are obscured. There's no good way for the parent to do it.

And on your second point, I don't know about these helpers. Would be useful to see some pseudocode or something when you have a minute.

begedin:
I'm not sure I understand these new helpers correctly either, but I'm thinking the following.

http://emberjs.com/api/classes/Ember.Templates.helpers.html

All children, once they figure out they ought to be obscured, will remain in that state for good. A `mut` helper tells the template, "when this action triggers, mutate the following property of the component". As soon as any child is obscured, the parent should show the "expand" graphic, meaning "showExpand" (or whatever) should be true from that point on.

```Javascript
// components/child.js

didRender() {
 +    let parentBottom = this.$().parent()[0].getBoundingClientRect().bottom;
 +    let elementBottom = this.$()[0].getBoundingClientRect().bottom;
 +
 +    if (elementBottom > parentBottom) {
 +      this.sendAction('hide', true);
 +    }
 +  }
```

```Handlebars
parent.hbs

{{child hide=(mut 'showExpander')}}
```

Instead of having an action handler in the component, we just have a `showExpander` property which starts out false and becomes true.

In hindsight, I don't think this is really an improvement. I just don't like the idea of actions being misused (which they kind of are, as events, here). Not sure there's much we can do about it, though.

The computed property approach would be as follows:

* The parent wraps the children data, giving them each an `obscured` property
* In the parent template, instead of the basic data, each component gets an enhanced item (with the new property).
* The child sets it's obscured property to true if it's obscured
* The parent has a computed `showToggler: Ember.computed.any('children', 'obscured')`

Again, I'm not sure that would work due to the new data-down-actions-up behavior. Setting the individual property by the child might not propagate up.

JoshSmith:
This is now ready for review. I expect to check the `matched` property later once skills are actually available here.

Also have not written a categories integration test as that will be moved elsewhere as part of onboarding, and we should a write a test for that then.

app/models/project.js
```diff
@@ -8,6 +8,40 @@ export default DS.Model.extend({
...
+  //   {
+  //     name: 'PostgreSQL'
+  //   },
+  //   {
+  //     name: 'ActiveModel::Serializers'
+  //   },
+  //   {
+  //     name: 'JSON API'
+  //   },
+  // ]
```
begedin: We keeping this for future manual testing, or did you miss it?

package.json
```diff
@@ -26,7 +26,7 @@
     "ember-cli-app-version": "^1.0.0",
     "ember-cli-babel": "^5.1.6",
     "ember-cli-bourbon": "1.2.1",
-    "ember-cli-content-security-policy": "^0.5.0",
+    "ember-cli-content-security-policy": "rwjblue/ember-cli-content-security-policy",
```
begedin: Is there an update I'm not aware of?

app/routes/projects-list.js
```diff
@@ -0,0 +1,11 @@
...
+
+export default Ember.Route.extend({
+  model() {
+    return this.store.findAll('project');
+  },
+
+  setupController(controller, model) {
+    controller.set('projects', model);
+  }
+});
```
begedin: Is there a reason this isn't just  `/projects/index.js` ?

begedin:
Looks good, but I do have a couple of questions. Probably good to merge, though.

package.json
JoshSmith: Yeah, 14 days ago, but couldn't get it working still the way I wanted.

app/models/project.js
JoshSmith: Was keeping it for the moment, planning on removing later.

begedin:
👍 🚢 🇮🇹

PR #127, opened 2016-05-25
Redo filtering

JoshSmith:
This is a spike for redoing the way post filtering works. Probably broke a bunch of tests and it's not really ideal right now in terms of its execution, but the general idea is there.


JoshSmith:
Whoo! Passing tests for a pretty large feature. I think we're probably missing some tests, but this is a pretty good milestone for this feature.

@begedin could really use your thoughts when you do have a moment. We could punt refactoring for later, maybe?

JoshSmith:
And I mean major refactoring. At least to get this merged.

app/components/post-filter-type.js
```diff
@@ -0,0 +1,22 @@
...
+      this.toggleProperty('hidden');
+    },
+    hide: function() {
+      this.set('hidden', true);
+    },
+    filterByType: function(type) {
+      this.sendAction('filterByType', type);
+    },
+  }
+});
```
begedin: Maybe I'm missing a binding somewhere, but couldn't we just use the `active` property directly here? `toggle` would toggle it as always, `hide` would set it to false.

app/components/scroll-top.js
```diff
@@ -0,0 +1,7 @@
+import Ember from 'ember';
+
+export default Ember.Component.extend({
+  click() {
+    window.scrollTo(0,0);
+  }
+});
```
begedin: I especially love this one :)

app/components/user-menu.js
```diff
@@ -1,7 +1,7 @@
 import Ember from 'ember';
 
 export default Ember.Component.extend({
-  classNames: ['user-menu'],
+  classNames: ['user-menu', 'dropdown'],
```
begedin: I like to keep a component-specific class name around just for the sake of easier test writing. Not sure where you stand on that.

app/controllers/project/posts/index.js
```diff
@@ -2,5 +2,106 @@ import Ember from 'ember';
...
+      if(Ember.isEmpty(postTypes)) {
+        this.set('postType', null);
+      } else {
+        let types = postTypes.join(',');
+        this.set('postType', types);
+      }
+
+      this.resetPage();
+    }
+  }
```
begedin: This is OK for now, but controllers will go away eventually and be replaced with a component, so we should keep that in mind.

begedin:
Marked as "needs work" only for the sake of that one comment on the dropdown component and the fact that the tests seem to have failed. I was under the impression they were passing for this one.

app/components/user-menu.js
JoshSmith: I generally agree but sometimes it leads to overblown class names and over-specific CSS. Already we need to do some refactoring on the CSS side because of this.

app/controllers/project/posts/index.js
JoshSmith: Yeah, it seems like they haven't figured this out yet in Ember, so I will play wait-and-see.

Issue #131, opened 2016-05-28
Dry up containsCode logic


JoshSmith:
There is some duplication in the `containsCode` logic in the comment and post models. I'm also not sure if this is the best way to be handling this, or if there's something that can be done server-side. For now, it works.

idelairre:
How about a mixin?

Also, I just took a whack at this using `includes` instead of `indexOf` and Phantomjs threw an error. I'm pretty sure this is because phantom hasn't implemented it yet. `includes` is more expressive but I'm not sure the best way to go about polyfilling Phantomjs or if its worth it.

In the mean time, I can just submit a PR for a mixin using `indexOf`.

JoshSmith:
@idelairre that sounds like a pretty good approach to me, so long as we can unit test the mixin, as well!

Issue #130, opened 2016-05-28
Fix the way new comments get posted


JoshSmith:
Right now when a comment gets posted, it just looks like the preview is being generated, and the button state does not change at all. We should fix this and make the experience a lot smoother.

JoshSmith:
Also appears to be the case that the preview is changing height or something.

JoshSmith:
Here's a video that demos the bugs that need fixed here: https://s3.amazonaws.com/f.cl.ly/items/413H2P1S16351W3U0y0r/BugDemo.mov

The height changes slightly (flickers) when previewing and submitting, and the comment form does not change back from the previewing state.

JoshSmith:
The bug, on a particular post:

- Write something.
- Hit preview button.
- Hit comment submit button.
- The comment will submit but the preview button remains highlighted and the preview remains there.

Expected behavior:

- Write something.
- Hit preview button.
- Hit comment submit button.
- The comment submits, the button changes to "Write", and the preview is cleared.

Issue #129, opened 2016-05-28
Fix post previewing


JoshSmith:
The way we're handling post previewing doesn't actually allow the post type or title to be set, which means that the type gets reset on pulling data down from the server. All we want to avoid is generating comment notifications and publishing a post, not altogether avoiding other data being set.

begedin:
I'm not sure I completely follow here.

The `post-title` component actually displays a property that is completely separate from the record. The user can edit that property as they see fit and then, when they click "save", the value of the property is set to the post title and the post record is saved.

if we are just generating the preview, then yes, upon a server response, post title will be reset, but we're still displaying the separate `newTitle` property. I guess we need the same behavior for `postType`?

I'll see about implementing that.

PR #132, opened 2016-05-28
Refine comment editing

JoshSmith:
Had to reopen this due to weird rebase issues.

I have broken out issues #129, #130, and #131 to not continue to make this PR larger than it needs to be.

I did sneak in some things like image optimizations and refactoring of CSS. And I'm sure I touched a number of other places.

Everything I wanted to test appears to be tested. Now could just use a review.


app/templates/components/code-theme-selector.hbs
```diff
@@ -0,0 +1 @@
+{{yield}}
```
begedin: Do we actually have a reason to yield here? As in, do we expect to use the component in a way that provides it HTML to render? If not, we could just have an empty file here instead (and remove the test that ensures yielding works, if there is one).

JoshSmith: I don't believe so. Not sure there is such a test either.

app/templates/project/posts/post.hbs
```diff
@@ -2,10 +2,12 @@
 <div class="post-timeline">
   {{post-details post=post}}
   {{post-comment-list comments=comments}}
+  <div>
   {{create-comment-form comment=newComment saveComment='saveComment'}}
   {{#if error}}
     {{error-formatter error=error}}
   {{/if}}
+  </div>
```
begedin: Missing some indentation between these div tags. Unless we prefer not to have it here?

JoshSmith: Just an oversight. Can indent.

app/templates/components/code-theme-selector.hbs
begedin: It's usually auto-generated if you're generating a component, so there might be.

JoshSmith: There's a component test, but I think I modified it.

begedin:
Looks good, other than the couple of very minor details. Feel free to merge :+1:

PR #135, opened 2016-05-30
Update mirage and rework tests

JoshSmith:
A collection of minor test changes to make our Mirage upgrade work.

### DIFF
--- a/app/routes/project/posts/post.js
+++ b/app/routes/project/posts/post.js
@@ -25,6 +25,7 @@ export default Ember.Route.extend({
   },
 
   setupController(controller, models) {
+    this._super(controller, models);
     controller.set('post', models.post);
     controller.set('newComment', models.comment);
     controller.set('comments', models.comments);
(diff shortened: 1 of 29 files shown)


JoshSmith:
@begedin I'm merging because I need to make use of this ASAP and it's relatively straightforward. But comment here if you have some issues and we can circle back in a separate issue/PR to correct them.

app/routes/project/posts/post.js
```diff
@@ -25,6 +25,7 @@ export default Ember.Route.extend({
   },
 
   setupController(controller, models) {
+    this._super(controller, models);
```
begedin: `setupController` is generally a hook where we might not want to call `_super`. The reason for that is that, by default, `setupController` looks like

```Javascript
setupController(controller, models) {
  return controller.set('model', models);
}
```

It doesn't really hurt anything, but it does add a "model" property which we aren't using to the controller.

begedin:
Everything looks good. There's that minor `setupController` thing, but It's not really an issue, so no need to change anything now.

PR #142, opened 2016-06-02
Hide pagination when less than 2 pages

JoshSmith:
We want to hide the pagination when there's only a single page. Makes no sense to have it in that case.


begedin:
👍 🚢 🇮🇹

PR #143, opened 2016-06-02
Add signup form for comment if signed out

JoshSmith:
This adds a signup form for comments if the user is signed out.


begedin:
Looks good and seems pretty straightforward. Good to merge as far as I'm concerned. 

Just a side-question, though. I believe ember-simple-auth will redirect back to the route you started the signup/signin from. Does this still work?

JoshSmith:
@begedin doesn't work on its own like that. Can't remember how to do it, either. Was trying to ask that yesterday in Ember Slack. If you find out, definitely need to know.

PR #145, opened 2016-06-03
Add onboarding

JoshSmith:
This introduces onboarding and adds current-user, navigation-menu, onboarding service, and associated components.

Lots needed to test here.

- [x] Write tests for current-user service
- [x] Write tests for navigation-menu service
- [x] Write tests for onboarding service
- [x] Write tests for categories-list component
- [x] Write tests for navigation-menu component
- [x] Write tests for progress-bar component
- [x] Write tests for index route
- [ ] Remove session.currentUser and rely on currentUser.user


JoshSmith:
Until we bring in those changes from `ember-cli-mirage`, I don't see any way that we can reasonably expect to integration test this. We need to come back around to it.

JoshSmith:
I made some comments here. The `include` functionality helps with tests, but I have to manually reload the user after a category gets added for the changes to reflect. Not sure if there's supposed to be some other way for this to be triggered.

JoshSmith:
Moving out of awaiting review because this is no longer ready. I got acceptance tests to work and refactored a bunch of things. I think more can be done, but the testing story is now accurately incomplete.

Will circle back later.

app/routes/index.js
```diff
@@ -0,0 +1,13 @@
...
+
+const { service } = Ember.inject;
+
+export default Ember.Route.extend({
+  session: service(),
+
+  beforeModel() {
+    if (this.get('session.isAuthenticated')) {
+      this.transitionTo('projects-list');
+    }
```
JoshSmith: Not tested yet.

app/services/navigation-menu.js
```diff
@@ -0,0 +1,26 @@
...
+      return this.get('defaultMenuType');
+    }
+  }),
+
+  isDefault: Ember.computed('menuType', 'defaultMenuType', function() {
+    return this.get('menuType') === this.get('defaultMenuType');
+  }),
+  isOnboarding: Ember.computed('menuType', 'onboardingMenuType', function() {
+    return this.get('menuType') === this.get('onboardingMenuType');
+  }),
```
JoshSmith: I wish there were a better way to do this.

tests/unit/controllers/start/expertise-test.js
```diff
@@ -0,0 +1,12 @@
...
+moduleFor('controller:start/expertise', 'Unit | Controller | start/expertise', {
+  // Specify the other units that are required for this test.
+  // needs: ['controller:foo']
+});
+
+// Replace this with your real tests.
+test('it exists', function(assert) {
+  let controller = this.subject();
+  assert.ok(controller);
+});
```
JoshSmith: Should we test here?

tests/unit/controllers/start/interests-test.js
```diff
@@ -0,0 +1,12 @@
...
+moduleFor('controller:start/interests', 'Unit | Controller | start/interests', {
+  // Specify the other units that are required for this test.
+  // needs: ['controller:foo']
+});
+
+// Replace this with your real tests.
+test('it exists', function(assert) {
+  let controller = this.subject();
+  assert.ok(controller);
+});
```
JoshSmith: Should we test here?

tests/unit/controllers/start/skills-test.js
```diff
@@ -0,0 +1,12 @@
...
+moduleFor('controller:start/skills', 'Unit | Controller | start/skills', {
+  // Specify the other units that are required for this test.
+  // needs: ['controller:foo']
+});
+
+// Replace this with your real tests.
+test('it exists', function(assert) {
+  let controller = this.subject();
+  assert.ok(controller);
+});
```
JoshSmith: Should we test here?

tests/unit/mixins/onboarding-controller-test.js
```diff
@@ -0,0 +1,12 @@
...
+import { module, test } from 'qunit';
+
+module('Unit | Mixin | onboarding controller');
+
+// Replace this with your real tests.
+test('it works', function(assert) {
+  let OnboardingControllerObject = Ember.Object.extend(OnboardingControllerMixin);
+  let subject = OnboardingControllerObject.create();
+  assert.ok(subject);
+});
```
JoshSmith: This probably needs tested.

tests/unit/mixins/onboarding-route-test.js
```diff
@@ -0,0 +1,12 @@
...
+import { module, test } from 'qunit';
+
+module('Unit | Mixin | onboarding route');
+
+// Replace this with your real tests.
+test('it works', function(assert) {
+  let OnboardingRouteObject = Ember.Object.extend(OnboardingRouteMixin);
+  let subject = OnboardingRouteObject.create();
+  assert.ok(subject);
+});
```
JoshSmith: This also probably needs tested.

tests/unit/routes/start-test.js
```diff
@@ -0,0 +1,11 @@
...
+
+moduleFor('route:start', 'Unit | Route | start', {
+  // Specify the other units that are required for this test.
+  // needs: ['controller:foo']
+});
+
+test('it exists', function(assert) {
+  let route = this.subject();
+  assert.ok(route);
+});
```
JoshSmith: Test this.

tests/unit/routes/start/expertise-test.js
```diff
@@ -0,0 +1,11 @@
...
+
+moduleFor('route:start/expertise', 'Unit | Route | start/expertise', {
+  // Specify the other units that are required for this test.
+  // needs: ['controller:foo']
+});
+
+test('it exists', function(assert) {
+  let route = this.subject();
+  assert.ok(route);
+});
```
JoshSmith: Test this.

tests/unit/routes/start/interests-test.js
```diff
@@ -0,0 +1,11 @@
...
+
+moduleFor('route:start/interests', 'Unit | Route | start/interests', {
+  // Specify the other units that are required for this test.
+  // needs: ['controller:foo']
+});
+
+test('it exists', function(assert) {
+  let route = this.subject();
+  assert.ok(route);
+});
```
JoshSmith: Test this.

tests/unit/routes/start/skills-test.js
```diff
@@ -0,0 +1,11 @@
...
+
+moduleFor('route:start/skills', 'Unit | Route | start/skills', {
+  // Specify the other units that are required for this test.
+  // needs: ['controller:foo']
+});
+
+test('it exists', function(assert) {
+  let route = this.subject();
+  assert.ok(route);
+});
```
JoshSmith: Test this.

app/services/session.js
```diff
@@ -8,7 +8,7 @@ export default ESASession.extend({
   setCurrentUser: function() {
     if (this.get('isAuthenticated')) {
       let id = this.get('session.authenticated.user_id');
-      this.get('store').findRecord('current-user', id).then((user) => {
+      this.get('store').findRecord('user', id).then((user) => {
         this.set('currentUser', user);
```
JoshSmith: We should really consider getting rid of the `session.currentUser` and relying instead on `currentUser.user`.

JoshSmith:
> Acceptance tests will likely provide better coverage for routes because routes are typically used to perform transitions and load data, both of which are tested more easily in full context rather than isolation.

So many of these route tests may be unnecessary if tested at the acceptance level.

app/services/navigation-menu.js
begedin: You could do 

```Javascript
Ember.computed.equal('menuType', 'default')
Ember.computed.equal('menuType', 'onboarding')
```

But then you lose the option of storing "default" and "onboarding" as property.

app/services/session.js
begedin: I agree. Now that we have the service, no point not using it.

tests/unit/controllers/start/expertise-test.js
begedin: Generally,  if It's a unit test, then I would only test the code that doesn't involve other parts of the app. Otherwise, if there is some behavior that needs testing, but it interacts with other parts, then I'd make an integration test. In this case, we'd likely want the integration test that ensures we have a user property.

In the case of the interests controller, there are two actions that need testing. Since they deal with the store, again, I'd go with an integration test.

begedin:
I'm not seeing anything I'd do differently here. Just needs test coverage.

I'd leave the "it exists" tests for routes, but yes, I agree that acceptance tests are probably best for general behavior of those. Still, there's room for unit tests if there's some sort of complex behavior within the route.

PR #147, opened 2016-06-05
Add the arts as a category to the front-end designs

JoshSmith:
This adds "Arts" as a category, per a request from @kakoga.

Went with "Arts" as wording because it is more all-encompassing, and although pluralized, is similar to "Politics" in that regard. It makes more sense to say that Code Corps projects are "improving the arts" than to say they are "improving art." Open for discussion, though.


begedin:
This actuall adds the graphics for the API-defined arts category and it doesn't involve any actual code, correct? Looks good, but I would maybe change the title of the PR for merging to make that clearer.

JoshSmith:
Yeah, the API itself is the only thing that adds that. There's a separate PR. It's just a line for seeding the DB.

PR #150, opened 2016-06-05
Rewrite flash and make category adding snappier

### LINKED ISSUE #149: Add loading state to category-item buttons
These are pretty non-responsive in the UI right now. We need to adding a loading indicator and modify the text while the change is happening.

JoshSmith:
Closes #149.

This rewrites the add/remove category buttons to include flash messages when errors happen, and also adds in a loading indicator, which takes advantage of closure actions returning a promise.

- [x] Add test for category item loading spinners
- [x] Add test for category item flash messages


JoshSmith:
Here's a demo of the loading state on a simulated 2 second load time. Most of the time it will be snappier and render without issue, which is why the spinner fades in, to avoid a flicker.



And here's a look at the error state on a simulated 2s load time with a simulated error. Unlikely to happen very often, but in case it does:

app/components/category-item.js
```diff
@@ -15,4 +18,37 @@ export default Ember.Component.extend({
     let userCategories = this.get('userCategories');
     return userCategories.hasCategory(category);
   }),
+
+  actions: {
+    addCategory(category) {
+      this.set('isLoading', true);
+      this.addCategory(category).catch(() => {
```
begedin: I'm sorry. I don't think I'm actually following this. Where is `this.addCategory` defined here?

JoshSmith: This is the beautiful thing about the closure actions. We're passing down `addCategory` into the component, and that's a function defined up top with `action (action addCategory)`, which itself is defined at the controller level.

begedin: Ah, got it. I'm not sure I absolutely love that, though. Having it one place is great, but It wasn't clear to me we're passing it down, for example.

Is it possible to change the name while passing it down, to make it clearer where it originates from? Alternatively, maybe `addCategory` and `removeCategory` could be functions of the `currentUser` service?

JoshSmith: Something like the latter, the `user-category` service, maybe makes the most sense since the data is being handled there anyway, and the way we add/remove a category via the API will not change implementation across UIs (like a user's profile editing, for instance). There's no need to have that happen in the controller.

`{{categories-list categories=model addCategory=(action 'addCategory') removeCategory=(action 'removeCategory')}}` <= and that is weird. It works right now, but doesn't mean it's right. Ostensibly we would pass _nothing_ into the component, but inside the component itself we can handle the `addCategory` action and then call out to the `userCategory` service.

begedin: That sounds good to me. Do you want to do it here, or in a separate PR?

JoshSmith: Maybe a separate issue and separate PR. We can come back to it as I'd like to move on to some other things for now.

begedin:
👍 looks good, considering our discussion. We can create a separate issue and merge this.

PR #152, opened 2016-06-06
Rewrite addCategory and removeCategory, fix spinner and change categories grid

JoshSmith:


tests/integration/components/navigation-menu-test.js
```diff
@@ -4,10 +4,6 @@ import Ember from 'ember';
 
 moduleForComponent('navigation-menu', 'Integration | Component | navigation menu', {
   integration: true,
-  // this.container.registry.register('service:current-user', mockCurrentUserService);
```
begedin: Just for future reference, we can also use `this.register`

begedin:
Looks good. I especially like that there's less stuff in controllers now. Less trouble later on.

PR #154, opened 2016-06-06
Replace session with current-user where appropriate and fix some styling issues

JoshSmith:
This replaces `session` with `current-user` where it makes sense and also has some minor styling fixes.


begedin:
:+1:

PR #155, opened 2016-06-07
Add style guide

### LINKED ISSUE #87: Organization of files
It is beneficial to a project to keep a clear organizational pattern to the files themselves. In my experience this has been the following for ember-cli modules:

* static attributes
* service attributes
* aliased/one-line computed attributes
* computed properties
* methods
* actions

It is also my opinion that within each of those groups that the items be alphabetized but this only becomes noticeably beneficial when working with larger modules, which one might argue should potentially be broken up in smaller files at that point.

An example ember-cli component (totally foobar) with the aforementioned organizational style

```js
import Ember from 'ember';

  classNames: ['example-component'],
  type: 'task',

  session: Ember.inject.service(),

  editing: Ember.computed.equal('mode', 'editing'),
  previewing: Ember.computed.equal('mode', 'previewing'),
  selectedItem: Ember.computed.alias('post')

  value: Ember.computed('selectedItem', function() {
    // ...
  }),

  didInitAttrs() {
    this._super(...arguments);
    this.mode = 'editing';
  },

  actions: {
    edit() {
      this.set('mode', 'editing');
    },

    preview() {
      this.set('mode', 'previewing');
      this.set('preview', 'Loading preview...');
      let content = this.get('input');
      this.sendAction('generatePreview', content);
    }
  }
});
```

JoshSmith:
Closes #87.

Needs expanded later but for now satisfies the need.

### DIFF
--- a/README.md
+++ b/README.md
@@ -40,6 +40,10 @@ Make use of the many generators for code, try `ember help generate` for more det
 * `ember build` (development)
 * `ember build --environment production` (production)
 
+### Style Guide
+
+You should refer to our [style guide](STYLEGUIDE.md) for writing Ember.js, HTML/Handlebars, and SCSS that makes it easier for everyone to work on Code Corps together.
+
 ### Deploying
 
 Specify what it takes to deploy your app.
--- a/STYLEGUIDE.md
+++ b/STYLEGUIDE.md
@@ -1,0 +1,141 @@
+# Code Corps Style Guide
+
+## Table of contents
+
++ [Ember.js](#ember.js)
++ [HTML and Handlebars](#html-and-handlebars)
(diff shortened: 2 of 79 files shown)


JoshSmith:
@caligoanimus adding the style guide here per #87. Made some modifications. It's a big changeset because I'm modifying files to conform to it, but you really only need to review the style guide and give thoughts there.

@begedin could also use your thoughts.

begedin:
I agree with the guidelines, but I'm not looking forward to the merge. A lot of changed files, which will likely clash with #120.

JoshSmith:
@begedin yeah, was going to come up one way or another, though! I'll do the merge into #120.

PR #159, opened 2016-06-08
Add way to use Mirage in development and without redis

### LINKED ISSUE #74: Investigate adding redis flag in development
Expecting users to have `redis-server` and our API running just to load up the UI and contribute is a high bar. We should see if there's some way to put this behind a flag so we can make it optional, rather than required.

### LINKED ISSUE #94: Get Mirage working in dev via some flag
Related to #74 - Allow for code-corps-ember to run in dev without needing the api running locally via ember-cli-mirage

JoshSmith:
Attempt to close #74 and close #94.

I'm not sure how big a fan I am of this approach. While it technically works, a good amount of work would required to keep it up to date. It may be easier to give our Ember developers a packaged Rails app API that they can hit. Alternatively, we could have a flag that sets the API endpoint and they hit some non-staging development server.

Certain things, like syntax highlighting, simply _require_ a working API.

I'm actually more inclined to say Mirage is not appropriate for this use case, and that I would probably vote in favor of closing this PR unmerged. #94 specifically would then be closed, and #74 would be addressed by not having a deploy step, but instead hitting the API endpoint in a different environment, circumventing `ember-cli-deploy` altogether.

### DIFF
--- a/config/environment.js
+++ b/config/environment.js
@@ -50,6 +50,14 @@ module.exports = function(environment) {
     }
   };
 
+  if (environment === 'mirage') {
+    ENV.OAUTH_SERVER_TOKEN_ENDPOINT = '/oauth/token/';
+
+    ENV['ember-cli-mirage'] = {
+      enabled: true
+    };
+  }
+
   if (environment === 'development') {
     // ENV.APP.LOG_RESOLVER = true;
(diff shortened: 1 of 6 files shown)


JoshSmith:
Closing in favor of #179.

PR #160, opened 2016-06-08
Redirect from /login and /signup when authed

### LINKED ISSUE #125: Redirect from signup when logged in
This route should only be accessible when logged out.

JoshSmith:
Fixes #125. Redirects when authed from routes for logging in and signing up. Really straightforward.


begedin:
👍 🚢 🇮🇹

PR #161, opened 2016-06-09
Add roles

JoshSmith:
This reworks user categories slightly and also adds roles for a user. Adds to the acceptance test for onboarding.

There's a lot of similar code between this and categories but I think it's just incidental duplication. Still might be worth thinking about ways to reduce some of that duplication.

A good opportunity might actually be with the buttons used in these components. They could be made into some sort of async check button component or something.


app/routes/start/expertise.js
```diff
@@ -2,4 +2,14 @@ import Ember from 'ember';
 import OnboardingRouteMixin from '../../mixins/onboarding-route';
 
 export default Ember.Route.extend(OnboardingRouteMixin, {
+  currentUser: Ember.inject.service(),
+
+  beforeModel() {
+    let user = this.get('currentUser.user');
+    return user.get('user-role');
```
begedin: Is this just for "preloading"?

begedin:
👍 Looks good. The question isn't really an issue, was just wondering.

app/routes/start/expertise.js
JoshSmith: Yeah, but doesn't really force it like I wanted to. Would like it to.

PR #163, opened 2016-06-10
Add user skills

JoshSmith:
This has no tests and the input's not there yet. Lots of work left to do.

Big thing I can use help on at this point is the `highlight-substrings` helper. I'm down a rabbit hole and can't see my way out with that one.

### DIFF
--- a/mirage/config.js
+++ b/mirage/config.js
@@ -1,24 +1,6 @@
 import Mirage from 'ember-cli-mirage';
 import Ember from 'ember';
 
-function generatePostMentions(schema, post, mentionStatus) {
-  let body = post.body || '';
-  let matches = body.match(/@\w+/g) || [];
-
-  matches.forEach((match) => {
-    let username = match.substr(1);
-    let matchedUser = schema.users.where({ username: username }).models[0];
-    if (matchedUser) {
-      let startIndex = body.indexOf(match);
-      let endIndex = startIndex + match.length - 1;
-      schema.postUserMentions.create({
-        username: username, indices: [startIndex, endIndex], status: mentionStatus,
-        userId: matchedUser.id, postId: post.id
-      });
-    }
-  });
-}
-
 function generateCommentMentions(schema, comment, mentionStatus) {
   let body = comment.body || '';
   let matches = body.match(/@\w+/g) || [];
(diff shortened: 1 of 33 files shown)


JoshSmith:
@begedin could use your help on this.

begedin:
@JoshSmith Looks very clean to me. I got no remarks, component or any other part of the code.

JoshSmith:
@begedin this is mostly done, it just needs a review and also needs some love on these tests. I can't figure out why they're passing on their own but failing in other instances. Can you take a look?

JoshSmith:
Weird. The tests pass here but not in my browser when run all at once.

begedin:
@JoshSmith Tests run fine on my machine. Could be it's the same issue I had in #120. I Updated the dependencies, so try pulling the latest, runing bower up and testing again.

mirage/config.js
```diff
@@ -85,6 +85,21 @@ export default function() {
 
   this.delete('/user_categories/:id');
 
+  this.get('/skills');
+  this.get('/user_skills');
+
+  // TODO: Make this work when relationships work
```
begedin: Which part of this doesn't work? A join model should be fine, since, unlike projects, there isn't a user -> skill many-to-many relationship here, just a user > userSkills, and skill > userSkills.

begedin:
looks good. there's that one question, because I'm not understanding why it doesn't work, but other than that, seems fine.

mirage/config.js
JoshSmith: Maybe I'm misunderstanding then. I can remove this TODO in that case?

begedin: You might even be able to outright remove the custom function and just have `this.post('/user_skills')`

JoshSmith: This is the case for `user-skill`, `user-role`, and `user-category` then, right?

JoshSmith: @begedin I don't think I can. I think Sam mentioned this was still many-to-many so it wouldn't work because Mirage doesn't support it out of the box right now.

begedin: It's not a many to many, though

It's a `one-to-many` `user` -> `userSkill` and a `one-to-many` `skill` -> `userSkill`. user -> skills would be a many to many relationship, but we don't have that on our model here. What we have is what mirage should already be supporting.

begedin: In any case, I would try and remove this custom method and see if the tests run without it. I believe they should.

JoshSmith: Yeah, let me verify.

Issue #165, opened 2016-06-14
Add analytics


JoshSmith:

JoshSmith:
Many of these should be tracked server-side for data integrity reasons. Only front-end specific actions should be tracked here.

JoshSmith:
This is closed by #170 for the most part. We may want to add more iterative things later, but the server-side stuff we're tracking is good enough at the moment.

PR #170, opened 2016-06-14
Add ember metrics

JoshSmith:
This does some of what's needed for #165. Have not implemented any of the details, but it's working with at least Segment and Intercom.


begedin:
👍

PR #178, opened 2016-06-16
Add footer

### LINKED ISSUE #176: Add footer
Would like a footer with:

At least minimal information about Code Corps. Smaller logo, too?

Maybe link to an about page? Also:

- Twitter
- Facebook
- Github
- Slack

JoshSmith:
Closes #176.

Need to add in the actual links later.


app/services/user-categories.js
```diff
@@ -25,14 +25,16 @@ export default Ember.Service.extend({
...
-    }.bind(this));
-    return userCategory;
+    if (userCategories) {
+      let userCategory = userCategories.find(function(item) {
+        let itemUserId = item.belongsTo('user').id();
+        let itemCategoryId = item.belongsTo('category').id();
+        let userId = this.get('user.id');
+        let categoryId = category.get('id');
+        return (itemUserId === userId) && (itemCategoryId === categoryId);
+      }.bind(this));
```
begedin: Instead of `bind(this)`, you should just be able to use the arrow function.
```Javascript
let oldThis = this;
let userCategory = userCategories.find((item) => {
   let newThis = this;
   return newThis === oldThis // true
});
```

begedin:
I would give it a go with the arrow function. Otherwise, this looks good :+1:

JoshSmith:
Let's circle back around to that change because there are other places for it, and I'm not sure that's tested right. I fixed a bug but no test for it.

JoshSmith:
Can you create an issue to go back and find/fix all `.bind` instances to use `=>`?

begedin:
#191

PR #185, opened 2016-06-16
Add post types based on permissions

### LINKED ISSUE #174: Fix selector on new post to be permissions-specific
Right now anyone can select any post type regardless of their ability to post. We need to modify `ember-can` to transform the array of possible post types on a per-user basis.

begedin:
Closes #174 

I used the existing ember-can `organization` ability, since permissions are organization based.

All of this simply removes the option from the select dropdown if the user has no ability. Technically, the HTML could still be edited to create an invalid selection, but we have API checks in place for that.

### DIFF
--- a/app/templates/components/post-new-form.hbs
+++ b/app/templates/components/post-new-form.hbs
@@ -1,12 +1,20 @@
 <img class="icon" src={{post.user.photoThumbUrl}} />
 <div class="post-new-form-body">
   <div class="input-group post-type {{post.postType}}">
-    {{select-dropdown
-      name='post-type'
-      items=types
-      selectedItem=post.postType
-      optionValuePath='slug'
-      optionLabelPath='label'}}
+    <select name="post-type" value={{post.postType}}>
+      {{#if (can "create task post in organization" organization membership=credentials.currentUserMembership)}}
+        <option value="task" selected={{eq post.postType 'task'}}>Task</option>
+      {{/if}}
+      {{#if (can "create issue post in organization" organization membership=credentials.currentUserMembership)}}
+        <option value="issue" selected={{eq post.postType 'issue'}}>Issue</option>
+      {{/if}}
+      {{#if (can "create progress post in organization" organization membership=credentials.currentUserMembership)}}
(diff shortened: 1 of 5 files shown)


tests/integration/components/post-new-form-test.js
```diff
@@ -64,3 +67,16 @@ test('it has a loading indicator when previewing and post is saving', function(a
...
+
+  this.register('service:credentials', Ember.Service.extend({
+    currentUserMembership: { isContributor: true }
+  }));
+
+  this.render(hbs`{{post-new-form post=post placeholder=placeholder}}`);
+
+  assert.equal(this.$('option[value=task]').length, 1, 'task option is rendered');
+  assert.equal(this.$('option[value=progress]').length, 1, 'progress option is rendered');
+});
```
JoshSmith: I don't think we're testing enough here. I think we need an explicit test for them not being a contributor. I would also test that the contributor has the other options still.

app/templates/components/post-new-form.hbs
```diff
@@ -1,12 +1,20 @@
...
+      {{#if (can "create issue post in organization" organization membership=credentials.currentUserMembership)}}
+        <option value="issue" selected={{eq post.postType 'issue'}}>Issue</option>
+      {{/if}}
+      {{#if (can "create progress post in organization" organization membership=credentials.currentUserMembership)}}
+        <option value="progress" selected={{eq post.postType 'progress'}}>Progress</option>
+      {{/if}}
+      {{#if (can "create idea post in organization" organization membership=credentials.currentUserMembership)}}
+        <option value="idea" selected={{eq post.postType 'idea'}}>Idea</option>
+      {{/if}}
+    </select>
```
JoshSmith: This is a good way to do this. Much better than I had initially imagined.

JoshSmith:
With those test changes this would be good to go.

begedin:
@JoshSmith I added tests for non-contributors there. This should be good to go now.

JoshSmith:
Perfect changes!

PR #194, opened 2016-06-17
Add long description to project, functional, needs styling

JoshSmith:
This still has a lot of work left to go on it.


JoshSmith:
I got this and will be able to work on it now since API side should be done.

PR #195, opened 2016-06-17
Fix bug with user details

### LINKED ISSUE #189: Don't show user details if they don't have them
On the user's profile, we should not show certain details if they don't have them, like twitter, URL, etc.

JoshSmith:
Closes #189.


app/templates/components/user-sidebar.hbs
```diff
@@ -1,13 +1,21 @@
 <img src={{user.photoLargeUrl}} />
 <h2>
-  <div class="name">{{user.name}}</div>
+  {{#if user.name}}
+    <div class="name">{{user.name}}</div>
+  {{else}}
+    <div class="name">{{user.username}}</div>
+  {{/if}}
   <div class="username">{{user.username}}</div>
```
begedin: This will cause a duplicate display of the username in the case where the user has no actual name:

```HTML
<div class="name">username</div>
<div class="username">username</div>
```

JoshSmith: Yep, this is what I'm aiming for and there's a test for it.

PR #203, opened 2016-06-17
Add submittable textarea

JoshSmith:
This adds a submittable textarea. You can hit ctrl+enter or cmd+enter to submit.


begedin:
:+1:

PR #210, opened 2016-06-18
Add project skill lists to UI

JoshSmith:
No tests really. Fixes some things with project skills.

### DIFF
--- a/app/components/skill-list-item.js
+++ b/app/components/skill-list-item.js
@@ -4,6 +4,18 @@ export default Ember.Component.extend({
   classNameBindings: ['matched'],
   tagName: 'li',
 
+  matched: Ember.computed.notEmpty('userSkill'),
+  userSkills: Ember.inject.service(),
+  user: Ember.computed.alias('currentUser.user'),
+  usersUserSkills: Ember.computed.alias('user.userSkills'),
+
+  userSkill: Ember.computed('skill', 'userSkills.userSkills.isFulfilled', function() {
+    let skill = this.get('skill');
+    let userSkills = this.get('userSkills');
+    let result = userSkills.findUserSkill(skill);
+    return result;
+  }),
+
   didRender() {
     this._super(...arguments);
     let parentBottom = this.$().parent()[0].getBoundingClientRect().bottom;
--- a/app/services/user-skills.js
+++ b/app/services/user-skills.js
@@ -23,6 +23,16 @@ export default Ember.Service.extend({
     return userSkill.save();
   },
 
+  hasSkill(skill) {
+    let userSkills = this.get('userSkills');
+    let matchedSkill = userSkills.find(function(item) {
+      let itemSkillId = item.belongsTo('skill').id();
+      let skillId = skill.get('id');
+      return (itemSkillId === skillId);
+    }.bind(this));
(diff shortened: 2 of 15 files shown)


app/components/skill-list-item.js
```diff
@@ -4,6 +4,19 @@ export default Ember.Component.extend({
...
+  userSkills: Ember.inject.service(),
+  user: Ember.computed.alias('currentUser.user'),
+  usersUserSkills: Ember.computed.alias('user.userSkills'),
+
+  userSkill: Ember.computed('skill', 'userSkills.userSkills.isFulfilled', function() {
+    let skill = this.get('skill');
+    let userSkills = this.get('userSkills');
+    let result = userSkills.findUserSkill(skill);
+    skill.set('matched', result);
+    return result;
```
JoshSmith: I don't really like the way I'm setting the virtual attr `matched` here. Would prefer a better way. This is not really data down.

app/routes/projects-list.js
```diff
@@ -1,6 +1,14 @@
 import Ember from 'ember';
 
 export default Ember.Route.extend({
+  currentUser: Ember.inject.service(),
+
+  beforeModel() {
+    this._super(...arguments);
+    let user = this.get('currentUser.user');
+    user.get('skill');
+  },
```
JoshSmith: Don't know if this is necessary.

tests/integration/components/skill-list-item-test.js
```diff
@@ -31,7 +31,7 @@ test('it renders and sends no action when not hidden', function(assert) {
   assert.expect(1);
 
   let skill = {
-    name: 'Ruby'
+    title: 'Ruby'
   };
```
JoshSmith: I would actually love to rename this in the backend to `name`.

JoshSmith:
@begedin I could use your thoughts now before I commit too much to this approach.

app/components/skill-list-item.js
begedin: Only other thing I could think of is to provide a callback to the sorting computed in the parent component. It's less weird, but it's also more work "at once", I think. Not sure which one would be quicker, though.

app/routes/projects-list.js
hbrysiewicz: What made you think it was necessary? Removing this seems to be ok.. but curious what your concern was and why it was added.

JoshSmith: I thought that the user's skills were not loading explicitly from the user skills service.

app/services/user-skills.js
```diff
@@ -23,6 +23,16 @@ export default Ember.Service.extend({
     return userSkill.save();
   },
 
+  hasSkill(skill) {
+    let userSkills = this.get('userSkills');
+    let matchedSkill = userSkills.find(function(item) {
+      let itemSkillId = item.belongsTo('skill').id();
+      let skillId = skill.get('id');
+      return (itemSkillId === skillId);
+    }.bind(this));
```
begedin: Can ditch this bind in favor of an arrow function.

begedin:
Looks good to me. You can decide on that bind vs arrow function thing.

PR #219, opened 2016-06-19
Fix post empty state messages

### LINKED ISSUE #209: Fix post empty state messages
If filtered, reads: "There are currently no posts of any type created for this project."

Should also change and better design the completely empty state.

JoshSmith:
Closes #209.


ethantanner:
👍

PR #220, opened 2016-06-20
Minor UI improvements

### LINKED ISSUE #206: Possible @@ on user profile
If user includes @ in their twitter handle then their profile will show @@. This can be fixed by prepending the input with an @ and by removing any @ from the twitter handle before storing or before displaying.

image

image

JoshSmith:
Just some minor improvements to the UI: buttons, forms, etc.

Closes #206.


ethantanner:
:+1:

PR #224, opened 2016-06-20
Fix and cleanup flash message styles

### LINKED ISSUE #221: Flash messages no longer have a background color
Success flash messages appear not to have a background color anymore. This is a regression that needs fixed.

begedin:
Closes #221 

I reorganized the style a bit there, so It's easier to figure out in the future.

I also added a left padding to the regular flash message container, since as it was, the text was at the far left edge of the page and looked weird.

Fixed look:

image


JoshSmith:
👍

PR #225, opened 2016-06-20
Added filtering by post status

### LINKED ISSUE #214: Add ability to filter to open/closed issues

begedin:
Closes #214 

I generalized the `post-filter-dropdown` component a bit, so we can use it both for type and status filters. The filter behavior itself, though, which is defined at controller level is semi-exclusive, meaning that the filter can be either open or closed (or neither). It cannot be both open and closed at the same time. Considering that, there is no "all" button to remove status filters from the page. Instead, we simply select another filter, or click to dismiss the current filter.

### DIFF
--- a/app/adapters/post.js
+++ b/app/adapters/post.js
@@ -20,6 +20,12 @@ export default ApplicationAdapter.extend({
       delete query.postType;
     }
 
+    // we don't want to send the status parameter to the API if it does not
+    // have a proper value
+    if (Ember.isEmpty(query.status)) {
+      delete query.status;
+    }
+
     // projectId is part of the url in `projects/:projectId/posts`, so we
     // do not want to see it in the query as well
     if (query.projectId) {
--- a/app/components/post-filter-dropdown.js
+++ b/app/components/post-filter-dropdown.js
@@ -4,16 +4,16 @@ export default Ember.Component.extend({
   classNames: ['dropdown-menu'],
   tagName: 'ul',
 
-  click: function() {
+  click() {
     this.sendAction('hide');
   },
 
   actions: {
-    filterByType: function(type) {
-      this.sendAction('filterByType', type);
+    filterBy(filter) {
+      this.sendAction('filterBy', filter);
     },
 
-    hide: function() {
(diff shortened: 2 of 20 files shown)


begedin:
Requires code-corps/code-corps-api#321

JoshSmith:
A couple notes on this:

- Going to `/posts` is the default and the open status is the default. 👍 
- Clearing the filter changes the URL to `/posts?status=null`. This is maybe made moot if we implement following changes.
- Think that having a dropdown for this isn't as good as having just a couple links side-by-side that are `.active` or not.
- Don't think these should be added then to the list of filters displayed.

What do you think?

begedin:
Links are something I considered, but the reason I went with dropdowns and dismissable filters once they are active was that we sill probably want to have the option to show all posts (both open and closed). In that case, we'd need either

* 3 exclusive buttons - "open", "closed", "both/all/any"
* 2 buttons that are toggleable, if just one status is enabled, show just that one status, if both are enabled, or both are disabled, show all posts and remove filter

I feel that the dropdown is more consistent with what we have, which is why I went with it. Having "status=null" in the URL is a bug, which I need to figure out.

JoshSmith:
I'm not sure if I agree that we want to have an option to show both open and closed posts. What would be the use case for that? Have you ever wanted it on GitHub?

begedin:
@JoshSmith I have used it before, simply to track down a specific issue for which I'm not sure if it's closed or not. It's a matter of removing the `is:open` part of the filter on github. Not sure how generally useful that is, though, but maybe we should look at how github does it either way.

They have a "closed" and "open" button as you said, as well as a textbox for custom filtering. We could go with the buttons now, and add the textbox in a future version.

JoshSmith:
@begedin I've simplified a lot and will push up soon.

There are some things we need to change regardless.

JoshSmith:
Here's a quick demo:

hannah:
Looks good to me! 👍 🎉

JoshSmith:
Yay. Thanks @hannah!

PR #242, opened 2016-06-21
Skill selection works by clicking again

### LINKED ISSUE #229: Start Skills selection only works with enter key, should be clickable
click-example

I think this may be a regression bug, as I feel it used to work. When I type a valid skill in to add to my profile, I must hit the "Enter" key and cannot select the option from the drop-down with my mouse.

JoshSmith:
Closes #229.

Needed to do `mousedown` to occur before the `blur` event. `click` happens too late in the cycle.


begedin:
:+1:

Issue #233, opened 2016-06-21
"post type" text cut off when making new post


ethantanner:
When creating a new post the post type dropdown text is cut off. See attached screenshot
![image](https://cloud.githubusercontent.com/assets/6895721/16221042/35c27004-3756-11e6-9226-a7bd30260dcb.png)

Found on Chrome 51.0.2704.103 on Windows 10

JoshSmith:
Screw cross-compatibility! Buy a Mac and download Chrome or something! Get off my lawn!

JoshSmith:
Question is whether we should really be overriding the dropdown, or if there's a better non-terrible way to do this. Will accept any PR that makes CSS less terrible.

PR #244, opened 2016-06-22
Reduce post-type dropdown vertical padding

ethantanner:
Fixes issue [233](https://github.com/code-corps/code-corps-ember/issues/233).

I reduced the vertical padding on the post-type dropdown.


JoshSmith:
I'm not sure if this necessarily solves the issue on Chrome. The design then looks a little odd for me on my screen. Is there something else that could be causing this?

ethantanner:
Can you post a screenshot of what you are seeing? For me reducing the vertical padding just uncovers the text and keeps it vertically centered. Here is how it looks on my machine with chrome:

Without changes:
image

With changes:
image

JoshSmith:
I guess this isn't a problem somehow? Does it even need padding?

ethantanner:
We don't need vertical padding but we definitely need some left padding. So we could take the vertical padding down to 0.

JoshSmith:
Okay, then let's kill the vertical padding altogether if the height is doing it.

JoshSmith:
@ethantanner can you rebase and squash? https://github.com/code-corps/code-corps-ember/blob/develop/CONTRIBUTING.md#before-we-can-merge

PR #248, opened 2016-06-23
Add basics for UI for approving users to projects

### LINKED ISSUE #167: Add UI for approving users for project
An organization needs to be able to approve new users. I imagine this would list users, their photo, their username, and their name. There would be accept and reject buttons. Ideally we would also list out a user's roles and skills, highlighting those that the project has.

JoshSmith:
Closes #167.

This adds a UI for approving users:


begedin:
👍 good to go IMO

PR #256, opened 2016-06-23
Add ability to open and close posts

### LINKED ISSUE #213: Add ability to close or reopen a post
To the left of the comment button should be a "Close post" button when open, and an "Reopen post" button when closed. It should be of the `clear` button style. You can only close or open a post if you're the user who created the post, or you're at least an admin on the project.

begedin:
Closes #213 

The button is put next to the "create comment" button on the post page. I made the `create-comment-form` component yield to it, so it can be placed from the `post.hbs` template directly, since that's the level that's more related to it.

Test's cover this behavior completely.

I also added a new, `post` ability, which allows us to quickly access other posts permissions. While doing this, I realized we can simply inject services into our ability class, meaning we don't have to do stuff like:

```Handlebars
{{#if (can "manage organization" organization membership=credentials.currentUserMembership) }}
```

Instead, we simply inject the `credentials` service into our `organization` ability and we get access to  `credentials.currentUserMembership` directly.

Then, our template code becomes:

```Handlebars
{{#if (can "manage organization" organization) }}
```

We can even go a step further and do 

```Handlebars
{{#if (can "manage organization") }}
```

since our `credentials` service also provides the current organization.

I'll create an issue for this suggested change/refactor.

### DIFF
--- a/app/abilities/post.js
+++ b/app/abilities/post.js
@@ -1,0 +1,21 @@
+import Ember from 'ember';
+import { Ability } from 'ember-can';
+
+export default Ability.extend({
+  credentials: Ember.inject.service(),
+  currentUser: Ember.inject.service(),
+
+  post: Ember.computed.alias('model'),
+
+  userIsAuthor: Ember.computed('post', 'currentUser.user', function() {
+    let postUserId = this.get('post.user.id');
+    let currentUserId = this.get('currentUser.user.id');
+
+    return postUserId === currentUserId;
+  }),
+
+  membership: Ember.computed.alias('credentials.currentUserMembership'),
+  userIsAtLeastAdmin: Ember.computed.or('membership.isAdmin', 'membership.isOwner'),
+
+  canEdit: Ember.computed.or('userIsAuthor', 'userIsAtLeastAdmin')
+});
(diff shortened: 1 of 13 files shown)


app/abilities/post.js
```diff
@@ -0,0 +1,21 @@
...
+    let postUserId = this.get('post.user.id');
+    let currentUserId = this.get('currentUser.user.id');
+
+    return postUserId === currentUserId;
+  }),
+
+  membership: Ember.computed.alias('credentials.currentUserMembership'),
+  userIsAtLeastAdmin: Ember.computed.or('membership.isAdmin', 'membership.isOwner'),
+
+  canOpenClose: Ember.computed.or('userIsAuthor', 'userIsAtLeastAdmin')
```
JoshSmith: I would change to `canEdit`. This will be reused in the future.

app/components/post-status-button.js
```diff
@@ -0,0 +1,22 @@
+import Ember from 'ember';
+
+export default Ember.Component.extend({
+  classNames: ['post-status-button'],
+  tagName: 'span',
+
+  postIsOpen: Ember.computed.equal('post.status', 'open'),
```
JoshSmith: Would probably just say `isOpen`.

JoshSmith:
My comments are all I have. Looks good. 👍 

Can you make changes if you agree and rebase?

JoshSmith:
👍

PR #260, opened 2016-06-24
Remove unnecessary afterModel in routes/organizations/slugged-route

### LINKED ISSUE #258: In app/routes/organizations/slugged-route.js, we're setting the organization twice.
In the [`sluged-route` `model` and `beforeModel` hooks](https://github.com/code-corps/code-corps-ember/blob/develop/app/routes/organizations/slugged-route.js), we're setting the organization, for the `credentials` service, twice:

```Javascript
  model(params) {
    return this.store.queryRecord('slugged-route', {
      slug: params.slugged_route_slug
    }).then((result) => {
      let organization = result.get('owner');
      return this.get('credentials').setOrganization(organization).then(() => {
        return organization;
      });
    });
  },

  afterModel(organization) {
    this.get('credentials').setOrganization(organization);
    return this._super(...arguments);
  },
```

This probably doesn't cause problems, but is unnecessary, probably causes a very slight performance impact and can be confusing.

JoshSmith:
Closes #258.

### DIFF
--- a/app/routes/organizations/slugged-route.js
+++ b/app/routes/organizations/slugged-route.js
@@ -14,9 +14,4 @@ export default Ember.Route.extend(AuthenticatedRouteMixin, {
       });
     });
   },
-
-  afterModel(organization) {
-    this.get('credentials').setOrganization(organization);
-    return this._super(...arguments);
-  },
 });


hannah:
👍

PR #268, opened 2016-06-24
Add mention prefetching and coalesce find requests

JoshSmith:
This adds mention prefetching to remove N+1 requests for comments when loading a post. This assumes comment mentions are included in the response from the server.

### DIFF
--- a/app/adapters/application.js
+++ b/app/adapters/application.js
@@ -9,7 +9,7 @@ export default DS.JSONAPIAdapter.extend(DataAdapterMixin, {
   // TODO: Fix this once many-to-many lands in ember-cli-mirage
   // but as of right now there is no way to make this work for
   // all possible cases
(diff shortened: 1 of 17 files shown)


JoshSmith:
Still need to fix post details tests and the comments acceptance tests.

begedin:
I honesty didn't expect the tests here to work as easily. I expected rewriting of mirage request handler functions to get it working.

Do we not need to enable find request coalescing in mirage as well, though?

http://www.ember-cli-mirage.com/docs/v0.2.x/shorthands/#get-shorthands

I'm slightly worried our tests pass when they shouldn't.

I would put the coalesce option to our array of objects route handlers and see if the tests still pass.

JoshSmith:
@begedin added an explicit coalesce and still works. I was surprised, too.

PR #271, opened 2016-06-25
Fix bug with user skills service

### LINKED ISSUE #270: When the user skills service is undefined, find cannot work
Trying to do `userSkills.find` is simply not going to work. This causes errors on `/projects`, for example, when logged out and there are no user skills.

JoshSmith:
Closes #270.


JoshSmith:
This tests the bare minimum of being able to at least visit `/projects`, which you cannot right now. We can expand on this acceptance test later, but this is working for now.

ethantanner:
👍

PR #278, opened 2016-06-26
Add demo

JoshSmith:
This adds a demo to the landing page. Not ideal. Wish that if you clicked on it that it would present a model. But it's the best I've got on short notice.


JoshSmith:
These tests need to just be made passing. Don't care how at this stage since this is kind of throwaway code at some point.

PR #287, opened 2016-06-27
Limit number of members visible on project card to 7

### LINKED ISSUE #281: All members of a project are displayed on the project
image
There needs to be a limit, if there were 1000 members then 1000 photos would be shown.

begedin:
Closes #281

If there is more than 7 members, we display a "+ # more", where "#" is the number of obscured members.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -19,3 +19,5 @@ testem.log
 .env
 .env.deploy.staging
 .env.deploy.production
+
+.DS_Store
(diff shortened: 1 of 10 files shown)


begedin:
Rebased and fixed a jshint error. Tests should be passing now.

JoshSmith:
@begedin tests are not passing on this.

JoshSmith:
👍

PR #289, opened 2016-06-27
Used ember-concurrency to throttle automcomplete search requests to the API

### LINKED ISSUE #283: Use ember-concurrency for skill input typeahead
Our skill input typeahead makes too many requests without waiting for previous requests to finish. We should use ember-concurrency for this, so that we only update with the most recent request.

begedin:
Closes #283 

Keep in mind that this will drop all requests within 500ms of the first one, so if the user starts typing in "ruby", the request for "r" will likely be the only one we send. 

I'm not really sure if 500ms is a good value here either, but we can tweak that.

I would love for a way to cancel ajax requests and only send the latest one, but while that technically still possible, the cancelled request will still hit and be handled by the API.


begedin:
I was too quick with this one. It can be done better, but I also need to figure out the tests. Marking as "Needs work"

thomascchen:
@begedin Looks like this PR has been open for a while. What are the next steps?

thomascchen:
Thanks @begedin. In that case, do you mind closing this PR?

PR #306, opened 2016-07-05
Split name into first and last

JoshSmith:
This matches up to the API PR that splits the user's name into first and last name.


alexwilkinson:
Looks good to me! The only generally thing would be I wouldn’t have expected the CSS changes to necessarily be in this commit—more specifically, the `.input-group input` changes. I only mention that because in this context it all goes together, but someone reviewing months down the road might wonder the reasoning behind the CSS change.

But minor point, 🚢

Issue #223, opened 2016-06-20
Add YUIDoc documentation across the app


JoshSmith:
This is a relatively big project but can be tackled in small parts. Would accept pull requests for any number of small pieces, even as low as a single function being documented would be super helpful.

JoshSmith:
@WenInCode can we add a list here of what all has been doc'ed and what hasn't, or a reference to Inch or something? Would be good to make it easier to know where to jump yet, or maybe create some separate issues for the remaining pieces.

WenInCode:
@JoshSmith Sure I'll add  a list of things that are left.

WenInCode:
# List of files to be documented
## Abilities
- [ ] organization.js
- [ ] post.js

## Adapters
- [ ] application.js
- [ ] comment.js
- [ ] post.js
- [ ] project.js
- [ ] slugged-route.js
- [ ] user.js

## Authenticators
- [ ] oauth2.js

## Authorizers
- [ ] oauth2.js

## Components
- [ ] categories-list.js
- [ ] category-item.js
- [ ] code-theme-selector.js
- [ ] comment-item.js
- [ ] create-comment-form.js
- [ ] demo-categories.js
- [ ] demo-category-item.js
- [ ] demo-posts.js
- [ ] demo-projects.js
- [ ] demo-skills.js
- [ ] drag-zone.js
- [ ] editor-with-preview.js
- [ ] error-formatter.js
- [ ] error-wrapper.js
- [ ] flash-messages.js
- [ ] image-drop.js
- [ ] landing-subsection.js
- [ ] login-form.js
- [ ] member-list-item.js
- [ ] navigation-menu.js
- [ ] organization-header.js
- [ ] organization-member.js
- [ ] organization-profile.js
- [ ] organization-settings-form.js
- [ ] organization-settings-menu.js
- [ ] organization-settings.js
- [ ] pager-control.js
- [x] post-comment-list.js
- [x] post-details.js
- [x] post-filter-dropdown.js
- [x] post-filter-type.js
- [x] post-header.js
- [x] post-item.js
- [x] post-new-form.js
- [x] post-status-button.js
- [x] post-status.js
- [x] post-title.js
- [ ] progress-bar-container.js
- [x] progress-bar.js
- [ ] project-card-members.js
- [ ] project-card-skills.js
- [ ] project-card.js
- [ ] project-categories-list.js
- [ ] project-category-item.js
- [ ] project-container.js
- [ ] project-details.js
[...]

PR #313, opened 2016-07-11
Add yuidoc, and example documentation

WenInCode:
# What's in this PR?
- This adds a `yuidoc.json` configuration file for documenting the ember app. See instructions for generating and serving the docs below. I am new to using yuidoc, but I think I covered the required bases in this configuration file.
- This also includes some basic documentation for one of the components `user-menu`
screen shot 2016-07-11 at 12 08 50 am

## Reasons for decisions
- I decided to go with plain `yuidoc` rather than the ember add-on, as I don't feel the add-on really gives you much more. The benefit to using the `ember-cli-yuidoc` add-on is that you can run `ember ember-cli-yuidoc` to generate the docs without installing `yuidoc` globally. Also, there is a nifty flag `--docs` on `ember s` but I found it wasn't always reliable between environments. That being said, I am open to suggestions/feedback on this decision.
- What we use for documenting things in ember wasn't *clear* to me, and that comes from most ember apps not needing/or simply not using docs. I decided to label each component as a `class` that has `properties` & `methods` rather than using yuidocs `element`, as I found it wasn't as nice for documenting each method within (I think it pretty much only lets you document attributes within elements). Once again, open to feedback on this. See example docs.

# How to get set-up
- pull down this branch
- `npm install -g yuidocjs`
- To compile a new set of docs run `yuidoc -c yuidoc.json` (using the projects configuration file)
- To serve docs run `yuidoc --server [your port of choice]` and visit `localhost:[your port of choice]/`

# References
- #223 
- [yuidoc](http://yui.github.io/yuidoc/)

### DIFF
--- a/STYLEGUIDE.md
+++ b/STYLEGUIDE.md
@@ -10,6 +10,7 @@
 
 + [Ember models](#ember-models)
 + [Ember modules](#ember-modules)
++ [Documentation](#documentation)
 
 ### Ember models
 
@@ -129,6 +130,34 @@ export default Ember.Component.extend({
   }
 });
 ```
+
+### Documentation
+
+Follow the syntax outlined on the [YUIDoc Syntax Page](http://yui.github.io/yuidoc/syntax/index.html). As YUIDoc was built for `Javascript`, not for `Ember` specifically, we've set our own rough guidelines for this `Ember` project.
+
+* `Components/Services/Routes/Models` will each be a module. All extensions of them should define which `@module` it belongs to.
+* All extensions of these modules are defined as `@class`
+* Actions are defined as `@method`, but should be specified that they are actions in the description.
+* Properties & Computed Properties should be defined as `@property`
+* Components should have an example of default usage
+
+
+        // some-component.js
+        
+        /**
+          @module Components
+          @extends Ember.Component
+          @class some-component
+        
+          Component description
+        
+          ## default usage
+          ```handlebars
+          {{some-component attrs=attrs}}
+          ```
+         */
+        export default Ember.Component.extend({
+
 
 ## HTML and Handlebars
 
--- a/README.md
+++ b/README.md
@@ -68,6 +68,23 @@ You should refer to our [style guide](STYLEGUIDE.md) for writing Ember.js, HTML/
 
 The app uses Circle for continuous integration and deploys automatically to a staging server when merging into `develop` and to production when merging into `master`.
 
+## Documentation
+
+The Code Corps Ember application uses [YUIDoc](http://yui.github.io/yuidoc/) for documentation. When contributing to the documentation please follow the [YUIDoc syntax](http://yui.github.io/yuidoc/syntax/index.html) and our [style guide](STYLEGUIDE.md) for application specific documenation standards.
+
+### Generating Documentation
+
+* `npm install -g yuidocjs`
+* `yuidoc -c yuidoc.json`
+
+The documentation will be generated in the `/docs` folder.
+
+### Serving Documentation
+
+* Generate the docs
+* `yuidoc --server [your port of choice]`
+* Visit `localhost:[your port of choice]` in your browser.
+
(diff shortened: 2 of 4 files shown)


WenInCode:
It may be better to leave the docs out of github & only generate them locally -- not sure though.

JoshSmith:
Maybe we could use https://github.com/offirgolan/yuidoc-ember-theme?

WenInCode:
Ya I was looking into that last night as well. I'll add it in today.

noslouch:
Will this also document Services? If so, is there another YUI abstraction besides Class that can be used to describe them? I can imagine a world where the navigation-menu component and service get mixed up.

WenInCode:
@noslouch It should be used to document all things ember. Though YUIdoc isn't built *for* ember so there doesn't seem to be a great set of tags for services/components/etc. My suggested solutions is to use classes and let the modules be Services/Components/Controllers/Routes. Though I should say I have not used YUIdoc in an ember app before so I am very open to other techniques.

JoshSmith:
Yeah, from my reading it looks like it's all supposed to be documented as `Class` right now, which is pretty meh. I wish it weren't so.

noslouch:
Have either of you seen the docs for ember-simple-auth? 

Or what about what the actual EmberJS guides use?

JoshSmith:
@noslouch looks like YUIdoc for `ember-simple-auth`: https://github.com/simplabs/ember-simple-auth/blob/bed56395e8ce0e376e28b6ea9a62a4e26585d875/ember-cli-build.js#L34

WenInCode:
@noslouch ember uses yuidoc as well

JoshSmith:
@WenInCode can you add instructions to the README for how to generate/view yuidoc locally?

As a later step, we could even make a Docker container similar to the one I have on the API side for generating docs with `aglio`.

WenInCode:
ya that's a good idea - I'll update the readme sometime today

STYLEGUIDE.md
```diff
@@ -130,6 +131,33 @@ export default Ember.Component.extend({
...
+
+/**
+  @module Components
+  @extends Ember.Component
+  @class some-component
+
+  Component description
+
+  ## default usage
+  [```handlbars]
```
WenInCode: Is there anyway to escape these within another code block? For now I just used `[]`.

STYLEGUIDE.md
```diff
@@ -130,6 +131,33 @@ export default Ember.Component.extend({
 });
 ```
 
+### Documentation
+
+Follow the syntax outlined on the [YUIDoc Syntax Page](http://yui.github.io/yuidoc/syntax/index.html). As YUIDoc was built for javascript, not for ember specifically, we have defined our own paradigm.
+
+* `Components/Services/Routes/Models` will each be a module. All extensions of them should define which `@module` it belongs to.
+* All extensions of these modules are defined as `@class`
+* Actions are defined as `@method`, but should be specified that they are actions in the description.
```
WenInCode: This is a bit unfortunate that actions can't be defined separately. Let me know if there are any suggestions here.

STYLEGUIDE.md
noslouch: I think there's a typo: `handlbars` instead of `handlebars`.

STYLEGUIDE.md
```diff
@@ -130,6 +131,34 @@ export default Ember.Component.extend({
 });
 ```
 
+### Documentation
+
+Follow the syntax outlined on the [YUIDoc Syntax Page](http://yui.github.io/yuidoc/syntax/index.html). As YUIDoc was built for javascript, not for ember specifically, we have defined our own paradigm.
```
JoshSmith: Minor, but would change to `JavaScript` and `Ember`.

Also maybe:

> we've set our own rough guidelines for this Ember project.

JoshSmith:
I think this is a good start so LGTM. 👍

Issue #141, opened 2016-05-31
Rewrite tests with page objects


JoshSmith:
http://ember-cli-page-object.js.org/docs/v1.4.x/components

Lets us do:

```
const page = PageObject.create({
  visit: visitable('/'),

  username: fillable('#username'),
  password: fillable('#password'),
  submit: clickable('button'),
  error: text('.errors')
});

test('my awesome test', function(assert) {
  page
    .visit()
    .username('admin')
    .password('invalid')
    .submit();

  andThen(() => {
    assert.equal(page.error, 'Invalid credentials');
  });
});
```

green-arrow:
@JoshSmith - I have this started with the `contributors` acceptance test. Would you prefer one large pull request with a checklist of each test that needs migrated, or would you prefer a checklist here and multiple PRs referencing this issue?

green-arrow:
@JoshSmith - so my apologies on this. I accidentally pushed some page object stuff to `develop` before you merged my other PR for the `bower` stuff. So the page object library and pattern is already in `develop` 😅 

This is a checklist of tests that need to be migrated to the page object pattern:

- [ ] Code theme
- [x] Contributors
- [ ] Login
- [ ] Logout
- [ ] Navigation
- [ ] Onboarding
- [ ] Organization projects
- [ ] Organization settings profile
- [ ] Organization
- [ ] Post comments
- [ ] Post creation
- [ ] Post editing
- [ ] Profile
- [ ] Project about
- [ ] Project settings
- [ ] Project
- [ ] Projects
- [ ] Settings profile
- [ ] Signup
- [ ] Slugged route
- [ ] Team

eablack:
A few of these done in https://github.com/code-corps/code-corps-ember/issues/381

JoshSmith:
🎉 all done!

PR #315, opened 2016-07-11
Acceptance Test Page Objects: Code Theme, Login, Logout, Navigation

noslouch:
The title says it all. Knocks off the first few items from #141.

This shows a good refactor path forward for tests and how we can abstract components on pages into Page Object Components, which can then be used in integration tests.


JoshSmith:
@noslouch is this ready for review? If so, can you mark as `awaiting review`?

JoshSmith:
👍 LGTM.

PR #317, opened 2016-07-12
Remove typekit and add font size variables

JoshSmith:
This removes Typekit, adds Google Fonts, and places font size variables across the app.


noslouch:
LGTM

noslouch:
:shipit:

PR #320, opened 2016-07-13
hastens acceptance tests with shorter timeouts in testing

noslouch:
sets a few timeouts to 0 if we're testing, make those acceptance tests go :rocket:


JoshSmith:
Nice! 👍

JoshSmith:
:shipit:

PR #321, opened 2016-07-13
fixes flaky input tests by manually triggering keydown events

noslouch:
inputs are weird. inputs in test environments are weirder. this forces
a keydown event on the inputs so the compnent runs its handler and
updates the proper props.

I think that `keyUp` is technically the proper event to use in the
component, but let's get a solid pattern first and think about that
refactor later, if at all.

also, this PR opts for a simpler syntax to trigger the focus event: `focus()`.

### DIFF
--- a/tests/integration/components/user-skills-input-test.js
+++ b/tests/integration/components/user-skills-input-test.js
@@ -72,10 +72,11 @@ test('it fetches results when changing the input', function(assert) {
   assert.expect(5);
   this.render(hbs`{{user-skills-input query=query}}`);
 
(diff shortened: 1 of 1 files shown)


JoshSmith:
I think there was a reason why I did `keyDown` and not `keyUp`, but not documented, so I don't know.

JoshSmith:
👍

JoshSmith:
You can merge yourself when these tests pass. I think you should have a commit bit already.

noslouch:
already out of date! what a nice, fast-moving project 🎆

PR #322, opened 2016-07-14
Update squashing docs

JoshSmith:
This updates the squashing docs to reference any editor, not just Sublime. Also breaks them out into their own separate area.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -35,43 +35,4 @@ Before We Can Merge
 
 If you've had a pull request reviewed and accepted, congratulations! Before we can merge your changes, we'll need you to rebase off `origin/develop` and squash your commits into one. This will give us a cleaner git history.
 
-Never done this before? No problem. We'll walk you through it, and you can read [a deeper guide about rewriting history to understand more](https://git-scm.com/book/en/v2/Git-Tools-Rewriting-History).
-
-On your command line you'll want to do:
-
-```
-git rebase -i origin/develop
-```
-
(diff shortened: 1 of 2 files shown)


CONTRIBUTING.md
```diff
@@ -35,43 +35,4 @@ Before We Can Merge
 
 If you've had a pull request reviewed and accepted, congratulations! Before we can merge your changes, we'll need you to rebase off `origin/develop` and squash your commits into one. This will give us a cleaner git history.
```
noslouch: Potentially change to "rebase onto" to match how it's referred to elsewhere (like in the universe)

docs/SQUASHING.md
```diff
@@ -0,0 +1,42 @@
+# How do I "squash" my commits?
+
+Before you start, [make sure your text editor is associated with Git](https://help.github.com/articles/associating-text-editors-with-git/) to not close without a response, otherwise you'll see some weird behavior.
+
+On your command line, run:
```
noslouch: The first step should probably be git fetch --all so their refs are updated

noslouch:
If this is really a first timer's guide, should we also walk through working off a fork and setting up two different remotes? One that points to code-corps and another that points to their own repo?

JoshSmith:
@noslouch that makes sense. Tackle that in a separate issue maybe?

noslouch:
yup, another issue sounds fine for this kind of thing

JoshSmith:
@noslouch do you want to open that?

noslouch:
@JoshSmith #326

PR #327, opened 2016-07-15
Onboarding Test refactor

noslouch:
another ✔️  for #141

### DIFF
--- a/tests/acceptance/onboarding-test.js
+++ b/tests/acceptance/onboarding-test.js
@@ -2,6 +2,8 @@ import Ember from "ember";
 import { module, test } from 'qunit';
 import startApp from '../helpers/start-app';
 import { authenticateSession } from 'code-corps-ember/tests/helpers/ember-simple-auth';
+import onboardingPage from '../pages/onboarding';
+import indexPage from '../pages/index';
 
 let application;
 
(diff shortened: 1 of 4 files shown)


JoshSmith:
Boy, this is so much nicer than what we had before. So worth the effort.

tests/pages/components/user-skills-input.js
```diff
@@ -0,0 +1,24 @@
+import {
+  fillable,
+  triggerable,
+  text,
+  clickable,
+  collection
+} from 'ember-cli-page-object';
```
JoshSmith: Do we want to alphabetize these?

tests/pages/onboarding.js
```diff
@@ -0,0 +1,61 @@
+import {
+  create,
+  attribute,
+  clickable,
+  fillable,
+  collection,
+  text,
+  hasClass,
+  visitable,
+} from 'ember-cli-page-object';
```
JoshSmith: Same here re: alpha.

tests/pages/onboarding.js
```diff
@@ -0,0 +1,61 @@
...
+    scope: '.user-skills-list',
+    itemScope: 'button',
+    item: {
+      text: text(),
+      click: clickable()
+    }
+  }),
+
+  userSkillsInput,
+  navMenu
```
JoshSmith: What are these doing? Just including the skills input and nav menu? Can we document that clearer?

tests/pages/onboarding.js
```diff
@@ -0,0 +1,61 @@
...
+
+});
```
JoshSmith: Would remove that empty space.

JoshSmith:
I just have some minor ordering/organizational concerns here. Do we want to introduce some explicit style guides here? Maybe start a discussion around what that should look like?

Also should maybe consider using something like [ember-suave](https://github.com/DockYard/ember-suave).

noslouch:
On board with both points. 

I don't really have a pref when it comes to ordering attrs, so if you want to sketch something out in an issue I can enforce here and refactor the others

JoshSmith:
@noslouch would you mind opening the issue and assigning it to me? I can then make some changes to it.

noslouch:
@JoshSmith updated import order  and cleaned up that trailing whitespace. I guess any discussion that happens in #330 can find their way into a ruleset for `ember-suave`

tests/acceptance/onboarding-test.js
```diff
@@ -38,99 +40,99 @@ test('A user can onboard as expected', (assert) => {
...
-    fillIn('input[name="lastName"]', 'Smith');
+    assert.ok(onboardingPage.startButton.isDisabled, 'start button disabled by default');
+    onboardingPage.firstName('Josh');
+    onboardingPage.lastName('Smith');
   });
 
   andThen(() => {
-    assert.equal(find('.start-actions button').is(':disabled'), false);
-    click('.start-actions button');
+    assert.notOk(onboardingPage.startButton.isDisabled, 'start button is enabled');
```
npendery: Not sure if this just makes more sense in my head or not, but should we add an `isEnabled` attribute to `startButton` and just `assert.ok` that it `isEnabled` instead of `assert.notOk ` that it `isDisabled`?

tests/acceptance/onboarding-test.js
```diff
@@ -38,99 +40,99 @@ test('A user can onboard as expected', (assert) => {
...
   andThen(() => {
-    assert.equal(find('.start-actions button').is(':disabled'), true);
-    click('.category-item button');
+    assert.ok(onboardingPage.startButton.isDisabled, 'start button is disabled');
+    onboardingPage.clickCategoryItem();
   });
 
   andThen(() => {
-    click('.start-actions button');
+    onboardingPage.startButton.click();
```
npendery: Should there be another assertion before this `click` that `startButton` is clickable?

tests/acceptance/onboarding-test.js
noslouch: Could do, but I'm not exactly sure on the syntax to bundle this up into a page object prop. the `attribute` selector doesn't seem to select a value, so I couldn't do `isEnabled: attribute('disabled', false)`, unless I can just make it CP?

tests/acceptance/onboarding-test.js
noslouch: sure

tests/acceptance/onboarding-test.js
npendery: Ah gotcha. Yeah I'm looking over the docs for ember-cli-page-object more and dont see anything that would work better than what you've got. Just thought there would be something...oh well!

npendery:
Looks great to me! Good job!

JoshSmith:
@noslouch is this good to review again? Good to merge? If so, can you squash? Thanks!

noslouch:
Yeah this is ready. I'll merge shortly!

PR #329, opened 2016-07-15
Add Documentation for a few post components

WenInCode:
Added some documentation for a few of the `post-` components. 
- `post-meta`
- `post-new-form`
- `post-title`
- `post-status-button`

Apparently I still haven't got this rebase thing right yet... There are a bunch of merge commits I must have pulled in.

### DIFF
--- a/app/components/post-meta.js
+++ b/app/components/post-meta.js
@@ -1,5 +1,0 @@
-import Ember from 'ember';
-
-export default Ember.Component.extend({
-  classNames: ['post-meta']
-});
--- a/app/components/post-new-form.js
+++ b/app/components/post-new-form.js
@@ -1,18 +1,49 @@
 import Ember from 'ember';
 
+/**
+  The post-new-form component is used for creating new posts. It includes the
+  post type, title, editor and preview
+
+  ## default usage
+
+  ```handlebars
+  {{post-new-form post=newPost savePost='savePostMethod'}}
+  ```
+
+  @class post-new-form
+  @module Component
+  @extends Ember.Component
+ */
 export default Ember.Component.extend({
   classNames: ['post-new-form'],
   classNameBindings: ['post.postType'],
   tagName: 'form',
 
+  /**
+    @property credentials
+    @type Ember.Service
+   */
   credentials: Ember.inject.service(),
 
+  /**
+    Holds the placeholder messages for each post type: task, issue, idea.
+
+    @property placeholders
+    @type Object
+   */
   placeholders: {
     task: "How can you describe the steps to complete the task so anyone can work on it?",
     issue: "What issue needs resolved? If it's a bug, how can anyone reproduce it?",
     idea: "What's your idea? Be specific so people can give more accurate feedback.",
   },
 
+  /**
+    Returns which placeholder message to use from the placeholders property
+    based on the post type.
+
+    @property placeholder
+    @type String
+   */
   placeholder: Ember.computed('post.postType', function() {
     let postType = this.get('post.postType');
     if (postType) {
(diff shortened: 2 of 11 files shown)


app/components/post-meta.js
```diff
@@ -1,5 +1,18 @@
 import Ember from 'ember';
 
+/**
+  The post-meta component is a badge that presents a post's status
```
JoshSmith: Something is off here. I think the component is named incorrectly. It looks like we probably want to change `post-meta` to `post-status` and move the `post.status` binding into the component itself. This PR is small enough we can tackle that here if we'd like.

app/components/post-new-form.js
```diff
@@ -1,18 +1,50 @@
...
 
+  /**
+    @property credentials
+    @type Ember.Service
+   */
   credentials: Ember.inject.service(),
 
+  /**
+    Holds the different plcaeholder messages for each type of post
+    (task/issue/idea).
```
JoshSmith: `placeholder` is misspelled. Maybe reword to "for each post type: task, issue, idea, etc."

app/components/post-title.js
```diff
@@ -1,15 +1,52 @@
...
 
+  /**
+    @property currentUser
+    @type Ember.Service
+   */
   currentUser: Ember.inject.service(),
 
+  /**
+    Computed alias for `currentUserIsPostAuthor` that returns whether or not the current
+    user can edit the current post.
```
JoshSmith: I'm not sure if we should describe what it's doing but rather what it's behavior is. "Returns whether or not the current user can edit the current post" seems simpler and better.

app/components/post-title.js
```diff
@@ -1,15 +1,52 @@
...
+
+    @property canEdit
+    @type Boolean
+   */
   canEdit: Ember.computed.alias('currentUserIsPostAuthor'),
+
+  /**
+    Computed alias for the current user's ID.
+
```
JoshSmith: Same here regarding what it is.

app/components/post-title.js
```diff
@@ -1,15 +1,52 @@
...
+    @property currentUserId
+    @type Number
+   */
   currentUserId: Ember.computed.alias('currentUser.user.id'),
+
+  /**
+    Computed alias for the post author's ID.
```
JoshSmith: And here.

JoshSmith:
Very nice work here. Want to pair at some point on rebase and I can troubleshoot? Would help me learn what's at issue.

I have some minor points, so let me know what you think.

@noslouch what do you think about my points on documentation, too?

app/components/post-meta.js
WenInCode: That's a fair point. I think most of the components here were pretty clear what they were just by naming whereas this one took a small amount of digging. I'll look into it this evening.

WenInCode:
Good points here - I'll update this later tonight.

app/components/post-meta.js
JoshSmith: Yeah, I don't know why the naming was that way anymore, but just further reinforces the need for the good work you're doing here documenting it all. Poorly named code can't survive in the light of good documentation!

WenInCode:
@JoshSmith addressed feedback on docs -- awaiting review

app/components/post-status.js
```diff
@@ -0,0 +1,24 @@
...
+  ```handlebars
+  {{post-status post=post}}
+  ```
+
+  @class post-status
+  @module Component
+  @extends Ember.Component
+ */
+export default Ember.Component.extend({
+  classNames: ['post-meta'],
```
JoshSmith: I would rename this to match the `post-status`, which will likely mean a CSS change. Not sure if there's a `post-meta` scss file.

JoshSmith:
@WenInCode one minor change. Fix that and rebase/merge yourself. 👍

Issue #305, opened 2016-07-05
Change to dasherized attributes and relationships to reflect API changes


JoshSmith:
We need to remove both for `keyForAttribute` and `keyForRelationship` in both `app/serializers/application` and `mirage/serializers/application` to reflect changes in the API, once merged.

JoshSmith:
Blocked by https://github.com/code-corps/code-corps-api/issues/391.

begedin:
Note that we might need to do some mirage serializer changes for this as well.

JoshSmith:
Will be done in #363.

PR #331, opened 2016-07-15
Clean up mirage config, make it less reliant on our key transform method

begedin:
This was initially started as a fix for #305, but we marked that as low priority due to requiring a lot of work on the API side, so this is the part that doesn't affect current app behavior, but improves on the code quality of our mirage requests a decent amount.


app/mixins/can-animate.js
```diff
@@ -11,7 +11,6 @@ export default Ember.Mixin.create({
     rect = this.get('boundingClientRect');
     windowHeight = this.get('windowHeight');
     windowWidth = this.get('windowWidth');
-    console.log(rect.top, windowHeight);
```
JoshSmith: Forgotten console.log.

mirage/config.js
```diff
@@ -100,47 +100,22 @@ export default function() {
   ///////////
 
   // POST /comments
-  this.post('/comments', (schema, request) => {
-    let requestBody = JSON.parse(request.requestBody);
-    let attributes = requestBody.data.attributes;
-    let relationships = requestBody.data.relationships;
-
+  this.post('/comments', function(schema) {
+    let attrs = this.normalizedRequestAttrs();
```
begedin: Need to use a function callback instead of an arrow function, in order to get access to `this`. `normalizedRequestsAttributes` returns a hash containing exactly what we need to either create or update a mirage record.

Example, a comment hash will look like:

```JavaScript
{
  body: '',
  markdown: '',
  userId: ''
  projectId: ''
}
```

which is the structure we were creating manually before.

tests/acceptance/organization-settings-profile-test.js
```diff
@@ -67,20 +67,14 @@ test("it allows editing of organization profile", (assert) => {
...
     done();
 
-    return {
-      data: {
-        id: organization.id,
-        type: "organizations",
-        attributes: params
-      }
-    };
+    return this._getJsonApiDocForRequest(request, "organization");
```
begedin: I'm not extremely excited with using `_getJsonApiDocForRequest `, since it's clearly intended to be private, but it's the only function I can find that gives us the structure we need, so it's still preferable to creating the structure manually. Alternatively, we could just return `JSON.parse(request.requestBody)`.

begedin:
There's still some cleanup I can do in mirage config here, so this still needs some work, but the changes here already show what it's going to look like.

JoshSmith:
@begedin just wanted to follow up on this, since we're getting closer to the time we'll need to start making Ember changes to conform to Phoenix.

begedin:
@JoshSmith I'd hold off on doing anything here until the phoenix side is done. However, it should be good to go. Just need to update the branch.

JoshSmith:
@begedin do you anticipate this having to be rolled back? I just don't like an open PR that might need to either be closed or is fine and new changes could be addressed elsewhere.

begedin:
@JoshSmith Looking at this again, it should be good to merge as is. The only way we roll this back is if a future mirage version rolls back on its internal changes.

PR #333, opened 2016-07-19
add the rest of the post component docs

WenInCode:
# What's in this PR?
Added documentation for the rest of the `post-` components. 

## Notes
I realized while doing this that some of the post logic is duplicated between a few components. I'm thinking it might be a good idea to separate that logic into a mixin. I can do that in a separate PR.

Also, I noticed we are doing some saving of models directly from components. It may be more DDAU to move that save logic into something like the route.

### DIFF
--- a/app/components/post-comment-list.js
+++ b/app/components/post-comment-list.js
@@ -1,5 +1,18 @@
 import Ember from 'ember';
 
+/**
+  The post-comment-list component composes the list of comments for a post.
+
+  ## default usage
+
+  ```handlebars
+  {{post-comment-list comments=comments}}
+  ```
+
+  @class post-comment-list
+  @module Component
+  @extends Ember.Component
+ */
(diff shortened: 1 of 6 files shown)


JoshSmith:
@WenInCode agreed on both of your points. Do you want to open separate issues now to address?

Also, I would like our issue descriptions to move more into "Problem:, Solution:" descriptions. See [this](http://rfc.zeromq.org/spec:42/C4/) for reference.

JoshSmith:
👍 on this. Really good work.

This PR is a pretty good length, but may even suggest to open separate PRs for each component in the future. The more discrete our tasks and PRs, the easier it is for someone to quickly review it.

Always ask yourself: what's the most minor change I can make that's still appropriate for a PR, but someone could review on a coffee break? That empathy for other contributors' time constraints will help a lot going forward.

Merge at will!

PR #334, opened 2016-07-20
refactor organization-projects-test with page objects

noslouch:
Another one for #141. I'm iffy on whether we need `tests/pages/components/project-list.js` as a separate file, but we might re use it in other tests.


npendery:
Looks good to me! I don't see the harm in keeping `tests/pages/components/project-list.js`, I like having something such as this separated out to see where the source of the list comes from, especially since `organization-project` isn't itself a component.

thomascchen:
Thanks! I'm 👍  on keeping `project-list.js` as well.

PR #343, opened 2016-07-26
add category-item documentation

WenInCode:
# What's in this PR?
- Adds documentation for the `category-item` component
- Minor refactors

## References
Issue: #223

### DIFF
--- a/app/components/category-item.js
+++ b/app/components/category-item.js
@@ -1,47 +1,122 @@
 import Ember from 'ember';
+import { notEmpty } from 'ember-computed';
 
-export default Ember.Component.extend({
+const {
+  Component,
+  computed,
+  get,
+  inject,
+  set,
+} = Ember;
+
+const { service } = inject;
+
+
+/**
+ * `category-item` composes a category's icon, and description.
+ *
+ * ## default usage
+ *
+ * ```handlebars
+ * {{category-item category=category}}
+ * ```
+ *
+ * @class category-item
+ * @module Component
+ * @extends Ember.Component
+ */
+export default Component.extend({
   classNames: ['category-item'],
   classNameBindings: ['selected'],
   isLoading: false,
 
-  flashMessages: Ember.inject.service(),
-  userCategories: Ember.inject.service(),
+  /**
+   * @property flashMessages
+   * @type Ember.Service
+   */
+  flashMessages: service(),
 
-  selected: Ember.computed.notEmpty('userCategory'),
+  /**
+   * @property userCategories
+   * @type Ember.Service
+   */
+  userCategories: service(),
 
-  userCategory: Ember.computed('category', 'userCategories.userCategories', function() {
-    let category = this.get('category');
-    let userCategories = this.get('userCategories');
+  /**
+   * Returns if the category has been selected by the user.
+   *
+   * @property selected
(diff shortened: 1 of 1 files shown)


app/components/category-item.js
```diff
@@ -1,47 +1,121 @@
...
+  userCategories: service(),
 
-  selected: Ember.computed.notEmpty('userCategory'),
+  /**
+   * Returns if the category has been selected by the user.
+   *
+   * @property selected
+   * @type Boolean
+   */
+  selected: computed.notEmpty('userCategory'),
```
thomascchen: This is a minor stylistic thing, and I'm not dead set on it, but `notEmpty` could be further destructured when do the imports above like:

```
const {
  Component,
  computed,
  get,
  inject,
  set
} = Ember;

const { service } = inject;
const { notEmpty } = computed;
```

and then on line 50 we can just do:
```
selected: notEmpty('userCategory')
```

Maybe a little cleaner that way?

app/components/category-item.js
```diff
@@ -1,47 +1,121 @@
...
+     */
     addCategory(category) {
-      this.set('isLoading', true);
-      let userCategories = this.get('userCategories');
+      set(this, 'isLoading', true);
+      const userCategories = get(this, 'userCategories');
+
       return userCategories.addCategory(category).catch(() => {
-        let message = `An error occurred trying to add ${category.get('name')}.`;
+        const message = `An error occurred trying to add ${category.get('name')}.`;
```
thomascchen: This may be opening a can of worms, but have we landed on a general policy around defaulting to `const` vs `let`? I used to be a "constantly const" person but ever since I learned that [@rwjblue subscribes to the "liberal let" philosophy](http://discuss.emberjs.com/t/what-is-the-difference-between-const-let-var/9559/3?u=tomchen), I've default to `let`. Again, I'm not dead set on this...

JoshSmith: What is Dockyard's stance on this, out of curiosity? Would like to have ember-suave in here at some point, I think.

thomascchen: From their style guide: https://github.com/dockyard/styleguides/blob/master/engineering/javascript.md#assignment

```
Never use var. Prefer let to declare a block scope local variable; use const to declare a constant whose value can not be re-assigned in the given scope (global or local).
```

Not super explicit, but I think I would interpret that as defaulting to `let` and then using `const` only when there is a variable that we specifically want to guard against reassignment.

thomascchen:
@WenInCode This looks great to me. Could you address the issues above and update this branch with `develop`? Afterwards, I think it looks good to merge  👍

app/components/category-item.js
thomascchen: Also :+1: on `ember-suave`

app/components/category-item.js
noslouch: Computed macros can be imported directly on one line like so:

import { notEmpty } from 'ember-computed';

app/components/category-item.js
noslouch: 👍 on liberal let

WenInCode:
Great points - thanks all! Going to circle back and update the rest of my doc PRs to address these issues too

PR #351, opened 2016-07-28
add editor-with-preview docs

WenInCode:
# What's in this PR?
- add `{{editor-with-preview}}` component documentation
- minor refactors

## Decisions
- decided to remove the `_handleEdit` & `_handlePreview` methods as they weren't really abstracting anything.
- in the `_fetchPreview` method, we were setting the `fetchingPreview` property to `true` after concluding that there was nothing to preview. I am thinking this may have been a mistake; makes more sense for it to no longer be fetching.

## References
Progress on issue #223

### DIFF
--- a/app/components/editor-with-preview.js
+++ b/app/components/editor-with-preview.js
@@ -1,91 +1,219 @@
 import Ember from 'ember';
+import {
+  equal,
+  or,
+} from 'ember-computed';
 
-export default Ember.Component.extend({
+const {
+  Component,
+  computed,
+  get,
+  inject,
+  isEmpty,
+  set,
+} = Ember;
+
+const { service } = inject;
+
+/**
+  `editor-with-preview` composes the comment/post editor with the ability to
+  preview the content.
+
+  ## default usage
+
+  ```Handlebars
+  {{editor-with-preview
+    input=comment.markdown
+    isLoading=comment.isSaving
+    modifiedSubmit="save"}}
+  ```
+
+  @class editor-with-preview
+  @module Component
+  @extends Ember.Component
+ */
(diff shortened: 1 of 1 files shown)


WenInCode:
Sorry about leaving this in `needs work` for so long. It's been a busy few weeks. Should be ready for review now.

JoshSmith:
I wonder if we should be documenting private functions or not. The fact that private functions aren't typically documented in Elixir seems like a generally good idea.

Would love some thoughts here.

JoshSmith:
Other than that question, 👍

thomascchen:
Agree with not documenting private functions. I think it's more important to document the public API rather than its implementation. The implementation might change but the API could stay the same.

thomascchen:
Oops thought I was commenting on a different PR -- disregard the API reference! Still agree that we don't necessarily need to document private functions though 😄

WenInCode:
@JoshSmith @thomascchen While I do agree with the not documenting private methods, part of me is still on the fence. I guess my thinking was that this documentation is a little different than your every day API documentation. I guess I figured it might be a good idea to leave some implementation details in for those learning about the component. None the less, I also understand the other side.

Thanks for the comments, would you mind giving a quick double check on the changes? Thank you!

JoshSmith:
I'm a bit on the fence, too, but I think @thomascchen's point is well taken: public functions can have lower churn and therefore need those docs, whereas private has higher and documenting can be a bit of a fool's errand there.

Either way, I think with the amount of documentation we have left to do, going after primarily public functions should get us to "done" faster.

JoshSmith:
👍

PR #363, opened 2016-08-13
Update Ember for Phoenix

JoshSmith:
This updates some things in the Ember app to get it to work correctly with Phoenix.


JoshSmith:
We need to add an issue, once this is merged, to update to `ember-simple-auth-token` once the issue with seconds vs milliseconds is fixed by a successful merge of the relevant PR. This will likely be indicated by a version bump to `2.0.0`.

The Ember app fails to refresh the token because the milliseconds version is _always_ higher than the seconds.

JoshSmith:
This also now fixes the type serialization to be properly singularized for the Phoenix API. See https://github.com/AgilionApps/ja_serializer/issues/118

JoshSmith:
If someone wants to help fix tests here while I focus on some of the other fires, would be helpful.

WenInCode:
@joshsmith I'll try take a look tomorrow if no one has by then

begedin:
I've fixed most of the acceptance tests which were all realted to slugged route now having a user and an organization relationship, instead of a polymorphic owner relationship-

I also fixed a test related to logging in due to now using token authentication.

There is a signup test that still needs fixing as well as a couple of integration and unit tests that fail  also probably due to the relationship changes.

JoshSmith:
It looks like acceptance tests were slowed significantly by adding the landing page. I haven't noticed since I've barely touched the Ember app since the conference launch, but we should try to fix this, if not here, then elsewhere.

JoshSmith:
`PATCH /users/me` doesn't exist in Phoenix, so we should probably use this to hit the `/users/:id` endpoint instead.

Any thoughts on whether we should keep `/users/me`? Seems like too much overhead to me.

rileytaylor:
https://github.com/code-corps/code-corps-api/pull/49 on the API notes that 'users/me' is usable by the user, while 'users/:id' is only usable by admins. Is this a separation we necessarily need? Or perhaps can we handle some permissions via the patch request for 'users/:id' and return some sort of 'sorry, you can't edit others stuff unless you are an admin' to cover that need?

JoshSmith:
@rileytaylor that's right, we want to use just `/users/:id`. I think it's better to condense down into that and use permissions on the API.

noslouch:
https://github.com/code-corps/code-corps-phoenix/pull/165 adds posts pagination to the phoenix api. the ember app will need to [update its post adapter to change the key for the desired page number](https://github.com/code-corps/code-corps-ember/blob/develop/app/adapters/post.js#L14) from `number` to `page`.

app/components/project-post-list.js
```diff
@@ -28,10 +28,10 @@ export default Ember.Component.extend({
...
     if (Ember.isPresent(meta)) {
       return {
-        currentPage: parseInt(meta.current_page, 10),
-        pageSize: parseInt(meta.page_size, 10),
-        totalPages: parseInt(meta.total_pages, 10),
-        totalRecords: parseInt(meta.total_records)
+        currentPage: parseInt(meta["current-page"], 10),
+        pageSize: parseInt(meta["page-size"], 10),
+        totalPages: parseInt(meta["total-pages"], 10),
+        totalRecords: parseInt(meta["total-records"])
```
noslouch: I believe these all come back `snake_cased`

JoshSmith: Okay, are you able to pull down this branch at all and see what else I'm missing, there are still some failures that I'm stepping through now.

JoshSmith:
Now tests are failing because query params aren't getting `page[page]` like they should.

app/routes/project/posts/post.js
```diff
@@ -37,8 +37,10 @@ export default Ember.Route.extend({
       comment.save().then(() => {
         route.refresh();
       }).catch((error) => {
-        if (error.errors.length === 1) {
-          this.controllerFor('project.posts.post').set('error', error);
+        let payloadContainsValidationErrors = error.errors.some((error) => error.status === 422 );
```
JoshSmith: Seems like this could be split out into a tiny separate function elsewhere.

JoshSmith:
Need to comment out post editing mention rendering test.

begedin:
Working on resolving conflicts now...

begedin:
Everything seems to be ticking nicely now.

PR #365, opened 2016-08-13
added error-formatter and error-wrapper documentation

rileytaylor:
I added some documentation for the error-formatter and error-wrapper components. 

One thing I am semi-unsure of is in error-wrapper.js line 71. 

What should the @type be? I couldn't find a comparable example in the previously finished documentation. Since it returns a method in the same file, I set that as the type.


app/components/error-wrapper.js
```diff
@@ -27,22 +54,49 @@ export default Ember.Component.extend({
...
+  */
   is404: Ember.computed('httpStatusCodes', function() {
     return this.get('httpStatusCodes').contains(404);
   }),
 
+  /**
+    Updates the background based on the error class.
+
+    @property observeErrorClass
+    @type error-wrapper.updateBackground
```
WenInCode: @rileytaylor I also think there is some uncertainty around what the `@type` of this property should be. *I think* it should be `@type Ember.observer` as it is an observer on a property. Maybe others can chime in here: @JoshSmith @thomascchen ?

app/components/error-wrapper.js
```diff
@@ -1,10 +1,31 @@
...
 export default Ember.Component.extend({
   classNames: ['error-wrapper', 'center-pseudo'],
 
   background: Ember.inject.service(),
 
+  /**
+    Returns a message based on the type of error thrown.
+
+    @property errorClass
+    @type string
```
WenInCode: I think we should use capitalized types `@type string` --> `@type String`.
We might want to add these things to the documentation style guide.

app/components/error-formatter.js
```diff
@@ -1,9 +1,39 @@
 import Ember from 'ember';
 
+/**
+  `error-formatter`returns a formatted error message. Place within an 'if'
```
WenInCode: Can you add a space after `error-formatter`.

WenInCode:
A few small comments, but this still looks great! Hopefully, we can see what others think about the `@type` question.

rileytaylor:
Ah yes, intended to capitalize everything. I'll make those edits and await the final decision on that @type. Ember.observer makes sense.

rileytaylor:
#223

rileytaylor:
Edits are made!

I went ahead and labeled that type as Ember.observer for now since that seems most likely.

app/components/error-wrapper.js
```diff
@@ -68,7 +68,7 @@ export default Ember.Component.extend({
     Updates the background based on the error class.
 
     @property observeErrorClass
-    @type error-wrapper.updateBackground
+    @type Ember.observer
```
JoshSmith: I think this should just be `@type function`, but not positive.

WenInCode: @rileytaylor it does make sense to document it as a method. See: http://yui.github.io/yuidoc/syntax/#method

I would suggest rewriting this as a `@method`

rileytaylor:
Changed to method, should be ready for merge now.

JoshSmith:
Awesome @rileytaylor looks good. Can you squash and then merge?

rileytaylor:
Closing due to squashing/merge issues, will reference in new pull req.

PR #366, opened 2016-08-14
add documentation for progress-bar-container

WenInCode:
This adds documentation for the `progress-bar-container` component, which essentially wraps a progress bar in a div with the `progress-bar-container` class. This is a really simply component, and sometimes simple things are really odd to document. So I just stated the obvious that it is a container for a progress-bar, leaving the implementation details out of it. 

This makes progress on #223


thomascchen:
Looks good to me. Merge at will 💯 💯 💯

PR #368, opened 2016-08-14
add documentation for project-card-skills component

WenInCode:
# What's in this PR?
Adds documentation for the `project-card-skills` component. 

## Questions
- Should we be consolidating the `showMore` & `showLess` actions into one `toggleOverflow` action or is it preferable to have explicit actions for each?

## References
- Progress on issue #223

### DIFF
--- a/app/components/project-card-skills.js
+++ b/app/components/project-card-skills.js
@@ -1,21 +1,76 @@
 import Ember from 'ember';
 
-export default Ember.Component.extend({
+const {
+  Component,
+  set,
+} = Ember;
+
+/**
+  `project-card-skills` composes the given project's list of skills on a
+  project's card.
+
+  ## default usage
+
+  ```handlebars
+  {{project-card-skills skills=project.skills}}
+  ```
+
+  @class project-card-skills
+  @module Component
+  @extends Ember.Component
+ */
+export default Component.extend({
   classNames: ['project-card-skills'],
+
+  /**
+    Returns whether or not the overflowing skills on the project card should be
+    displayed.
+
+    @property overflowHidden
+    @type Boolean
+    @default true
+   */
   overflowHidden: true,
+
+  /**
+    Returns whether or not the toggle for showing overflowing skills should be
+    visible or not.
+
+    @property showToggle
+    @type Boolean
+    @default false
+   */
   showToggle: false,
 
   actions: {
+
+    /**
+      Action that hides the overflowing skills.
+
+      @method showLess
+     */
     showLess() {
-      this.set('overflowHidden', true);
+      set(this, 'overflowHidden', true);
     },
 
+    /**
+      Action that shows the overflowing skills.
+
+      @method showMore
+     */
     showMore() {
-      this.set('overflowHidden', false);
+      set(this, 'overflowHidden', false);
     },
 
+    /**
+      Action that enables the overflow skills toggle. The skills self calculate
+      if they are overflowing. If there is an overflow, this action is called
+      and the toggle is shown; hiding the overflowing skills by default.
+
+      @method skillItemHidden
+     */
     skillItemHidden() {
-      this.set('showToggle', true);
+      set(this, 'showToggle', true);
     },
   },
 });


app/components/project-card-skills.js
```diff
@@ -1,21 +1,73 @@
...
+
+  @class project-card-skills
+  @module Component
+  @extends Ember.Component
+ */
+export default Component.extend({
   classNames: ['project-card-skills'],
+
+  /**
+    Returns whether or not the overflow skills are hidden.
```
JoshSmith: This is a little hard to parse. I would maybe rewrite and explain what's going on here some, that there's an overflow when there are too many skills, how we handle it, etc.

JoshSmith:
I personally like the explicitness of `showMore` and `showLess`.

WenInCode:
I agree, going to keep that the way it is. Going to update this based on feedback right away

app/components/project-card-skills.js
WenInCode: @JoshSmith I think I am a little confused about what is going on here. I don't see the criteria for when there is an overflow. Also I know that we do some computed stuff around if the user has the skills, but are we just showing all the common & not common skills from this union? https://github.com/code-corps/code-corps-ember/blob/develop/app/components/skill-list-items.js#L13 

Thanks :)

rileytaylor: I'm confused as well. It seems that if you call `skillItemHidden()` in the handlebars for `skill-list-items` it will trigger `skillItemHidden()` in this component, which just sets it to true.

So right now there isn't any criteria, it's just set and there isn't a value set to regulate how many skills are visible, so quite frankly I'm unsure how this control even works right now. Seems to me it will just display a bunch of skills regardless of how many there are and offer a toggle that is useless.

[project-card-members ](https://github.com/code-corps/code-corps-ember/blob/develop/app/components/project-card-members.js) handles this a lot better, though its does it through a constant which doesn't give is much control. Ideally, I feel it should be a variable that we can set via handlebars or not set and have it use a default value.

JoshSmith:
@WenInCode just checking on status.

app/components/project-card-skills.js
WenInCode: @JoshSmith still seeking clarification on this

JoshSmith: Ah okay, assigning to self and will address. Or if @begedin wants to provide thoughts.

JoshSmith:
Question in commit above.

JoshSmith:
@begedin assigned to you for your cleanup phase

app/components/project-card-skills.js
begedin: It's been a while since we've worked on this, so I don't really remember what the idea was.

Do we want the same end-result as with project members, as in

* Show a total of X skills on the card
* If there are more than X skills total, show an indicator that some are hidden

If so, as far as I can remember, this is implemented differently due to two reasons

* We are showing only the skills the user also possesses (an overlap of project skills and user skills)
* Skills, unlike project members, are labels of variable length, not icons, so we can't really maintain a simple visual consistency by showing the same number of skills on each card.

We should probably look into making the behavior similar, to project members, for clarity's sake, but I don't think we can make it too similar.

What I think would simplify things a bit, or at least make them clearer in the project/user overlap part is merging of #466. At that point, a project has project-skills and a user has user-skills, so we overlap those two by skill id, then fetch the skills for it, instead of going with the `hasSkill` route.

What to do about variable label length, though, I'm really not sure. If we can somehow constrain the labels to all be the same size, it would make things simpler, but it would also make the list look less appealing. Any ideas here are welcome.

JoshSmith: We're showing both skills they have and skills they don't have. Skills they have come first.

JoshSmith: The overflow criteria actually comes from the skill itself. It's very "data up" in some sense. I would love a different way of doing it, because it's way too funky right now. Would consider reaching out to someone on the Ember slack, someone in our slack with deeper experience, or posting to StackOverflow. As of now, I'm kind of stumped for any better way to do it so I did something hacky and moved on.

JoshSmith:
Just rebase and squash @WenInCode and this is good to merge.

PR #372, opened 2016-08-18
Project-card-members and project-card-skills

rileytaylor:
This finishes project-card documentation 

part of issue #223


app/components/project-card-members.js
```diff
@@ -2,20 +2,59 @@ import Ember from 'ember';
 
 const VISIBLE_MEMBERS_COUNT = 8;
 
+/**
+  `project-card-members' displays the first 8 members of a project and some statistics,
```
thomascchen: I think there is an apostrophe around `project-card-members` that should be a back tick

app/components/project-card-members.js
```diff
@@ -2,20 +2,59 @@ import Ember from 'ember';
...
+  including total number of memebers and how many members are not shown on the card, if any.
```
thomascchen: Could you fix this typo? `memebers` => `members`

thomascchen:
@rileytaylor Couple typos, but other than that looks 👍

PR #378, opened 2016-08-20
Add IntelliJ IDE folder to gitignore.

reviling:


WenInCode:
@reviling seems like a swell idea 😄 . I don't have intelliJ so I don't really know, but looking at other examples do you also need these or no? Thanks!
```
# Intellij
.idea/
*.iml
*.iws
```

reviling:
AFAIK, *.iml or *.iws is not necessary. IntelliJ usually keeps things in the .idea directory.

WenInCode:
@reviling cool - if we need to we can always add more to it later. This looks good - feel free to merge!

PR #388, opened 2016-08-21
Add project-item-category documentation

### LINKED ISSUE #387: Add project-category-item documentation
Add documentation for the project-category-item component.

### References
further work on #223

rileytaylor:
## What's in this PR?

Documentation for the project-item-category component.

### Notes

I believe I'm understanding Ember.computed.notEmpty correctly (line 25) but if someone could double check my description for that property, I would appreciate it. Still working on leveling up my ember and js knowledge.

### References
Closes #387

### DIFF
--- a/app/components/project-category-item.js
+++ b/app/components/project-category-item.js
@@ -1,13 +1,59 @@
 import Ember from 'ember';
 
+/**
+  `project-category-item` represents a category with a project. When selected
+  the user is taken to a list of projects using this category.
+
+  ## Default usage
+  ```Handlebars
+    {{project-category-item category=category}}
+  ```
+
+  @class project-category-item
+  @module Component
+  @extends Ember.Component
+ */
 export default Ember.Component.extend({
   tagName: ['li'],
 
+  /**
+    Returns true if 'userCategory' is not empty
+
+    @property selected
+    @type Boolean
+   */
   selected: Ember.computed.notEmpty('userCategory'),
+
+  /**
+    A service that returns all userCategories.
+
+    @property userCategories
+    @type Ember.Service
+   */
   userCategories: Ember.inject.service(),
+
+  /**
+    The current user.
+
+    @property user
+    @type DS.Model
+   */
   user: Ember.computed.alias('currentUser.user'),
+
+  /**
+    Categories that the user belongs to.
+
+    @property usersUserCategories
+    @type Ember.Array
+   */
   usersUserCategories: Ember.computed.alias('user.userCategories'),
 
+  /**
+    Returns the category if it exists in the user's categories.
+
+    @property userCategory
+    @type Ember.Object
+   */
   userCategory: Ember.computed('category', 'userCategories.userCategories.isFulfilled', function() {
     let category = this.get('category');
     let userCategories = this.get('userCategories');


app/components/project-category-item.js
```diff
@@ -1,13 +1,59 @@
...
+  /**
+    Categories that the user belongs to.
+
+    @property usersUserCategories
+    @type Ember.Array
+   */
   usersUserCategories: Ember.computed.alias('user.userCategories'),
 
+  /**
+    Returns the category.
```
WenInCode: you might want to elaborate on this - I think it only returns the category if the category is in the user's categories.

app/components/project-category-item.js
```diff
@@ -1,13 +1,59 @@
...
+    @property userCategories
+    @type Ember.Service
+   */
   userCategories: Ember.inject.service(),
+
+  /**
+    The current user.
+
+    @property user
+    @type Ember.Object
```
WenInCode: I am not sure what the current policy is for `@type` on properties. Up until now I have been just using `Object` `String` etc. as they link in the docs to the corresponding definitions. I do understand this returns an `Ember.Object` but I am not sure if we are specifying this? Either way works for me, maybe @JoshSmith @npendery or @thomascchen  can chime in here

WenInCode:
This is looking pretty good 👍 - just got one comment and a question for others

app/components/project-category-item.js
rileytaylor: This would be good to clarify. Ember is just wrapping 'Array' and 'Object' so realistically we could just only specify 'Ember' for things like services or Ember specific classes. I'll hold off on submitting an updated commit to preserve the commenting on this until we get some further consensus.

WenInCode: @rileytaylor How about we just continue with what we have been doing so far and go with the non ember specific types (save for something like `Ember.Service`). Kind of just want this to be unblocked

JoshSmith: I'm curious what @begedin thinks on it, actually.

begedin: I don't have an opinion myself, really. Both works.

However, looking at how emberjs itself does it, it does look like they go with the `Ember.` prefix for most stuff. I can't find an instance of `@type Ember.Object`, though. I did find instances of `@extends Ember.Object`.

I also found instances of both `Service` and `Ember.Service` when it's clearly the same type, so who knows.

I would like to, however, note that in this case, if we want to be specific with our type, this is technically a `DS.Model`, in this case, no?

Maybe we should go with `@class DS.Model` instead of `@type Object` to be clearer?

rileytaylor: `DS.Model` would be more descriptive. I think it would be good to note models that way to differentiate them more.

We should probably go with the ember way. If ember is more specific, we should be too. Maybe a vote on specificity?

Whatever we decide I'll go through all the docs we have and make changes to reflect whatever we standardize here.

JoshSmith: I would vote for `DS.Model` here, in favor of specificity. If you're diving into docs and want to know how something works `Ember.Object` isn't particularly revealing. I'd like to be able to jump into `ember-data` instead and see what methods are available on this object.

PR #408, opened 2016-08-28
Make Expertise Buttons Equal-Width and Align Checkboxes

### LINKED ISSUE #399: Expertise choice format is misleading
screen shot 2016-08-26 at 11 32 50 am

Aligning these choices to the center like this makes it seem like the choices are hierarchical. For example, in the screenshot above, it looks like "Data Science" is a child of "Backend Development".

Possible fixes:
- Make the width of the choices the same
- Don't align them to the center

appleJax:
# What's in This PR?
- On start/expertise.hbs, make expertise buttons equal-width, and align the checkboxes within the buttons
# References
- Closes #399


appleJax:
Aw crap, just realized I left a test ID on the button element in role-item.hbs. I'm not home right now, I'll fix it tonight. Feedback on the appearance would be appreciated though, not sure if that's what we were going for.

rileytaylor:
Looks good! Drop those test ID's and squash and you'll be good to merge. Image as reference for others: 
image

appleJax:
Ok, took out the test id, squashed and pushed. I realized I had the styles in the wrong place, so I moved them from expertise.scss to role-item.scss. Also made the button width a percentage instead of pixels.

appleJax:
@JoshSmith I think this PR was waiting on your feedback/approval. See screenshot above.

JoshSmith:
Hmm idk how I feel about this. I think it may need a different design approach altogether to be honest. The simplicity/consistency of buttons was nice and all, but it looks like we're trying to hamfist buttons in.

JoshSmith:
Can I sit on this and circle back? Have some ideas but need to flesh them out when I have a moment.

appleJax:
Ok, no prob.

eriksandall:
I've seen an auto-complete input used for questions like this, which is something I prefer for selects of 7 or more items. But that requires some dev...

rileytaylor:
That is a good point. However it does make things undiscoverable immediately, and we'd want to program in synonyms.

There already is an autocomplete control in the onboarding process so it wouldn't be as difficult to implement. 

But I'm wary of this for the discoverability aspect.

JoshSmith:
Marking as blocked and assigning to me. I've got some ideas on this. Would maybe consider not having buttons but moving to a better designed checkbox-like idea. Largely the same as currently is but without the border/background. Still left checkbox/checkmark and a loading icon for loading state.

JoshSmith:
@appleJax maybe vote to close this for now. Haven't heard much of an issue from folks recently.

PR #409, opened 2016-08-28
Issue 217

appleJax:
First stab at making the onboarding process more clear.


appleJax:
Also, on the right-hand side of start/hello.hbs, there is an image with code-corps members' pictures. Maybe having the member's name underneath their picture would make the page more congruent?

Issue #288, opened 2016-06-27
Add padding to outside of main container


JoshSmith:
Right now if you're at a constrained width, the main container goes right up to the edge of the browser.

PR #411, opened 2016-08-28
Add Padding to Body to Prevent Nav From Butting Against Edge of Screen 

appleJax:
# What's in This PR?
- Add 15px padding to body element to prevent nav from sitting on the edge of the screen on a small-resolution device

# References
- Closes issue #288

### DIFF
--- a/app/styles/main.scss
+++ b/app/styles/main.scss
@@ -11,6 +11,7 @@ html {
 
 body {
   margin: 0;
+  padding: 0 10px;
 
   > .ember-view:first-of-type {
     height: 100%;


app/styles/main.scss
```diff
@@ -11,6 +11,7 @@ html {
 
 body {
   margin: 0;
+  padding: 15px;
```
JoshSmith: Maybe `0 10px`?

appleJax:
Yeah, I realized after the fact I shouldn't have added padding to the top and bottom. 10px is plenty too.

screen shot 2016-08-30 at 10 57 48 pm

JoshSmith:
👍

appleJax:
Good to merge?

PR #415, opened 2016-08-29
Add docs for project-post-list

### LINKED ISSUE #407: Add project-post-list documentation
Adds documentation for the project-post-list component.

Further work on #223

rileytaylor:
# What's in this PR?

Documentation for the project-post-list component.

Check out comments for a few questions I have, I'm unsure about a few types.

## References
Fixes #407 

Progress on: #223

### DIFF
--- a/app/components/project-post-list.js
+++ b/app/components/project-post-list.js
@@ -1,25 +1,87 @@
 import Ember from 'ember';
 
+/**
+  `project-post-list` provides a list of posts within the project and allows 
+  for filtering and pagination.
+
+  ## Default usage
+  ```handlebars
+    {{project-post-list posts=posts project=project}}
+  ```
+  ## Adding post filtering
+    These additional properties allow for filtering.
+  ```handlebars
+    {{project-post-list
+      isFiltered=isFiltered
+      isFilteringClosedPosts=isFilteringClosedPosts
+      isFilteringOpenPosts=isFilteringOpenPosts
+      filterByStatus="filterByStatus"
+      filterByType="filterByType"
+      pageNumber=page
+      posts=model
+      project=project
+      removeTypeFilter="removeTypeFilter"
+      selectedTypes=selectedTypes
+    }}
+  ```
+
+  @class Component
+  @module project-post-list
+  @extends Ember.Component
+ */
+
 export default Ember.Component.extend({
   classNames: ['project-post-list'],
 
+  /**
+    @property store
+    @type Ember.Service
+   */
   store: Ember.inject.service(),
 
+  /**
+    Returns information to handle pagination in the `pager-control`
+
+    @property meta
+    @type Object
+   */
   meta: Ember.computed.alias('posts.meta'),
 
+  /**
+    Returns information for the `pager-control` to handle pagination.
+
+    @property options
+    @type Object
+   */
   options: Ember.computed('meta', function() {
     return this._normalizeMeta(this.get('meta'));
   }),
 
   actions: {
+
+    /**
+      Action that sends the selected type to be added to filtering.
+
+      @method filterByType
+     */
     filterByType(type) {
       this.sendAction('filterByType', type);
     },
 
+    /**
+      Action that sends the selected type to be removed from filtering.
+
+      @method removeTypeFilter
+     */
     removeTypeFilter(type) {
       this.sendAction('removeTypeFilter', type);
     },
 
+    /**
+      Action that sends the selected status for filtering.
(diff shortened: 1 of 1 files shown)


app/components/project-post-list.js
```diff
@@ -1,25 +1,88 @@
...
+    @type Ember.Service
+   */
   store: Ember.inject.service(),
 
+  /**
+    Returns information about the posts in order to sort and display posts in
+    the control.
+
+    @property meta
+    @type
```
rileytaylor: Not sure what to call this. It's pulling off of `posts`, so would it just be `DS.Model`? But at the same time `meta` isn't necessarily a defined property (I'm assuming it has something to do with meta tags in the html, but I also didn't see any Ember docs on this sort of thing.

app/components/project-post-list.js
```diff
@@ -1,25 +1,88 @@
...
+   */
   meta: Ember.computed.alias('posts.meta'),
 
+  /**
+    Returns information about the posts to the `pager-control` to handle page
+    size.
+
+    @property options
+    @type Ember.Computed
```
rileytaylor: it's not Ember.Computed, thought I had deleted that. Also unsure how to define this, though it is returning an Object so perhaps it's just generic `Ember.Object`?

app/components/project-post-list.js
WenInCode: @rileytaylor looks like this here is just a regular `Object`.
code_corps___build_a_better_future_

app/components/project-post-list.js
WenInCode: @rileytaylor `_normalizeMeta` returns an `Object`, if anything. But it is not an `Ember.Object`. The prototype is just `Object`, so that's what I would put the type as.

app/components/project-post-list.js
```diff
@@ -1,25 +1,88 @@
...
+   */
   options: Ember.computed('meta', function() {
     return this._normalizeMeta(this.get('meta'));
   }),
 
   actions: {
+
+    /**
+      Action that sends the the selected type to be added to filtering.
```
WenInCode: Woops! `the the` <-- must have accidentally typed this twice

app/components/project-post-list.js
```diff
@@ -1,25 +1,88 @@
...
+
+      @method filterByType
+     */
     filterByType(type) {
       this.sendAction('filterByType', type);
     },
 
+    /**
+      Action that sends the the selected type to be removed from filtering.
```
WenInCode: 🍝 Same as [above](https://github.com/code-corps/code-corps-ember/pull/415/files#r78126056)

app/components/project-post-list.js
```diff
@@ -1,25 +1,88 @@
...
+
+      @method removeTypeFilter
+     */
     removeTypeFilter(type) {
       this.sendAction('removeTypeFilter', type);
     },
 
+    /**
+      Action that sends the the selected status for filtering.
```
WenInCode: once more

app/components/project-post-list.js
WenInCode: Also, this is information is used for the pagination logic in the `pager-control` component, more so than `page size`. Maybe we can update this and the `meta` doc to reflect that 👌

app/components/project-post-list.js
```diff
@@ -1,25 +1,88 @@
 import Ember from 'ember';
 
+/**
+  `project-post-list` provides a list of posts within the project and allows filtering and paging.
```
WenInCode: I'd opt for `pagination` over `paging` but that is just a 💄

app/components/project-post-list.js
```diff
@@ -1,25 +1,88 @@
...
+
+  ## Default usage
+  ```Handlebars
```
WenInCode: also I think this has to be lower case for it to work (haven't checked, so it might be a good idea to do so) it sets the syntax highlighting for this.

WenInCode:
Thanks for putting this together @rileytaylor - Left some feedback / opinions, hopefully they are helpful 🙌 .

JoshSmith:
@rileytaylor when you have time to get to @WenInCode's comments.

app/components/project-post-list.js
rileytaylor: how does this happen?! I need to go home and rethink my life...

JoshSmith:
@WenInCode would you mind re-reviewing when you have a chance?

Issue #419, opened 2016-08-31
Speed up CI builds


rileytaylor:
This will serve as a master Issue for optimizing our build/testing process.
# Problem

Circle CI builds take a long time, usually 3 minutes, to complete. This is largely due to running `npm install` every time a build is running.
# Solutions

An optimization to take advantage of parallelism would help this, which can be done through`ember-exam`, which replaces the `ember test` sequence with a functionally similar method that uses concurrency.

There isn't any "official" support for this, but there are several workaround to prevent npm from making network requests before checking the cache, with the main npm discussion [here](https://github.com/npm/npm/issues/2568) which is what the npm foundation uses in their CI environment.
## Subtasks
- [x] Implement ember-exam and add it to the circle build config file
- [x] Implement parallel building through ember-exam
- [x] Edit documentation to reflect testing changes
- [x] Implement ember-exam in Circle CI with ember-try
- [x] #371 
- [ ] reduce npm install time by relying on cache unless there is a newer version available (--cache-min-infinity SHOULD do this, but didn't in initial testing)
## References

Ember-exam: https://github.com/trentmwillis/ember-exam

--cache-min Infinity: http://blog.cloudoki.com/npm-features-and-quirks/

JoshSmith:
@rileytaylor what's the status of `ember-exam`? Anything I can do to help unblock that and get a PR in?

rileytaylor:
I haven't had a chance to sit down on it, however I believe I have it added and set up for running in three threads. I need to test it. I have a chunk of time I can work on it right now and I'll update you shortly!

JoshSmith:
@rileytaylor I'm going to close (even though builds are slow) since this is relatively outdated now (we use `yarn`, for example).

rileytaylor:
Totally agreed!

JoshSmith:
@rileytaylor I have opened #1083 and #1084 as follow-up tasks.

PR #420, opened 2016-09-01
Circle CI script changes for faster build times

rileytaylor:
# What's in this PR?

This modifies the `npm install` part of our Circle CI to reduce build times (slightly)

## Changes

`npm install --cache-min Infinity` forces npm to check the cache first and compares it to project.json, then it will download any packages that have been updated. npm doesn't do this be default. It queries all packages listed in project.json with http and then uses the returned information to download or not download. Those http requests take a lot of time, so reducing how many are made makes a HUGE difference.

## Benefits

`npm install` has been an average of 30 seconds. The new command reduces this down to 12 seconds as long as there aren't many packages to update.

## References

Work on #419

### DIFF
--- a/circle.yml
+++ b/circle.yml
@@ -9,7 +9,7 @@ dependencies:
     - bower_components
   override:
     - npm install -g bower
-    - npm install
+    - npm install --cache-min Infinity
     - bower install
     - npm rebuild node-sass
 deployment:


rileytaylor:
`npm install --cache-min Infinity` resulted in a 12-second duration, as opposed to our usual 30-second durations. Not a ton of time saved, but it'll help.

FYI: What this command does is check the cache vs. project.json and only updates if project.json introduces a newer version/something that doesn't exist. Normal `npm install` checks the network first, _then_ looks at what we have, which is obviously very inefficient esp. if you have a slow connection (which circle doesn't thank goodness)

npendery:
Nice this is sweet! Was looking at other ways to achieve this [npm_lazy](https://github.com/mixu/npm_lazy) but seems like that is the best way to do it for a CircleCI build

PR #418, opened 2016-08-31
Make Code Corps responsive. Add bourbon media query breakpoints.

appleJax:
# What's in this PR?

I took a stab at making the site responsive.

Basically, I used bourbon media queries and changed the `@include span-columns()` attribute of certain elements to make them fit better on smaller screens. The first thing I did was create _breakpoints.scss to house the media query breakpoints.

I used the same breakpoints Bootstrap uses, except with max-widths instead of min-widths since Code Corps is already desktop-first. That means if you set styles for a medium-sized screen, the styles will apply to everything below that size unless you override them with another media query for a smaller screen. Large screens (1200px or greater) are the default and don't need a media query. The breakpoints and variable names are as follows:
- $xs-screen: max-width 767px
- $sm-screen: max-width 991px
- $md-screen: max-width 1199px

Here's a bourbon-flavored media query:

``` scss
.subsection {
  @include span-columns(4);
  @include media($xs-screen) {
    @include span-columns(12);
  }
}
```

screen shot 2016-09-15 at 12 21 31 pm

screen shot 2016-09-15 at 12 21 57 pm

screen shot 2016-09-15 at 12 22 09 pm

screen shot 2016-09-15 at 12 22 27 pm

screen shot 2016-09-15 at 12 22 41 pm

This is by no means a final solution. Just wanted to put this out on the table as an option and get some dialogue going.

PR #421, opened 2016-09-05
Add light gray background to footer

appleJax:
# What's in this PR?

Added a light gray background to the site footer. It's a small change, but it really differentiates the footer, especially on mobile.

screen shot 2016-09-04 at 8 43 19 pm

Here it is on mobile (note: this screenshot is from a different branch that has a responsive layout):

screen shot 2016-09-04 at 11 36 31 pm

### DIFF
--- a/app/styles/components/site-footer.scss
+++ b/app/styles/components/site-footer.scss
@@ -1,5 +1,7 @@
 .site-footer {
-  margin: 50px 0 20px 0;
+  background: #f8f8f8;
+  margin: 50px -10px 0;
+  padding: 20px 0 30px;
 
   h4 {
     margin: 7px 0;


npendery:
LGTM! :shipit:

olegko:
digging the color background change.. good to separate it from the rest of the page. The only time it *may* look awkward is on a short page (so footer would take up majority of real estate).

not a big fan of the single column format in footer though, thoughts?

appleJax:
@olegko not 100% sold on the single column footer, but keep in mind that layout is for phones. Similar to heroku.com, codeschool.com ...

ivy-g:
are we planning to add links to the "help"  and "learn" sections?  Otherwise, subtitles seem unnecessary

appleJax:
@ivy-g good point, maybe we should make the subtitles themselves be links to the email and blog instead of having redundant links?

rileytaylor:
The single column is how a lot of people go about it on mobile. But we should explore double column options and even a simplified footer. I like the idea of icons only for social.

ivy-g:
if we took out the subtitles and simplified the social section, it might look cleaner as a single column.

ivy-g:
@appleJax  i agree, making help an email link and learn link to the blog might be a lot nicer

npendery:
Just want to put in that we should create an issue for general footer formatting improvements for both web and mobile views. I think that this is a good discussion and should be retained for further improvements and not removed once this small change gets merged

appleJax:
Yeah, I just realized most of this discussion actually has nothing to do with the gray background, which is the only change this PR makes, lol. All great stuff though, maybe we should move this discussion to PR #418

ivy-g:
@appleJax agreed

appleJax:
So yay or nay on the gray background?

ivy-g:
yay

appleJax:
@npendery @olegko @rileytaylor @WenInCode ?

yay: 2
nay: 0

WenInCode:
I like it

appleJax:
Ok, I think we have a majority vote to add the gray background. Permission to merge?

WenInCode:
:shipit:

PR #427, opened 2016-09-13
Updated package info for S3 hosting

### LINKED ISSUE #375: Determine a new deploy path for Ember
Right now we're deploying to Redis, which is very tied up with our existing Rails API.

We're removing Redis as a single point of failure and would like to figure out the best deployment option.

Some considerations:

- We want the deployment in development to match production as closely as possible
- We would still like to use `ember-cli-deploy` if possible, because their pipeline is the gold standard
- We should look to what other deployments look like where Phoenix serves as the API backend
- We have existing infrastructure on AWS with S3 and CloudFront
- We also have some minimal infrastructure with CloudFlare

npendery:
Closes #375 

Need to add config changes on AWS as well. Explained [here](https://github.com/ember-cli-deploy/ember-cli-deploy-s3-index#configuring-amazon-s3)

### DIFF
--- a/config/deploy.js
+++ b/config/deploy.js
@@ -10,7 +10,7 @@ module.exports = function(deployTarget) {
     redis: {
       allowOverwrite: true,
       keyPrefix: 'code-corps-ember:index'
-    },
+    }
   };
 
   if (VALID_DEPLOY_TARGETS.indexOf(deployTarget) === -1) {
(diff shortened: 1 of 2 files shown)


config/deploy.js
```diff
@@ -53,13 +57,17 @@ module.exports = function(deployTarget) {
...
+    ENV['s3-index']['region'] = process.env.STAGING_S3_REGION;
   }
 
   if (deployTarget === 'production') {
     ENV.build.environment = 'production';
     ENV.redis.url = process.env.PRODUCTION_REDIS_URL;
     ENV.s3.bucket = process.env.PRODUCTION_S3_BUCKET;
     ENV.s3.region = process.env.PRODUCTION_S3_REGION;
+    ENV['s3-index']['bucket'] = process.env.STAGING_S3_BUCKET;
+    ENV['s3-index']['region'] = process.env.STAGING_S3_REGION;
```
JoshSmith: These need changed to `PRODUCTION_`.

PR #444, opened 2016-09-18
Make footer stick to bottom of screen

appleJax:
# What's in this PR?

Makes footer stick to bottom of page.

### DIFF
--- a/app/styles/main.scss
+++ b/app/styles/main.scss
@@ -23,6 +23,10 @@ html, body {
   height: 100%;
 }
 
+.body-container {
+  min-height: calc(100vh - 317px);
+}
+
 p {
   font-size: $body-font-size-normal;
 }
--- a/app/templates/application.hbs
+++ b/app/templates/application.hbs
@@ -1,8 +1,10 @@
 {{#drag-zone}}
   {{navigation-menu}}
-  {{flash-messages}}
-  <div class="main container {{codeTheme.className}}">
-    {{outlet}}
+  <div class="body-container">
+    {{flash-messages}}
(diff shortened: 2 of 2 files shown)


app/styles/main.scss
```diff
@@ -23,6 +23,10 @@ html, body {
   height: 100%;
 }
 
+.body-container {
+  min-height: calc(100vh - 317px);
```
JoshSmith: What happens when the footer's height changes? Like when we make it responsive?

appleJax: We'd add a media query to change the pixel subtraction at the same breakpoint that the footer height changes.

appleJax: We could even put the nav and footer heights into a variable to make the "magic number" more clear.

JoshSmith: Part of what I'm getting at is that if the footer height changes in the future (someone adds another social link, etc) then we end up having to modify this CSS to make it not break.

I think maybe the complexity is higher than the benefit.

Thoughts?

appleJax: Yeah, good point.

JoshSmith: My real fear is just maintainability vs value. A sticky footer is a nice to have, but maintainable code is a must-have.

Anyway, that's why I'm glad to have seen the PR so we can see what the tradeoff looks like. I think the tradeoff is too much. Would you agree?

appleJax: Yeah, I agree. This solution is not maintainable. I'll keep the sticky footer in the back off my mind though, maybe there's a better way to do it.

JoshSmith:
Closing due to discussion above.

PR #447, opened 2016-09-18
Add meta tags

JoshSmith:
# What's in this PR?

Added meta tags for dynamic meta tag creation. We can add more to this later, but was a good way to get started on this for future use.


app/routes/application.js
```diff
@@ -12,6 +13,153 @@ export default Ember.Route.extend(ApplicationRouteMixin, {
...
+        name: 'twitter:title',
+        content: 'Code Corps | Build a better future.'
+      },
+    },
+    {
+      type: 'meta',
+      tagId: 'twitter-url',
+      attrs: {
+        name: 'twitter:url',
+        content: ENV.WEB_BASE_URL
```
WenInCode: @JoshSmith should this be the link to the twitter and not to the `WEB_BASE_URL`

JoshSmith: This doesn't even look like a legit tag.

JoshSmith:
@WenInCode yeah, I'll remove and we'll use that library as well. Will open an issue.

app/routes/application.js
```diff
@@ -12,6 +13,137 @@ export default Ember.Route.extend(ApplicationRouteMixin, {
...
+        property: 'og:site_name',
+        content: 'Code Corps'
+      },
+    },
+    {
+      type: 'meta',
+      tagId: 'meta-og-title',
+      attrs: {
+        property: 'og:title',
+        content: 'Code Corps | Build a better future.'
```
WenInCode: Not sure if this should stay if we are removing the other title, but it's not blocking 💄

JoshSmith: `og:title` is different, so it should stay.

PR #453, opened 2016-09-19
Make footer sticky by adding flexbox container around nav, body, and …

appleJax:
# What's in this PR?

Adds a .flexbox-container div around the nav, body, and footer to make footer stick to bottom of screen.

Also adds a .flexbox-spacer div beween the body and footer to make the body appear as normal (without the spacer, the body is centered between the nav and footer, leaving more space between the nav and body)

Styles for .flexbox-container:
- flex-direction: column
- justify-content: space-between
- min-height: 100vh

Styles for .flexbox-spacer:
- flex-grow: 2

Before:

screen shot 2016-09-18 at 6 45 17 pm

After:

screen shot 2016-09-18 at 7 05 16 pm

Alternate option without spacer (NOT PART OF THIS PR, SHOWN ONLY FOR COMPARISON):

screen shot 2016-09-18 at 6 43 02 pm

### DIFF
--- a/app/styles/components/login-form.scss
+++ b/app/styles/components/login-form.scss
@@ -1,4 +1,5 @@
 .login-form {
   @include span-columns(5);
   @include shift(3.5);
+  padding-top: 10vh;
 }
--- a/app/styles/main.scss
+++ b/app/styles/main.scss
@@ -23,6 +23,17 @@ html, body {
   height: 100%;
 }
 
+.flexbox-container {
+  display: flex;
+  flex-direction: column;
+  justify-content: space-between;
+  min-height: 100vh;
+}
+
+.flexbox-spacer {
+  flex: auto;
+}
+
 p {
   font-size: $body-font-size-normal;
 }
--- a/app/styles/components/signup-form.scss
+++ b/app/styles/components/signup-form.scss
@@ -1,6 +1,7 @@
 .signup-form {
   @include span-columns(5);
   @include shift(3.5);
+  padding-top: 10vh;
 
   input[type="submit"] {
     width: 100%;
--- a/app/templates/application.hbs
+++ b/app/templates/application.hbs
@@ -1,12 +1,15 @@
 {{title "Code Corps"}}
 
 {{#drag-zone}}
-  {{navigation-menu}}
-  {{flash-messages}}
-  <div class="main container {{codeTheme.className}}">
-    {{outlet}}
+  <div class="flexbox-container">
+    {{navigation-menu}}
+    {{flash-messages}}
+    <div class="main container {{codeTheme.className}}">
+      {{outlet}}
+    </div>
+    <div class="flexbox-spacer"></div>
+    {{#unless isOnboarding}}
+      {{site-footer}}
+    {{/unless}}
   </div>
-  {{#unless isOnboarding}}
-    {{site-footer}}
-  {{/unless}}
 {{/drag-zone}}


app/styles/main.scss
```diff
@@ -23,6 +23,17 @@ html, body {
...
 
+.flexbox-container {
+  display: flex;
+  flex-direction: column;
+  justify-content: space-between;
+  min-height: 100vh;
+}
+
+.flexbox-spacer {
+  flex-grow: 2;
```
JoshSmith: What does `flex-grow: 2` mean?

appleJax: Flex grow is a ratio that controls how much space an element takes up on the page. `flex-grow: 2` means that div can expand to take up twice as much space as the other elements on the page.

JoshSmith: Why is that needed in this case? I mean why, specifically, twice as much?

JoshSmith: Will `flex: auto;` not do the same? Serious flexbox ignorance here.

appleJax: Since the `min-height` of the `.flexbox-container` is 100% of the screen height, the div takes up no space unless the content doesn't fill the screen. If the content does not fill the screen, the `.flexbox-spacer` div will expand, pushing the footer to the bottom of the page, and the body towards the nav.

appleJax: The more I compare the different versions, the more I like the alternate version though, the one where the body is centered between the nav and footer.

appleJax:
The alternate version looked weird at first with so much space between the nav and body, but I think the user is going to be focused on the content of the body (in this case, the sign-up form), so it actually makes sense to have it sit in the center of the screen.

app/styles/main.scss
appleJax: `flex` is a shorthand for `flex-grow, flex-shrink, flex-basis`, so I don't think `flex: auto` is valid syntax

appleJax: When I was testing `flex-grow`, any value greater than 1 resulted in the same output on screen, so I stuck with '2' arbitrarily because it was the nearest whole number greater than 1

appleJax: After getting used to the alternate version, though, I honestly think we should get rid of the spacer.

JoshSmith: @appleJax https://css-tricks.com/almanac/properties/f/flex/#article-header-id-3

JoshSmith: I disagree on the alternate version. On other pages it will look particularly odd. This is a kind of edge case.

appleJax: Oh, good tip on `flex: auto`. Let me try it real quick...

appleJax: So yeah, looks like `flex: auto` has the same result as `flex-grow: 2` in this case

JoshSmith: Yeah, I would change to that. It looks clearer to me and does what it says: fills the space completely.

appleJax: Agreed, I didn't know about that trick or I would have used it.

appleJax: I still think you need to give the centered layout more time though. I totally understand how weird it looks at first glance, and it must be way weirder for you since you've seen it the other way for much longer than I have. I really think the body is king though, and the user knows where to find the nav if they need it.

JoshSmith: Awesome! We both learned something.

JoshSmith: I'm mostly referring to other pages (and designs I have in my head right now, which will be made more difficult by such a global style).

appleJax: Can you elaborate a little? What kind of designs are you thinking about? And which specific pages?

app/styles/main.scss
```diff
@@ -23,6 +23,17 @@ html, body {
   height: 100%;
 }
 
+.flexbox-container {
+  display: flex;
+  flex-direction: column;
+  justify-content: space-between;
+  min-height: 100vh;
```
JoshSmith: What does `100vh` mean?

appleJax: 100% of view height. I should look that up actually, lol, not sure how it's different from 100%.

appleJax: I think vh is a percent relative to the screen, and % is based on parent container. So maybe %   is more clear in this case?

JoshSmith: I don't know enough to say for sure.

appleJax:
Added `padding-top: 10vh` to signup-form.scss and `padding-top: 15vh` to login-form.scss to give some breathing room to the forms on those pages. Tried percentages which looked fine full-screen, but the padding did not shrink with a limited view height.

screen shot 2016-09-20 at 9 29 38 am

screen shot 2016-09-20 at 9 29 55 am

app/styles/components/login-form.scss
```diff
@@ -1,4 +1,5 @@
 .login-form {
   @include span-columns(5);
   @include shift(3.5);
+  padding-top: 15vh;
```
JoshSmith: Would prefer `10vh` here to make it consistent.

appleJax: Ok

PR #455, opened 2016-09-20
Added ember-exam

rileytaylor:
# What's in this PR?

This adds `ember-exam` and runs tests through that package, which wraps `ember test` and enhances the process and allowing for parallelism, randomization, etc.

Originally this PR included changes to enable ember exam parallelism in circle ci, but that has been postponed for a later PR for the sake of getting ember exam tooling to everyone for local use.

## References
Progress on: #419

### DIFF
--- a/README.md
+++ b/README.md
@@ -52,8 +52,12 @@ Make use of the many generators for code, try `ember help generate` for more det
 
 ### Running Tests
 
-* `ember test`
-* `ember test --server`
+We use [ember-exam](https://github.com/trentmwillis/ember-exam) for running tests since it allows for parallel testing, randomized orders, and other neat configurations. You can check out the repository readme there to see the available options. A few handy ones to use when running tests locally are outlined below.
+
+* `ember exam` will run the tests
+* `ember exam --split=3 --weighted --parallel` will run tests in 3 PhantomJS instances in parallel with an equal split
+* `ember exam --random` will run the tests in a random order
+* `ember exam --filter='acceptance'` will only run acceptance tests. The same sytax can be used for other types of tests, such as `ember exam --filter='unit'` and `ember exam --filter='integration'`
 
 ### Building
 
--- a/package.json
+++ b/package.json
@@ -10,7 +10,7 @@
   "scripts": {
     "build": "ember build",
     "start": "ember server",
-    "test": "ember test"
+    "test": "ember exam"
   },
   "repository": "",
   "engines": {
@@ -66,6 +66,8 @@
     "ember-concurrency": "0.7.9",
     "ember-data": "^2.8.0",
     "ember-deferred-content": "0.2.0",
+    "ember-disable-proxy-controllers": "^1.0.1",
+    "ember-exam": "0.5.0",
     "ember-export-application-global": "^1.0.5",
     "ember-load-initializers": "^0.5.1",
     "ember-metrics": "0.6.3",
--- a/testem.json
+++ b/testem.json
@@ -2,6 +2,7 @@
   "framework": "qunit",
   "test_page": "tests/index.html?hidepassed",
   "disable_watching": true,
+  "parallel": 3,
   "launch_in_ci": [
     "PhantomJS"
   ],
--- a/.gitignore
+++ b/.gitignore
@@ -27,3 +27,6 @@ testem.log
 .*swp
 .*swo
 .idea
+.vscodeignore
+jsconfig.json
+/typings/*/*


package.json
```diff
@@ -48,18 +48,20 @@
     "ember-cli-pace": "0.1.0",
     "ember-cli-page-object": "1.6.0",
     "ember-cli-password-strength": "0.2.1",
-    "ember-cli-qunit": "^3.0.0",
+    "ember-cli-qunit": "3.0.1",
```
JoshSmith: Why locking to this version number?

testem.json
```diff
@@ -2,6 +2,7 @@
   "framework": "qunit",
   "test_page": "tests/index.html?hidepassed",
   "disable_watching": true,
+  "parallel": 4,
```
JoshSmith: Is this working? Or we need to do something else to activate it?

jsconfig.json
```diff
@@ -0,0 +1 @@
+{"compilerOptions":{"target":"es6","experimentalDecorators":true},"exclude":["node_modules","bower_components","tmp","vendor",".git","dist"]}
```
JoshSmith: What does this do? Does it need documented?

JoshSmith:
Also, don't shoot the messenger, but doesn't look like there was a speed up. You were looking at times for deployment builds. Regular builds were already around 4 mins.

Parallelism is really needed to ensure any kind of real speedup.

testem.json
rileytaylor: This tells testem that it is allowed to use up to 4 containers in parallel. That's just an arbitrary number I set for testing between 2-4 containers.

jsconfig.json
rileytaylor: I think this might have been one of my VS Code addons... let me explore that. I thought I had removed this from the commit.

rileytaylor:
I see the deployment vs. regular. I'll go ahead and mark this as in progress and continue working on getting parallelism functioning.

package.json
rileytaylor: That wasn't intentional, thanks for pointing that out.

jsconfig.json
rileytaylor: This is definitely a Visual Studio Code file. Adding it to the gitignore.

package.json
```diff
@@ -10,7 +10,7 @@
   "scripts": {
     "build": "ember build",
     "start": "ember server",
-    "test": "ember test"
+    "test": "ember exam --split=3 --weighted --parallel"
```
JoshSmith: Do we want to override this in `circle.yml` instead? Maybe change to `ember exam` here and then that config won't cause problems locally for people.

What do you think @rileytaylor?

README.md
```diff
@@ -52,8 +52,16 @@ Make use of the many generators for code, try `ember help generate` for more det
 
 ### Running Tests
 
-* `ember test`
-* `ember test --server`
+We use [ember-exam](https://github.com/trentmwillis/ember-exam) for running tests since it allows for parallel testing, randomized orders, and other neat configurations. You can check out the repository readme there to see the available options. A few handy ones to use when running tests locally are outlined below.
+
+* `ember exam` will run the tests in order
```
JoshSmith: What order, exactly?

README.md
```diff
@@ -52,8 +52,16 @@ Make use of the many generators for code, try `ember help generate` for more det
...
+* `ember exam --split=3 --weighted --parallel` will run tests in 3 PhantomJS instances in parallel with an equal split. This will be a little faster but unstable on systems with low RAM.
```
JoshSmith: How low until you start seeing problems? Do you have a measurement of expected memory usage?

README.md
```diff
@@ -52,8 +52,16 @@ Make use of the many generators for code, try `ember help generate` for more det
...
+* `ember exam --random` will run the tests in a random order
+* `ember exam --filter='acceptance' will only run acceptance tests. Other options here include Unit and Integration. You can also exclude
```
JoshSmith: Looks like you're missing a trailing ` here and that the sentence just sort of trails off. I would also include the actual usage for `='unit'` and `='integration'`.

README.md
```diff
@@ -52,8 +52,16 @@ Make use of the many generators for code, try `ember help generate` for more det
...
+
+If you run into issues with the tests not finishing (PhantomJS is sometimes unstable on Windows) you may want to run the tests in a different browser. To do so, you will need to edit the `testem.json` file and replace 'PhantomJS' with another browser. You can use `testem launchers` to see what browsers are available for use. If testem isn't available as a command, run `npm install -g testem` to install testem on your system.
+
+Make sure you don't commit changes to `testem.json`
```
JoshSmith: Why? What's the motivation here?

package.json
rileytaylor: We could do that, no complaints here.

README.md
rileytaylor: I believe it runs it in compile order, just like `ember test`. I could just modify to "'ember exam' will run the tests" really, no reason to specify.

README.md
rileytaylor: I was actually going to ask if we had any better ideas for swapping or testing using different browsers. I haven't seen a way to specify in the testing command that you want to run the tests in chrome instead.

However, on second thought, this may be unnecessary. We can just deal with people who have errors running tests on a one by one basis. Thoughts on just removing this section @JoshSmith ?

JoshSmith: @rileytaylor yeah that signifies some issues, but I was more wondering about the "Don't commit changes..." part.

package.json
JoshSmith: Okay, you'll need to update that override then.

README.md
rileytaylor: If someone modifies the launcher it might cause some problems for CI if it doesn't have that launcher installed

JoshSmith:
@rileytaylor do we need a separate issue now to address actually running it in parallel in CI?

rileytaylor:
Possibly, It's listed as a task in the super-issue for speeding up CI, but I can create a smaller issue for it as well.

JoshSmith:
Yeah, let's create a separate issue.

JoshSmith:
Just rebase and squash. You'll have a conflict in `package.json`.

PR #466, opened 2016-09-23
Eliminate many to many (join) relationship from all models

### LINKED ISSUE #465: Deleting `node_modules` followed by `npm install` makes some tests fail
# Problem

As described in the title, deleting the node modules folder, followed by a reinstall makes some tests fail. My guess is, caching meant some of the addons didn't update. I would love it if someone would be able to reproduce.

begedin:
# What's in this PR?

The changes here eliminate the many-to-many (join) relationships in the app. I went in that direction mainly due to having issues with failing tests and this seemed like the most straightforward way to fix it, while making the code simpler and easier to follow.

Really, the only place where we outright use the relationships that got eliminated was on the `user-details`, `organization-profile` and `project-card` components, just to plain list those records. They aren't used literally anywhere else. And also, that's just the user-organization join. None of the others were even used.

@JoshSmith  I believe we already discussed wanting to do this, so I hope I went in the right direction.

I would also like to look into eliminating any remaining such relationships from the app (ex, on project) even as part of this PR. 

Then we can move on to removing those properties from our phoenix views. Using coalesce, performance should be good enough and considering how much simpler parts of the ember app become, I think it's worth it.

Open for discussion, though, of course.

## References
Fixes #465


JoshSmith:
@begedin so to be clear, we're talking about the join, which is a separate model in Ember from `organization-membership`?

JoshSmith:
Also, this makes me super happy:



Is there anything we lose by doing this?

begedin:
As far as I could tell, we lose absolutely nothing, though it needs some manual testing. I wanna see how the computed I added to the components behaves and i only got it through automatic tests up to this point. Not gonna have a chance to do it today.

JoshSmith:
@begedin is this ready for review now?

begedin:
@JoshSmith I'm sorry. Yes, yes it is. I was sure I added the label.

JoshSmith:
Needs rebased.

begedin:
@JoshSmith I did some local dev testing and everything seems to be in order. There is also the matter of https://github.com/code-corps/code-corps-api/pull/274, but merging that PR is not strictly necessary for this one to work.

Issue #430, opened 2016-09-14
Replace didInitAttrs with init in STYLEGUIDE.md


begedin:
# Problem

`didInitAttrs` was deprecated in ember 2.6 in favour of plain `init`. The two functions serve the same purpose. 

We do not use the hook in any of our components, but it is mentioned in `STYLEGUIDE.md`

## Subtasks
- [ ] Rewrite `STYLEGUIDE.md` so it talks about `init` instead of `didInitAttrs`

## References

http://emberjs.com/blog/2016/06/08/ember-2-6-and-2-7-beta-released.html

PR #479, opened 2016-09-25
Replace didInitAttrs with init in STYLEGUIDE.md

ivy-g:
Rewrote STYLEGUIDE.md so it talks about init instead of didInitAttrs

fixes:
#430


JoshSmith:
👍 looks good. Merge at will @ivy-g

Issue #336, opened 2016-07-20
Move saving logic from components to routes


WenInCode:
# Problem

While documenting the `post-` components I saw that there was some save logic that was being handled at the component level. This doesn't do a great job of following the Ember best practice of _Data Down, Actions Up_.  I specifically noticed this in the `post-title` & `post-details` components.
# Solution

We should move the `save` logic out of the components and into the route or controller and send actions from the components to the route/controller to save.
## References

[Dockyard: Data Down, Actions Up](https://dockyard.com/blog/2015/10/14/best-practices-data-down-actions-up)

PR #483, opened 2016-09-27
Move Saving Logic for 'post-title' from Component to Route

appleJax:
# What's in this PR?

Installed ember-route-action-helper addon and moved saving logic for 'post-title' from the component to the route. 

Progress on: #336 

I cannot get the integration test to pass. Any advice would be appreciated. The code works when I test it manually.

### DIFF
--- a/package.json
+++ b/package.json
@@ -75,6 +75,7 @@
     "ember-moment": "7.0.0-beta.3",
     "ember-page-title": "3.0.9",
     "ember-resolver": "^2.0.3",
+    "ember-route-action-helper": "^2.0.0",
     "ember-simple-auth": "1.1.0",
     "ember-simple-auth-token": "^1.1.1",
     "ember-sinon": "0.5.1",
--- a/app/components/task-title.js
+++ b/app/components/task-title.js
@@ -93,21 +93,15 @@ export default Component.extend({
 
       @method save
      */
-    save() {
+    applyEdit() {
       let component = this;
       let task = this.get('task');
       let newTitle = this.get('newTitle');
 
       task.set('title', newTitle);
-      task.save().then(() => {
+      this.get('saveTask')(task).then(() => {
         component.set('isEditing', false);
-      }).catch((error) => {
-        let payloadContainsValidationErrors = error.errors.some((error) => error.status === 422 );
-
-        if (!payloadContainsValidationErrors) {
-          this.controllerFor('project.tasks.task').set('error', error);
-        }
       });
-    },
-  },
+    }
+  }
 });
--- a/app/routes/project/tasks/task.js
+++ b/app/routes/project/tasks/task.js
@@ -32,6 +32,16 @@ export default Ember.Route.extend({
   },
 
   actions: {
+    save(task) {
+      return task.save().catch((error) => {
+        let payloadContainsValidationErrors = error.errors.some((error) => error.status === 422 );
+
+        if (!payloadContainsValidationErrors) {
+          this.controllerFor('project.tasks.task').set('error', error);
+        }
+      });
+    },
+
     saveComment(comment) {
       let route = this;
       comment.save().then(() => {
@@ -43,6 +53,6 @@ export default Ember.Route.extend({
           this.controllerFor('project.tasks.task').set('error', error);
         }
       });
-    },
-  },
+    }
+  }
 });
--- a/app/templates/components/task-header.hbs
+++ b/app/templates/components/task-header.hbs
@@ -1,2 +1,2 @@
 <div class="task-icon"></div>
-{{task-title task=task}}
+{{task-title task=task saveTask=saveTask}}
--- a/app/templates/components/task-title.hbs
+++ b/app/templates/components/task-title.hbs
@@ -1,7 +1,7 @@
 {{#if isEditing}}
   <div class="title-actions input-group">
     <button class="clear cancel" {{action 'cancel'}}>Cancel</button>
-    <button class="default save" {{action 'save'}}>Save</button>
+    <button class="default save" {{action 'applyEdit'}}>Save</button>
   </div>
   {{input type="text" value=newTitle name="title"}}
   {{#each task.errors.title as |error|}}
--- a/app/templates/project/tasks/task.hbs
+++ b/app/templates/project/tasks/task.hbs
@@ -1,4 +1,4 @@
-{{task-header task=task saveTitle="saveTaskTitle"}}
+{{task-header task=task saveTask=(route-action "save")}}
 {{task-status task=task}}
 <div class="task-timeline">
   {{task-details task=task}}
--- a/tests/integration/components/task-title-test.js
+++ b/tests/integration/components/task-title-test.js
@@ -4,28 +4,28 @@ import Ember from 'ember';
 
 let mockCurrentUser = Ember.Service.extend({
   user: {
-    id: 1
-  }
+          id: 1
+        }
 });
 
 let mockDifferentUser = Ember.Service.extend({
   user: {
-    id: 2
-  }
+          id: 2
+        }
 });
 
 let mockTask = Ember.Object.create({
   title: 'Original title',
-  body: 'A <strong>body</strong>',
-  number: 12,
-  taskType: 'issue',
-  user: {
-    id: 1,
-  },
-  save() {
-    this.set('title', this.get('title'));
-    return Ember.RSVP.resolve();
-  }
+    body: 'A <strong>body</strong>',
+    number: 12,
+    taskType: 'issue',
+    user: {
+      id: 1,
+    },
+    save() {
+      this.set('title', this.get('title'));
+      return Ember.RSVP.resolve();
+    }
 });
 
 moduleForComponent('task-title', 'Integration | Component | task title', {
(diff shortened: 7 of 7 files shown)


JoshSmith:
@begedin @WenInCode @rileytaylor or anyone else who wants to jump on helping @appleJax out would be greatly appreciated!

app/components/post-title.js
```diff
@@ -93,15 +93,9 @@ export default Ember.Component.extend({
       let newTitle = this.get('newTitle');
 
       post.set('title', newTitle);
-      post.save().then(() => {
+      this.get('doSave')(post).then(() => {
```
WenInCode: this is currently resulting in a few errors that cause some post editing acceptance tests to fail. This is because you aren't actually returning the promise from the save call, so `this.get('doSave')(post)` returns `undefined` when you really want it to return a promise.

app/routes/project/posts/post.js
```diff
@@ -32,6 +32,16 @@ export default Ember.Route.extend({
   },
 
   actions: {
+    save(post) {
+      post.save().catch((error) => {
```
WenInCode: So here you are probably going to want to return the promise.
`return post.save().catch((err....`

appleJax: @WenInCode, good call on adding `return` to the save function. One problem though, I deleted all references to the route's save function from the post-title test, and it still passes. This is because the component's save function sets the post.title field, and the last assert is only checking the value of post.title. The test doesn't check that the post's title has been persisted to the database. Do you have any idea how to pull that off? Maybe at least test that the route action is being called somehow? Not sure how to do that.

appleJax: @WenInCode I actually really don't understand what's going on in the test. As it currently stands, everything passes. If you think it's OK, I trust you.

app/components/post-title.js
JoshSmith: I'm not sure I like this pattern. Give me a second to look through this.

app/templates/project/posts/post.hbs
```diff
@@ -1,4 +1,4 @@
-{{post-header post=post saveTitle="savePostTitle"}}
+{{post-header post=post saveTitle="savePostTitle" doSave=(route-action "save")}}
```
JoshSmith: Why not have `saveTitle` be the action we use here?

JoshSmith: I'm not actually sure I know where that action is being used at all.

appleJax: Ok sure, one sec...

app/routes/project/posts/post.js
```diff
@@ -32,6 +32,16 @@ export default Ember.Route.extend({
   },
 
   actions: {
+    save(post) {
+      return post.save().catch((error) => {
+        let payloadContainsValidationErrors = error.errors.some((error) => error.status === 422 );
+
+        if (!payloadContainsValidationErrors) {
+          this.controllerFor('project.posts.post').set('error', error);
```
JoshSmith: I know this is how we did it originally, but not a fan of what's happening here.

tests/integration/components/post-title-test.js
```diff
@@ -75,7 +75,11 @@ test('it saves', function(assert) {
   this.register('service:current-user', mockCurrentUser);
 
   this.set('post', mockPost);
-  this.render(hbs`{{post-title post=post}}`);
+  this.on('routeSave', (post) => {
+    post.save();
+    return Ember.RSVP.resolve();
+  });
+  this.render(hbs`{{post-title post=post doSave=(action 'routeSave')}}`);
```
JoshSmith: This looks like it needs changed.

package.json
```diff
@@ -75,6 +75,7 @@
     "ember-moment": "7.0.0-beta.3",
     "ember-page-title": "3.0.9",
     "ember-resolver": "^2.0.3",
+    "ember-route-action-helper": "2.0.0",
```
JoshSmith: Would make it `^2.0.0`

app/templates/components/post-header.hbs
```diff
@@ -1,2 +1,2 @@
 <div class="post-icon"></div>
-{{post-title post=post}}
+{{post-title post=post saveTitle=saveTitle}}
```
JoshSmith: There's actually no reason not to call this `save`. We're not just saving one thing. We're only setting one thing and then saving.

appleJax: True. We'll probably end up using the save function for saving comments, no?

JoshSmith: Maybe not for comments (that's a different beast), but for anything that needs saved on the post, yes.

appleJax: The only problem I can see is getting confused about which `save` function is being used. The route's function is called `save`, the component's function is called `save`, and now we're going to call the route's passed down function `save`.

appleJax: Maybe `routeSave` just to be clear?

JoshSmith: I think `savePost`.

appleJax: Ok, done.

JoshSmith: I'm not sure about being confused about them, though. My 2¢.

appleJax: For me it's confusing, maybe I'm just too new to all this. For instance, if I see `{{post-header save=(route-action 'save'}}`, `{{post-title save=save}}`, and in the 'post-title' component, I see `<button {{action 'save'}}`. Then in `post-title.js` I see 
```javascript
save() {
  this.sendAction('save', post);
}
```
I really lose my bearings.

appleJax: If you already know the ins-and-outs of how ember works, it's not a problem, but if you're new and trying to figure out how things work, it's a nightmare.

app/routes/project/posts/post.js
appleJax: I'm not entirely sure what that code is doing. How would you do it differently?

app/templates/components/post-header.hbs
WenInCode: I see where it could be confusing for someone who is not that familiar with embers Data Down Actions up paradigm. I could see why it seems odd when a `named` action is just sending the same `named` action up the line.

I don't think we *really* need to change the names between them. Though, if we were to want to have the names of the actions be different I would be more in favour of the route level actions being consistently called `save`. Something like:

*some component*
```javascript
savePost() {
  this.sendAction('save', post);
}
```

Then again that's just another opinion. Ultimately I am not sure if the naming is too important. The most important thing to convey is that we are not saving just part of the post I guess.

package.json
appleJax: Done

tests/integration/components/post-title-test.js
appleJax: I got rid of the `post.save();` line. What else could be changed?

app/templates/components/post-header.hbs
JoshSmith: @begedin can I get your opinion here, too?

begedin: Generally, the naming that makes sense to me would be something like

```Handlebars
{{post-title save='savePostTitle'}}
```

The reasoning here is
* The post title component has one job, one action, so there's no need for the name to say specifically what the action does. It should be clear from the component itself.

Past that component, though, the name of the action that handles it should be more specific.

I'm even thinking of going a step further. The `post-title` component isn't really saving anymore. It's actually, `applying` or `cancelling` the `edit`. so maybe the action should be `applyEdit`. Maybe, even, the component should be named `editable-post-title` to make it even more self-commenting.

JoshSmith: @begedin I like your thinking here.

@appleJax you okay with those changes?

app/routes/project/posts/post.js
begedin: @JoshSmith I wouldn't touch this here, since it's really out of the PRs scope and I believe we do the same in other places. We should probably create a separate issue.

JoshSmith:
@appleJax this branch is out of date because we renamed "post" to "task" everywhere. The rebase may be a bit tricky.

appleJax:
Sorry, I've been swamped with work, traveling this week. I should be back to normal schedule next week. Those merge conflicts look like fun, lol. I looked at it briefly, I will get to it as soon as I can.

appleJax:
Ok, hopefully it's close to acceptable now! I tried changing `{{task-title}}` to `{{editable-task-title}}`, but I found it was more difficult to find the file because it wasn't grouped together with the other task components anymore. Maybe `{{task-title-editable}}`? Or leave it as is?

JoshSmith:
@appleJax looks good, and I think good to merge once you rebase and squash.

appleJax:
I think @WenInCode has to approve too.

WenInCode:
LGTM

JoshSmith:
🙌  awesome work @appleJax!

appleJax:
Thanks @JoshSmith! Big thanks to @pixelhandler for helping me out on this one. Hopefully I can tackle some other components by myself.

JoshSmith:
Well 🙌  @pixelhandler, too!

PR #484, opened 2016-09-27
Remove our explicit false value for shouldBackgroundReloadRecord

JoshSmith:
# What's in this PR?

Removing `shouldBackgroundReloadRecord` returning `false` because it's caching things we don't want and is not the default way for `ember-data` any longer.


JoshSmith:
I still need to review the behavior in-app.

PR #488, opened 2016-09-27
Fix project card bugs

JoshSmith:
# What's in this PR?

This fixes some bugs being experienced on the front-end right now.

No regression tests because broken in production.

Project card's skills and categories were both broken, and mirage models were out of sync with other models.

Partial fix for #489. Need more thorough evaluation.

### DIFF
--- a/app/components/project-card.js
+++ b/app/components/project-card.js
@@ -21,5 +21,7 @@ const {
 export default Component.extend({
   classNames: ['project-card'],
 
-  projectOrganizationMembers: Ember.computed.mapBy('project.organization.organizationMemberships', 'member')
+  projectCategories: Ember.computed.mapBy('project.projectCategories', 'category'),
+  projectOrganizationMembers: Ember.computed.mapBy('project.organization.organizationMemberships', 'member'),
+  projectSkills: Ember.computed.mapBy('project.projectSkills', 'skill')
 });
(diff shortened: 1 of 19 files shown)


JoshSmith:
Failing test, probably due to Mirage. :(

WenInCode:
projects_ _code_corps

@JoshSmith this is what the project card looks like for me now. Keep in mind this is off a freshly seeded database. seems a bit off doesn't it?

JoshSmith:
@WenInCode your API is missing an endpoint.

app/components/project-card.js
```diff
@@ -21,5 +21,7 @@ const {
 export default Component.extend({
   classNames: ['project-card'],
 
-  projectOrganizationMembers: Ember.computed.mapBy('project.organization.organizationMemberships', 'member')
+  projectCategories: Ember.computed.mapBy('project.projectCategories', 'category'),
+  projectOrganizationMembers: Ember.computed.mapBy('project.organization.organizationMemberships', 'member'),
+  projectSkills: Ember.computed.mapBy('project.projectSkills', 'skill')
```
JoshSmith: As an example, inside of a project's page may show its categories and skills. May use the same thing on the contributors page when approving a new contributor to see the overlap with their categories/skills and the projects'.

WenInCode: @JoshSmith can you create an issue to circle back on this.

PR #496, opened 2016-09-28
Rename posts to tasks

dennybiasiolli:
Fixes #454

PR #497, opened 2016-09-28
Fix tests with phantomjs 1.9.x

dennybiasiolli:
this fix timeout in circleci tests because it install phantomjs version 1.9.x
(working) instead of 2.x that causes tests break on "just cloned" repo


JoshSmith:
Thanks @dennybiasiolli 🎈

dennybiasiolli:
Thanks to you @JoshSmith , now I can rebase #496 ;)

PR #537, opened 2016-09-30
Add ember-cli to GreenKeeper ignore list

### LINKED ISSUE #481: Configure Greenkeeper to exclude some libraries from automatic-upgrades
# Problem

Greenkeeper is making a lot of PRs and that's great, but at the very least, we should skip updating any libraries that are part of Ember-CLI. Those are updated alongside ember-cli, so there should be no reason to do so ahead of time.

https://github.com/greenkeeperio/greenkeeper#repository-options

- [ ] Identify libraries to exclude
- [ ] Exclude libraries
- [ ] Sign off that it works as expected with follow-up issue

QzSG:
# What's in this PR?

Add ember-cli to GreenKeeper ignore list
```
"greenkeeper": {
    "ignore": [
      "ember-cli"      
    ]
  }
```
## References
Fixes #481

### DIFF
--- a/package.json
+++ b/package.json
@@ -84,5 +84,10 @@
     "ember-typed": "0.1.3",
     "loader.js": "^4.0.1",
     "phantomjs": "^1.9.20"
+  },
+  "greenkeeper": {
+    "ignore": [
+      "ember-cli"      
+    ]
   }
 }


JoshSmith:
@begedin can you confirm if this is what you were hoping for?

begedin:
I'm afraid not. What I was hoping for was to ignore those dependencies pulled into our `package.json` / `bower.json` by ember-cli, not ember-cli itself.

However, giving this a second thought, I'm not sure it's achievable, except for maybe just a couple of packages, since the list changes almost with every release.

We can't even do some sort of wildcard based ignore, since we are also using ember-cli addons that aren't part of the base package (such as the moment wrapper).

If greenkeeper supported some sort of whitelist, then that might do the trick, but as is, I'm afraid I've sent us in the wrong direction.

What this PR would do is ignore ember-cli, but still try to update any ember cli dependencies in our `package.json`, which would be almost the opposite of what we want.

begedin:
@QzSG Sorry about making you do this work. I see you also tried to do some research in the greenkeeper repo about wildcard support, but I'm afraid even that wouldn't be enough.

@JoshSmith I would close this.

JoshSmith:
@QzSG yeah now that I think about it, I realize what I was missing. This would skip `ember-cli` and not skip the other things, but, in particular, skipping `ember-cli` itself is the _most_ important thing we want updated.

Will close, and will close the issue, too.

Issue #98, opened 2016-02-22
Write proper model tests


JoshSmith:
None of our models are really unit tested right now.

We should test, at least:

- [x] Category
- [x] Comment
- [x] Organization
- [x] OrganizationMembership
- [x] Preview
- [x] Project
- [x] ProjectCategory
- [x] ProjectSkill
- [x] Skill
- [x] Slugged route
- [x] Task
- [x] User
- [x] UserCategory
- [x] UserRole
- [x] UserSkill

And any computed properties they may have.

begedin:
This might help

https://guides.emberjs.com/v2.3.0/testing/testing-models/

filipecrosk:
@JoshSmith I think we're done here

PR #546, opened 2016-10-02
update category unit model tests

### LINKED ISSUE #524: Write unit tests for the Category model
Write unit tests for the [`Category` Model](https://github.com/code-corps/code-corps-ember/blob/develop/app/models/category.js). There is an [existing unit test file](https://github.com/code-corps/code-corps-ember/blob/develop/tests/unit/models/category-test.js), but it could use more coverage (computed properties & relationship declarations). 

## References
- Progress on: #98 

WenInCode:
# What's in this PR?
Update `category` unit model tests

## References
- Fixes #524 
- Progress on: #98

### DIFF
--- a/tests/unit/models/category-test.js
+++ b/tests/unit/models/category-test.js
@@ -1,4 +1,5 @@
 import { moduleForModel, test } from 'ember-qunit';
+import { testForHasMany } from '../../helpers/relationship';
 
 moduleForModel('category', 'Unit | Model | category', {
   // Specify the other units that are required for this test.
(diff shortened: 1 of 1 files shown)


JoshSmith:
Do you not want to test the attributes of the `Category` model here, too?

WenInCode:
@joshsmith just following the ember model testing guide, my understanding is that there is value in testing functions, computer properties and relationship declarations. I'm not sure how much value there is in testing plain attributes... Unless there is validations on them

begedin:
Same as https://github.com/code-corps/code-corps-ember/pull/547#issuecomment-250958697

It would make sense to add a test that ensures all attributes are accounted for.

WenInCode:
I've updated this to test attribute existence. Reference [this](https://github.com/code-corps/code-corps-ember/pull/547#issuecomment-250982294) for method reasoning.

WenInCode:
Going to update this with relationship helpers from #547 once they get merged.

tests/unit/models/category-test.js
```diff
@@ -10,3 +11,17 @@ test('it exists', function(assert) {
...
+
+test('it should have all its attributes', function(assert) {
+  assert.expect(3);
+
+  const category = this.subject();
+  const attributes = Object.keys(category.toJSON());
+
+  assert.ok(attributes.includes('description'), 'should have description attribute');
+  assert.ok(attributes.includes('name'), 'should have name attribute');
+  assert.ok(attributes.includes('slug'), 'should have slug attribute');
```
JoshSmith: We should refactor these into an `assert.hasAttribute(attributes, 'attribute'`) which itself just runs `assert.ok()` with the description we use there.

Let's make an issue to circle back there. Will DRY these up slightly.

PR #547, opened 2016-10-02
update task model tests

### LINKED ISSUE #521: Write unit tests for the Task model
Update unit tests for the `Task` Model. These have already been started, but could use more coverage. We should be testing relationship declarations & computed properties (see [testing guide](https://guides.emberjs.com/v2.8.0/testing/testing-models/))

## References
- Progress on: #98 

WenInCode:
# What's in this PR?
Update `task` unit model tests

## References
- Fixes #521
- Progress on: #98

### DIFF
--- a/tests/helpers/relationship.js
+++ b/tests/helpers/relationship.js
@@ -1,0 +1,25 @@
+import { test } from 'ember-qunit';
+import Ember from 'ember';
+
+// source: https://gist.github.com/he9qi/b6354a81a0672dc63294
+export function testForHasMany(name, many) {
+  test('should have many ' + many, function(assert) {
+    assert.expect(2);
+    const Model = this.store().modelFor(name);
+    const relationship = Ember.get(Model, 'relationshipsByName').get(many);
+
+    assert.equal(relationship.key, many, 'has relationship with ' + many);
+    assert.equal(relationship.kind, 'hasMany', 'kind of relationship is hasMany');
+  });
+}
+
+export function testForBelongsTo(name, belongsTo) {
+  test('should belong to ' + belongsTo, function(assert) {
+    assert.expect(2);
+    const Model = this.store().modelFor(name);
+    const relationship = Ember.get(Model, 'relationshipsByName').get(belongsTo);
+
+    assert.equal(relationship.key, belongsTo, 'has relationship with ' + belongsTo);
+    assert.equal(relationship.kind, 'belongsTo', 'kind of relationship is belongsTo');
+  });
+}
--- a/tests/unit/models/task-test.js
+++ b/tests/unit/models/task-test.js
@@ -1,4 +1,5 @@
 import { moduleForModel, test } from 'ember-qunit';
+import { testForBelongsTo, testForHasMany } from '../../helpers/relationship';
 
 moduleForModel('task', 'Unit | Model | task', {
   // Specify the other units that are required for this test.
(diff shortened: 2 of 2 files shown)


begedin:
I discussed this with @JoshSmith and we think we should sort of look at ember-side model tests as if they were phoenix-side view tests. 

@filipecrosk did this when submitting his PR, by adding a test to assert all required attributes are accounted for. You can see an example at https://github.com/filipecrosk/code-corps-ember/blob/2197e7926fa6f3b1d30eee63a3705623e2a15f18/tests/unit/models/project-test.js#L17

Adding something like this here would probably bee a good idea.

On one hand, we could look at this as "requiring the contributor to add a new attribute to the model twice", but really, we should look at a test like that as a fail-safe to make sure the contributor hasn't accidentally removed a model attribute. Sort of a "you did this, are you sure?".

WenInCode:
I've updated this to test attribute existence. I tried a few methods and the simplest was the `Object.keys` route that @filipecrosk used in his PR.

tests/unit/models/task-test.js
```diff
@@ -17,6 +22,72 @@ test('it exists', function(assert) {
...
+  let task = this.subject();
+  let attributes = Object.keys(task.toJSON());
+
+  assert.ok(attributes.includes('body'), 'task should have the body attribute');
+  assert.ok(attributes.includes('insertedAt'), 'task should have the insertedAt attribute');
+  assert.ok(attributes.includes('likesCount'), 'task should have the likesCount attribute');
+  assert.ok(attributes.includes('markdown'), 'task should have the markdown attribute');
+  assert.ok(attributes.includes('number'), 'task should have the number attribute');
+  assert.ok(attributes.includes('taskType'), 'task should have the taskType attribute');
+  assert.ok(attributes.includes('status'), 'task should have the status attribute');
```
JoshSmith: Would move up one to stay alpha-ordered.

tests/unit/models/task-test.js
```diff
@@ -17,6 +22,72 @@ test('it exists', function(assert) {
...
+});
+
+test('it should belong to a user', function(assert) {
+  assert.expect(2);
+
+  const task = this.store().modelFor('task');
+  const relationship = get(task, 'relationshipsByName').get('user');
+
+  assert.equal(relationship.key, 'user', 'has relationship with user');
+  assert.equal(relationship.kind, 'belongsTo', 'kind of relationship is belongsTo');
```
JoshSmith: These tests are starting to seem really duplicative. Wondering if we can instead do something like `assert.hasManyRelationship('task', 'user')` or `assert.belongsToRelationship('task', 'project')` which just wraps everything.

tests/unit/models/task-test.js
```diff
@@ -17,6 +22,72 @@ test('it exists', function(assert) {
   assert.ok(!!model);
 });
 
+test('it should have all of it\'s attributes', function(assert) {
```
JoshSmith: The `\'` isn't needed here, since it's "its".

JoshSmith:
@WenInCode improved but could use some minor changes first, then you can decide whether to open a separate issue (we'll need one to circle back anyway) or handle here.

tests/unit/models/task-test.js
JoshSmith: In fact, here we go: https://gist.github.com/he9qi/b6354a81a0672dc63294/

tests/unit/models/task-test.js
WenInCode: english 👍

tests/unit/models/task-test.js
WenInCode: @JoshSmith ya that's a good idea. Want to put that in a separate PR? Then I can circle back and update all of these.

WenInCode:
@JoshSmith I brought in those helpers, and addressed remaining feedback. How does this look now?

tests/helpers/relationship.js
```diff
@@ -0,0 +1,23 @@
+import { test } from 'ember-qunit';
+import Ember from 'ember';
+
+// source: https://gist.github.com/he9qi/b6354a81a0672dc63294
+export function testForHasMany(name, many) {
+  test('should have many ' + many, function(assert) {
+    const Model = this.store().modelFor(name);
```
JoshSmith: I would insert `assert.expect(2);` before everything.

tests/unit/models/task-test.js
```diff
@@ -38,55 +34,11 @@ test('it should have all of it\'s attributes', function(assert) {
...
-  const relationship = get(task, 'relationshipsByName').get('taskUserMentions');
-
-  assert.equal(relationship.key, 'taskUserMentions', 'has relationship with taskUserMentions');
-  assert.equal(relationship.kind, 'hasMany', 'kind of relationship is hasMany');
-});
+testForBelongsTo('task', 'project');
+testForBelongsTo('task', 'user');
+testForHasMany('task', 'comments');
+testForHasMany('task', 'commentUserMentions');
+testForHasMany('task', 'taskUserMentions');
```
JoshSmith: This looks so much better, don't you think? Way easier to write, too.

tests/helpers/relationship.js
WenInCode: sure I'll update this

tests/unit/models/task-test.js
WenInCode: Ya it's a lot better

JoshSmith:
Rebase, squash, and merge! 🎉

PR #555, opened 2016-10-02
Add refresh token endpoint

### LINKED ISSUE #510: Access token could not be refreshed - server responded with .
# Problem

After sitting idle for some time, Ember attempts to get a refresh token from the server. Only problem? No such endpoint exists.

## Subtasks
- [x] Need to figure out what the simple-auth-token library is doing
- [x] Come up with a solution
- [x] If necessary, open up an API issue

JoshSmith:
# What's in this PR?

Closes #510 by adding refresh token endpoint in conjunction with https://github.com/code-corps/code-corps-api/pull/306

- Adds refresh token endpoint
- Updates to use `ember-simple-auth-token` by setting `timeFactor` (works now!)
- Minor updates to README, unrelated

## References
Fixes #510

### DIFF
--- a/config/environment.js
+++ b/config/environment.js
@@ -84,7 +84,6 @@ module.exports = function(environment) {
     // ENV.APP.LOG_VIEW_LOOKUPS = true;
 
     ENV.API_BASE_URL = 'http://api.lvh.me:49235';
-    ENV.SERVER_TOKEN_ENDPOINT = 'http://api.lvh.me:49235/login';
     ENV.WEB_BASE_URL = 'http://localhost:4200';
 
     ENV.sentry.development = true;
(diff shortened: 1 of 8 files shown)


config/environment.js
```diff
@@ -84,7 +84,8 @@ module.exports = function(environment) {
     // ENV.APP.LOG_VIEW_LOOKUPS = true;
 
     ENV.API_BASE_URL = 'http://api.lvh.me:49235';
-    ENV.SERVER_TOKEN_ENDPOINT = 'http://api.lvh.me:49235/login';
+    ENV.SERVER_TOKEN_ENDPOINT = 'http://api.lvh.me:49235/token';
```
barruumrex: You could make your life slightly easier by storing the sub-uri's in a constant and appending that to these env vars. That way when you need to change a route, you don't have to search for every instance of "login" or such.

JoshSmith: Thanks, will implement.

JoshSmith:
Thanks for code review @barruumrex 🙌

PR #570, opened 2016-10-04
Update model tests to use the new hasAttributes helper

### LINKED ISSUE #552: Add preview model unit test to ensure all attributes (body, markdown) are accounted for.
# Problem

I've detailed the "problem" in https://github.com/code-corps/code-corps-ember/pull/547#issuecomment-250958697 but to outline it...

Having a single test, per model, that simply checks all attributes are accounted for, makes sense. We basically look at the model test in ember in the same way as the view test in phoenix.

An example of such a test is https://github.com/filipecrosk/code-corps-ember/blob/2197e7926fa6f3b1d30eee63a3705623e2a15f18/tests/unit/models/project-test.js#L17

On one hand, we could look at this as "requiring the contributor to add a new attribute to the model twice", but really, we should look at a test like that as a fail-safe to make sure the contributor hasn't accidentally removed a model attribute. Sort of a "you did this, are you sure?".

It would make sense to add a test that ensures all attributes are accounted for.

We've already merged #543 without doing this, so this issue is to cover that.

## Subtasks
- [ ] Add test to ensure preview model has `body` and `markdown` attributes

## References
- Progress on: #98 

filipecrosk:
# What's in this PR?

Update model tests to use the new hasAttributes helper. 
`assert.hasAttributes(actualAttributes, expectedAttributes);`

## References
Fixes #552 

Progress on: #98

### DIFF
--- a/tests/unit/models/task-test.js
+++ b/tests/unit/models/task-test.js
@@ -1,5 +1,6 @@
 import { moduleForModel, test } from 'ember-qunit';
 import { testForBelongsTo, testForHasMany } from '../../helpers/relationship';
+import '../../helpers/has-attributes';
 
 moduleForModel('task', 'Unit | Model | task', {
   // Specify the other units that are required for this test.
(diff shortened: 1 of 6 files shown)


tests/unit/models/task-test.js
```diff
@@ -19,19 +20,23 @@ test('it exists', function(assert) {
...
-  assert.ok(attributes.includes('likesCount'), 'task should have the likesCount attribute');
-  assert.ok(attributes.includes('markdown'), 'task should have the markdown attribute');
-  assert.ok(attributes.includes('number'), 'task should have the number attribute');
-  assert.ok(attributes.includes('status'), 'task should have the status attribute');
-  assert.ok(attributes.includes('taskType'), 'task should have the taskType attribute');
-  assert.ok(attributes.includes('title'), 'task should have the title attribute');
+  let actualAttributes = Object.keys(task.toJSON());
+  let expectedAttributes = [
+    "body",
+    "commentUserMentions",
```
filipecrosk: @JoshSmith the only problem is here, this `commentUserMentions` is not a common attribute but a `hasMany`

JoshSmith: @filipecrosk that's fine, we can circle back and fix if needed.

PR #575, opened 2016-10-05
Attempt to add code coverage

### LINKED ISSUE #564: Add ember-cli-code-coverage
# Problem

Add some code coverage reporting. https://github.com/kategengler/ember-cli-code-coverage
## Subtasks
- [ ] Add the reporting
- [ ] Make the reporting work with Circle and Coveralls
- [ ] Add the Coveralls badge to the README

JoshSmith:
# What's in this PR?

Attempting to add code coverage. Doesn't work yet.
## References

Fixes #564

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -13,6 +13,7 @@
 /.sass-cache
 /connect.lock
 /coverage/*
+/coverage*
 /libpeerconnection.log
 npm-debug.log*
 testem.log
--- a/README.md
+++ b/README.md
@@ -2,7 +2,8 @@
 
 ![Code Corps Ember Logo](https://d3pgew4wbk2vb1.cloudfront.net/images/github/code-corps-ember.png)
 
-[![CircleCI](https://circleci.com/gh/code-corps/code-corps-ember.svg?style=shield)](https://circleci.com/gh/code-corps/code-corps-ember) [![devDependency Status](https://david-dm.org/code-corps/code-corps-ember/dev-status.svg)](https://david-dm.org/code-corps/code-corps-ember#info=devDependencies) [![Inline docs](http://inch-ci.org/github/code-corps/code-corps-ember.svg?branch=develop&style=shields)](http://inch-ci.org/github/code-corps/code-corps-ember) [![Slack Status](http://slack.codecorps.org/badge.svg)](http://slack.codecorps.org)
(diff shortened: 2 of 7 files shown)


JoshSmith:
No coverage being generated. Trying with `COVERAGE=true ember test`. Would appreciate someone else pulling down or spotting what I'm doing wrong.

emosher:
I tried this too.  What error are you seeing?  For me it seems to chug along just fine until this happens:
```
ok 898 PhantomJS 1.9 - Unit | Utility | mention parser: it returns the unmodified task string when there are no mentions
not ok 899 PhantomJS 1.9 - Global error: INVALID_STATE_ERR: DOM Exception 11: An attempt was made to use an object that is not, or is no longer, usable. at http://localhost:7357/5484/tests/index.html?hidepassed, line 1071
    ---
        Log: |
            { type: 'error',
              text: 'INVALID_STATE_ERR: DOM Exception 11: An attempt was made to use an object that is not, or is no longer, usable. at http://localhost:7357/5484/tests/index.html?hidepassed, line 1071\n' }
    ...
not ok 900 PhantomJS 1.9 - Unit | Utility | mention parser: global failure
    ---
        actual: >
            null
        stack: >
            http://localhost:7357/5484/tests/index.html?hidepassed:1071
        message: >
            INVALID_STATE_ERR: DOM Exception 11: An attempt was made to use an object that is not, or is no longer, usable.
        Log: |
    ...

1..900
# tests 900
# pass  898
# skip  0
# fail  2
```

It then has another block saying "request aborted" and some additional error spewage.

JoshSmith:
@emosher I don't actually see an error or even _any_ output at all. Nothing happens. No coverage generated whatsoever.

JoshSmith:
We need to run with `COVERAGE=true`.

Also concerned that this is causing issues:

```
// in mirage/config.js

  this.passthrough('/write-coverage');
  this.namespace = 'api';  // It's important that the passthrough for coverage is before the namespace, otherwise it will be prefixed.
```

We don't use `namespace`, but wondering if perhaps we should be.

JoshSmith:
@begedin thoughts on the above comment?

begedin:
@JoshSmith I'm not sure how the namespace would interact with our actual app setup (where `api.` is just part of the URL). We should probably give it a try to see how it works.

JoshSmith:
Actually it looks like `namespace` is not a subdomain but something like `/baz` in `foo.com/baz`.

PR #582, opened 2016-10-07
Remove spaces between checkboxes

rileytaylor:
# What's in this PR?

Changes formatting on the Issues template so that there are no newlines between the checkboxes

### DIFF
--- a/.github/ISSUE_TEMPLATE.md
+++ b/.github/ISSUE_TEMPLATE.md
@@ -3,11 +3,9 @@
 Detail the problem here, including any possible solutions.
 
 ## Subtasks
-- [ ]
-
 - [ ] 
-
-- [ ]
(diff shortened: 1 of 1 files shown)


.github/ISSUE_TEMPLATE.md
```diff
@@ -3,11 +3,10 @@
 Detail the problem here, including any possible solutions.
 
 ## Subtasks
-- [ ]
```
JoshSmith: Can we remove this line, too?

rileytaylor:
line removed. I only had it there since markdown technically likes a newline after headers. But it doesn't have an effect on compiling so it's not a problem.

JoshSmith:
@rileytaylor just rebase, squash, and merge!

PR #597, opened 2016-10-10
Match Mirage config to API

### LINKED ISSUE #495: Go through mirage config and ensure it matches our API completely
# Problem

We need to ensure our mirage server config is actually mirroring our API. A thorough pass through it should do the trick, but it would be a good idea to circle back every now and then to.

* https://github.com/code-corps/code-corps-api is our API repo.
* https://github.com/code-corps/code-corps-api/blob/develop/web/router.ex would be the first step
* We should also look at the controllers and tests to see if there's any special behavior

## Also

* Mirage models should match up app models
* Both mirage and app models should match up API models and views. Views are probably more important here, especially with how relationships are defined

JoshSmith:
# What's in this PR?

- Added missing endpoints
- Added missing coalesces

## References
Fixes #495 

Also identified that the API has two endpoints that are not in Mirage, and may not be necessary to have in the API:

- https://github.com/code-corps/code-corps-api/issues/340
- https://github.com/code-corps/code-corps-api/issues/341

### DIFF
--- a/mirage/config.js
+++ b/mirage/config.js
@@ -74,7 +74,16 @@ export default function() {
   */
 
   // GET /categories
-  this.get('/categories');
+  this.get('/categories', { coalesce: true });
+
+  // POST /categories
+  this.post('/categories');
+
+  // GET /categories/:id
+  this.get('/categories/:id');
+
+  // PATCH /categories
+  this.patch('/categories/:id');
 
   /**
   * Comment user mentions
@@ -103,6 +112,9 @@ export default function() {
   });
 
   // GET /comments/:id
+  this.get('/comments/:id');
+
+  // PATCH /comments/:id
   this.patch('/comments/:id', function(schema) {
     let attrs = this.normalizedRequestAttrs();
     let comment = schema.comments.find(attrs.id);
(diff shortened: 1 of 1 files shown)


mirage/config.js
```diff
@@ -469,21 +535,30 @@ export default function() {
...
   this.delete('/user-roles/:id');
 
   //////////////
   // User skills
   //////////////
 
-  // GET /user-skills
-  this.get('/user-skills');
+  // POST /user-skills
+  this.post('/user-skills', { coalesce: true });
```
JoshSmith: This needs fixed.

mirage/config.js
```diff
@@ -469,21 +535,30 @@ export default function() {
   /////////////
 
   // POST /user-roles
+  this.post('/user-roles', { coalesce: true });
```
JoshSmith: This needs fixed.

JoshSmith: Should be a `GET`.

mirage/config.js
JoshSmith: Should be a `GET`.

JoshSmith:
Fixed, needs review. @begedin?

mirage/config.js
```diff
@@ -178,26 +198,41 @@ export default function() {
...
+
   ///////////
   // Projects
   ///////////
 
   // GET /projects
+  this.get('/projects', { coalesce: true });
+
+  // POST /projects
   this.get('/projects');
```
begedin: Should be a post

Issue #594, opened 2016-10-09
Fix code styling offences throughout the project


sarupbanskota:
# Problem

Through $499 and #593, we introduced `eslint` and `ember-suave`. The changes were made in a manner that the current state of the project stays green. Now it's time to improve styling, so we gotta attack each of the rules that have been turned off, fix them, and remove them off the list.

The full list is outlined below.

## Subtasks
- [ ] comma-dangle
- [ ] spaced-comment
- [ ] object-shorthand
- [ ] quotes
- [ ] brace-style
- [ ] max-statements-per-line
- [ ] keyword-spacing
- [ ] no-var
- [ ] space-in-parens
- [ ] generator-star-spacing
- [ ] indent
- [ ] space-infix-ops
- [ ] operator-linebreak
- [ ] object-curly-spacing
- [ ] dot-notation
- [ ] no-console
- [ ] prefer-template
- [ ] space-before-function-paren
- [ ] comma-spacing
- [ ] no-multiple-empty-lines
- [ ] camelcase
- [ ] no-trailing-spaces
- [ ] key-spacing
- [ ] new-cap
- [ ] one-var
- [ ] space-before-blocks
- [ ] ember-suave/no-const-outside-module-scope
- [ ] ember-suave/no-direct-property-access
- [ ] ember-suave/prefer-destructuring
- [ ] ember-suave/require-access-in-comments
- [ ] ember-suave/require-const-for-ember-properties

## How to fix

1. On the [ESLint file](https://github.com/code-corps/code-corps-ember/blob/develop/.eslintrc.js), remove any rule that you're interested in fixing.
2. This will lead to errors corresponding to offences on that rule.
3. Once you fix them, you may start a new PR 

Let's get this going!

idelairre:
I can work on this during the day (its 1:20am where I'm at). I should be able to take care of them all at once.

sarupbanskota:
@idelairre yay, even better, as I'm going to have a busy week thanks to $DAYJOB. 
Just be sure to keep a separate commit per offence, for easy reviewing.

idelairre:
hey, so in order to get the "no-var" rule to pass I need to change every `var` into a `let` or `const`. `const` is used periodically throughout the project but `let` is preferred even if the variable isn't reassigned. Should I add the "prefer-const" rule so there are clear terms on when to use `const` or just change every `var` into a `let`?

JoshSmith:
I'm in the prefer let camp.

[...]

PR #600, opened 2016-10-10
Fix ESLint offences

idelairre:
# What's in this PR?

Lints files and enforces a bunch of rules.

## References
Progress on: #594


idelairre:
I've also finished the 'no-direct-property-access' but some tests are failing. Hopefully I'll have that up later tonight.

mirage/config.js
```diff
@@ -464,19 +464,19 @@ export default function() {
...
   this.post('/user-roles');
 
   // DELETE /user-roles
   this.delete('/user-roles/:id');
 
-  //////////////
+  // ////////////
   // User skills
-  //////////////
+  // ////////////
```
JoshSmith: These are rather awkward changes. Should we be doing this another way altogether?

tests/integration/components/editor-with-preview-test.js
```diff
@@ -159,7 +159,7 @@ test('it sets the editor min-height to the editor height when previewing and sti
 
   this.$('.preview').click();
 
-  let style = 'min-height: ' + height + ';';
+  let style = `min-height: ${  height  };`;
```
JoshSmith: It looks like many of these replacements were made with lots of spacing left over. Not sure if intentional.

app/components/scroll-top.js
```diff
@@ -2,6 +2,6 @@ import Ember from 'ember';
 
 export default Ember.Component.extend({
   click() {
-    window.scrollTo(0,0);
+    window.scrollTo(0, 0);
```
JoshSmith: This is a weird case.

mirage/config.js
idelairre: This was definitely from using --fix on eslint. I should have given that a closer look. Maybe we should use a different way of making comment blocks stand out, e.g.

```
/**
* User Skills
*/
```
or
```
/*----------------------*
* User Skills
*------------------------/
```
Or we could take out the rule enforce it softly by writing it into a style guide.

tests/integration/components/editor-with-preview-test.js
idelairre: That is a little strange looking, again, probably a result of using --fix. I can hunt for weird spacing in the next commit.

idelairre:
I'm copying the comment I left in the issue thread here.

> The only things left are no-console, camel-case, new-cap, and require-access-in-comments.
> 
> A few notes:
> 
> - I think enforcing camel-case would break the app since the snake-cased properties are coming from the api
> - I wasn't comfortable deleting the console.error entries since one or two of them had comments
> - new-cap was throwing because of jQuery in one test. I'm sure there is a clever way around this or maybe we could just ignore that rule in those instances
> - I have to study the spec for require-access-in-comments to understand how to document correctly, then I can have a go at it.

idelairre:
@JoshSmith Is this good to go or should I add one more commit to fix the funky spacing and comments?

JoshSmith:
I think fix then good!

JoshSmith:
@idelairre not sure why that's causing test failures...

idelairre:
Yeah, that was freaking me out. I couldn't figure out the issue so I redid the commit fixing the spacing on the templates, seems to be ok now.

JoshSmith:
Okay, great. Can you rebase and squash now? I think these are actually too many commits to merge individually but can bundle it all up in one commit message.

JoshSmith:
🙌

PR #606, opened 2016-10-13
Move 'containsCode' function out of task and comment models into mixin

### LINKED ISSUE #131: Dry up containsCode logic

idelairre:
# What's in this PR?

DRYs up comment and task models. Creates "containsCode" mixin.

## References
Fixes #131

### DIFF
--- a/app/mixins/contains-code.js
+++ b/app/mixins/contains-code.js
@@ -1,0 +1,39 @@
+import Ember from 'ember';
+
+const {
+  computed,
+  Mixin
+} = Ember;
+
+/**
+ * `contains-code` provides a computed property to the extended class
+ * that returns whether or not the `body` property contains `code`
+ * tags.
+ *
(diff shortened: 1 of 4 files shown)


app/mixins/contains-code.js
```diff
@@ -0,0 +1,16 @@
+import Ember from 'ember';
+
+const {
+  computed,
+  Mixin
+} = Ember;
+
+export default Mixin.create({
```
JoshSmith: This needs YUIdoc documentation.

idelairre:
I just added documentation to the commit. Should be all set.

JoshSmith:
🙌 more good stuff @idelairre. Thanks again!

PR #612, opened 2016-10-14
Move saving logic for task-details to route

appleJax:
# What's in this PR?

First stab at moving saving logic for {{task-details}} from the component to the route. 

I attempted to write a test for the saving logic, but I cannot get the test to pass. There is currently no test for the saving action, so maybe there's something tricky about it? Any help would be greatly appreciated.

The current error message for the test is `TypeError: 'undefined' is not a function (evaluating 'this.get('mentionFetcher').fetchBodyWithMentions(task, 'task')')`

## References

Progress on: #336

### DIFF
--- a/app/components/task-details.js
+++ b/app/components/task-details.js
@@ -107,19 +107,12 @@ export default Component.extend({
 
       @method saveTaskBody
      */
-    saveTaskBody() {
-      let component = this;
+    applyEditTask() {
       let task = this.get('task');
 
-      task.save().then((task) => {
-        component.set('isEditingBody', false);
+      this.get('saveTask')(task).then((task) => {
+        this.set('isEditingBody', false);
         this._fetchMentions(task);
-      }).catch((error) => {
-        let payloadContainsValidationErrors = error.errors.some((error) => error.status === 422);
-
-        if (!payloadContainsValidationErrors) {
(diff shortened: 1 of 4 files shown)


JoshSmith:
@appleJax I would consider a `rebase` and see if that fixes it. Just did some dependency fixes, although that does not look like it's your issue. Still worth trying.

appleJax:
Ok, the test is passing now.

app/components/task-details.js
```diff
@@ -107,19 +107,13 @@ export default Component.extend({
 
       @method saveTaskBody
      */
-    saveTaskBody() {
+    applyEditTask() {
       let component = this;
```
sbatson5: I know it is out of the scope of this PR, but is this still necessary? The arrow function below allows us to keep the lexical `this`. I think we can delete this and update line 115 to be `this.set('isEditingBody', false)` and reduce some redundancy.

appleJax: Good call @sbatson, I will change that when I get to my laptop.

JoshSmith:
@appleJax can you rebase and squash?

JoshSmith:
Will approve once that's done.

appleJax:
@JoshSmith done. Looks like it's already approved. :shipit:?

JoshSmith:
🙌 good work @appleJax!

Issue #505, opened 2016-09-29
Configure Circle CI to use Ember exam


rileytaylor:
# Problem

Circle can't run ember exam with any of its special properties (e.g. it can't parallelize or randomize)

The best solution to solve this would be to use [ember-try](https://github.com/ember-cli/ember-try), which essentially allows ember-exam to access what it needs and also allows us to . [Ember-exam](https://github.com/trentmwillis/ember-exam) details how to integrate it with ember-try and CI in it's readme.

## References
Continuation of #455 

Progress on: #419

PR #616, opened 2016-10-17
configure ember-try to use ember exam for testing in CI

rileytaylor:
# What's in this PR?

This uses ember-try to allow the use of ember exam in Circle CI. Instead of running the usual `ember test`, circle now runs `ember exam --split=3 --weighted --parallel` to run the tests in parallelization. This saves some time (Our average non-deploy builds have been around 5:50, this runs about 4:10).

Note that ember-try is now included by default within ember-cli, so there isn't a dependency to add to enable this.

I also included several configration options (actually, these are the same ones used in ember-stripe-service) that we can use locally to test against beta, canary, or any other configurations. This is pretty easy to use (you can also set up your own to test adding a certain package, for example) and I'll update the documentation with instructions for this. In short, to run our package.json, use `ember try:one default`, but if you want to test against the current beta build of ember, you'd run `ember try:one beta`.

## References
Solves #505

### DIFF
--- a/config/ember-try.js
+++ b/config/ember-try.js
@@ -1,0 +1,44 @@
+module.exports = {
+  command: "ember exam --split=3 --weighted --parallel",
+  scenarios: [
+    {
+      name: 'default',
+      bower: {
+        dependencies: { }
+      }
+    },
+    {
+      name: 'ember-release',
+      bower: {
+        dependencies: {
+          'ember': 'components/ember#release'
+        },
+        resolutions: {
+          'ember': 'release'
+        }
+      }
+    },
+    {
+      name: 'ember-beta',
+      bower: {
+        dependencies: {
+          'ember': 'components/ember#beta'
+        },
+        resolutions: {
+          'ember': 'beta'
+        }
+      }
+    },
+    {
+      name: 'ember-canary',
+      bower: {
+        dependencies: {
+          'ember': 'components/ember#canary'
+        },
+        resolutions: {
+          'ember': 'canary'
+        }
+      }
+    }
+  ]
+};
--- a/package.json
+++ b/package.json
@@ -10,7 +10,7 @@
   "scripts": {
     "build": "ember build",
     "start": "ember server",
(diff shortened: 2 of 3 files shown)


bower.json
```diff
@@ -16,4 +16,4 @@
   "resolutions": {
     "route-recognizer": "^0.2.7"
   }
-}
+}
```
JoshSmith: I think your config is chopping off trailing line endings. I'm pretty sure this violates our `editorconfig` but not sure.

config/ember-try.js
```diff
@@ -0,0 +1,44 @@
+module.exports = {
+  command: "ember exam --split=3 --weighted --parallel",
+  scenarios: [
+    {
+      name: 'default',
+      bower: {
+        dependencies: { }
```
JoshSmith: Do you need to designate this this way? Is it required to have an empty object?

package.json
```diff
@@ -10,7 +10,7 @@
   "scripts": {
     "build": "ember build",
     "start": "ember server",
-    "test": "ember exam"
+    "test": "ember try:one default"
```
JoshSmith: Does Circle run with `test`?

config/ember-try.js
rileytaylor: Yes unfortunately... but I will double check.

package.json
rileytaylor: Yes, it runs `test` by default unless you override it.

rileytaylor:
I'll add the docs

bower.json
rileytaylor: Yes, it is. VS Code doesn't take the .editorconfig by default, but I've got the extension installed now.

rileytaylor:
@JoshSmith Updated discussion and fixed those issues. I'll squash and merge if things look good.

JoshSmith:
@rileytaylor squash and rebase.

JoshSmith:
🙌

Issue #473, opened 2016-09-25
Add new credit card component


JoshSmith:
# Problem

We want to create a new credit card component to add your card information, as seen on https://projects.invisionapp.com/share/SP8PSQGNM#/screens/191332630
## Subtasks
- [x] Show all credit card icons (can share exports of these) - #668
- [x] Fields for card number, MM/YY, CVC - #668
- [x] Link to terms of use - #668
- [x] Generate stripe token for card when pressing donate - #668
- [x] Pressing donate takes Stripe token and POSTs to server to create subscription - #668
- [x] Pressing donate button shows loading state; changes on error or success - #668
- [x] Create credit card for user when submitted. List created cards if any. - #668

Moved to #669 
- Uses `ember-stripe-service` to generate client-side errors per input field
- Shows server-side errors on input fields, where appropriate
- Use existing code from another project we've done for some of the fields
- Highlight inputs in red when they have errors
- Add error information below each input when errors are present
- _optional_ Credit card highlights as the correct one when entered

sbatson5:
@JoshSmith do you have an export of the credit card icons?

JoshSmith:
@sbatson5 I do!

JoshSmith:
[credit-card-icons.zip](https://github.com/code-corps/code-corps-ember/files/539769/credit-card-icons.zip)

sbatson5:
@JoshSmith Do you have an export for the input icons?

<img width="269" alt="screen shot 2016-10-20 at 2 33 17 pm" src="https://cloud.githubusercontent.com/assets/881981/19575257/9d432e0a-96db-11e6-9162-ead2fed2cfb0.png">

sbatson5:
Also, what are we using for input validations on this repo?

JoshSmith:
@sbatson5 I would like to use `ember-changeset` with `ember-changeset-validations`. Think there's an issue regarding bringing those in. #609

PR #625, opened 2016-10-19
473 credit card component

sbatson5:
# What's in this PR?

This adds a donation component for accepting credit card donations. This PR is for adding the component only and not for wiring it up with the API and stripe.

This PR also adds some new accessibility classes for inputs (such as always wrapping inputs with labels and visually hiding content but leaving for screen readers).

WIP

## References

Progress on: https://github.com/code-corps/code-corps-ember/issues/473

## Screen recording of component

credit-card


app/routes/project.js
```diff
@@ -6,30 +6,30 @@ const {
...
-    if (model) {
-      return {
-        slugged_route_slug: model.get('organization.slug'),
-        project_slug: model.get('slug')
-      };
-    } else {
-      return this._super(...arguments);
-    }
-  }
+  // model(params) {
```
sbatson5: Just commenting out now for testing.

app/components/donation/credit-card.js
```diff
@@ -0,0 +1,7 @@
+import Ember from 'ember';
+
+const { Component } = Ember;
+
+export default Component.extend({
+  classNames: ['credit-card-donation']
```
JoshSmith: Wondering if this could be higher level, like just `credit-card-form` or something?

app/routes/project.js
JoshSmith: What are the pains of testing with this implemented? Want to fix those.

app/routes/project/donate.js
```diff
@@ -0,0 +1,4 @@
+import Ember from 'ember';
+
+export default Ember.Route.extend({
+});
```
JoshSmith: Is this route necessary?

app/styles/main.scss
```diff
@@ -146,3 +146,14 @@ p.subtext {
...
+.visually-hidden {
+  border: 0;
+  clip: rect(0 0 0 0);
+  height: 1px;
+  margin: -1px;
+  overflow: hidden;
+  padding: 0;
+  position: absolute;
+  width: 1px;
+}
```
JoshSmith: Can you explain this one a bit to me?

tests/integration/components/donation/credit-card-test.js
```diff
@@ -0,0 +1,24 @@
...
+  assert.equal(this.$().text().trim(), '');
+
+  // Template block usage:
+  this.render(hbs`
+    {{#donation/credit-card}}
+      template block text
+    {{/donation/credit-card}}
+  `);
+
+  assert.equal(this.$().text().trim(), 'template block text');
```
JoshSmith: Assuming we'll replace these with component tests?

tests/integration/components/donation/donation-container-test.js
```diff
@@ -0,0 +1,24 @@
...
+  assert.equal(this.$().text().trim(), '');
+
+  // Template block usage:
+  this.render(hbs`
+    {{#donation/donation-container}}
+      template block text
+    {{/donation/donation-container}}
+  `);
+
+  assert.equal(this.$().text().trim(), 'template block text');
```
JoshSmith: Same here, TBD?

tests/unit/helpers/format-currency-test.js
```diff
@@ -0,0 +1,25 @@
...
+  let result = formatCurrency([12345]);
+  assert.equal(result, '$12,345.00');
+});
+
+test('formats millions correctly', function(assert) {
+  assert.expect(1);
+
+  let result = formatCurrency([234567890.55]);
+  assert.equal(result, '$234,567,890.55');
+});
```
JoshSmith: Do we have any way to format back again to `POST` up in cents?

JoshSmith:
Also hoping that we have query params for donation amount here (in cents), so that you could ostensibly hit that URL and have it work just the same. That way you could ostensibly share that URL with others.

JoshSmith:
Also guessing that we still need to work in things like `ember-stripe-service` into the mix for the component.

app/router.js
```diff
@@ -41,9 +41,7 @@ AppRouter.map(function() {
     });
   });
 
-  this.route('project', {
-    path: '/:slugged_route_slug/:project_slug'
-  }, function() {
+  this.route('project', function() {
```
JoshSmith: Is this just for testing, too? This breaks a lot of the way the application works otherwise.

tests/integration/components/donation/credit-card-test.js
sbatson5: Yup, I'll add test coverage once I have these components fleshed out.

app/router.js
sbatson5: Yeah, this was just for rendering the page. Didn't want to go through adding projects just to visually look at the component. I can leave a `TODO` here so it is more clear but this will be fixed when this is ready to be merged.

app/styles/main.scss
sbatson5: We want this to be focusable yet visually hidden but we want our content to be read by screen readers.

app/routes/project.js
sbatson5: I meant testing as in testing during dev, not writing tests. This should be fixed without much issue.

app/templates/components/donation/donation-container.hbs
```diff
@@ -0,0 +1,28 @@
...
+</p>
+
+{{donation/credit-card canDonate=canDonate}}
+
+<footer>
+  <p class="additional-content">
+    Your donation will repeat automatically each month.
+    You can cancel or edit your donation at any time.
+    By making this donation, you are agreeing to our
+    {{link-to "Terms of use" "index"}}.
```
sbatson5: Is there a terms of use route yet?

app/components/donation/donation-container.js
```diff
@@ -0,0 +1,14 @@
...
+  Component,
+  computed: { bool }
+} = Ember;
+
+export default Component.extend({
+  classNames: ['donation-container'],
+  donationAmount: 0,
+  projectTitle: null,
+
+  canDonate: bool('projectTitle')
```
JoshSmith: This is a little confusing to me. Can you explain this?

sbatson5: This just returns true if there is a `projectTitle` and false if there isn't.

Technically, we could do `{{#if projectTitle}}` but I think `{{#if canDonate}}` better explains there scenario--the user cannot make a donation, as no project has been selected.  It's a it of additional code, but I think it makes the template easier to read.

Let me know if you'd prefer the logic to be on the `projectTitle` explicitly

sbatson5: I should also note that there shouldn't be a case of having no `projectTitle` but I would like this component to be as stateless as possible in regards to where it is used.

JoshSmith:
Can you explain the reasoning behind the component being under `donation` rather than just a generalized form?

JoshSmith:
It seems like the `credit-card` component is in kind of an awkward position of straddling general and specific. I think making it a `credit-card-form` that takes in some settings (like the submit button text, for example) and not tying it to `donation` would generalize it more.

app/components/donation/donation-container.js
JoshSmith: Do you think relying on `projectTitle` is fragile?

sbatson5: On the component level, I don't think so. This component is meant to be used with a project that has a title.  If the title isn't present (somehow), we display some sort of message indicating that no project is selected.  In that sense, we've covered our one edge case.

sbatson5:
@JoshSmith I'm happy to generalize the credit-card component.  I am of the belief that you build what you need and don't over optimize for the future. Do you foresee a scenario where this credit card component is used outside of the context of a donation?

sbatson5:
Also, while I have your attention, do you have the svg icons for the inputs (https://github.com/code-corps/code-corps-ember/issues/473#issuecomment-255207342)? Once I have those, the styling of this PR should be done.

app/components/donation/donation-container.js
JoshSmith: Maybe we could check for the existence of the project more directly though?

JoshSmith:
@sbatson5 are you only waiting on those icons from me?

JoshSmith:
Was just wondering what else here needs done for this to be fully reviewed and then merged.

sbatson5:
Yeah, I think this is in a good place, so I should be ready for a review once I have those. I'd like to address wiring this up with with stripe models/api requests in a separate PR. I don't want this PR to bloat too much.

JoshSmith:
Here's the sprite: https://cl.ly/3k3f3m3S1C2I

You'll need to figure out the coordinates from there and use them like we do in other instances. You can see the other components of the sprite we've added in the Sass.

app/components/donation/donation-container.js
sbatson5: My assumption is that checking for the existence of a project would happen at the route or controller level and this component would only ingest the title.  The action of donation would bubble up to the parent, so we shouldn't have a need for the entire model here.

JoshSmith:
🙌

Issue #608, opened 2016-10-13
Replace keyboard keycodes with ember-keyboard


JoshSmith:
# Problem

We use `event.which` to decode keyboard input.

Why not make it nice and friendly with [`ember-keyboard`](http://null-null-null.github.io/ember-keyboard)?
## Subtasks
- [ ] Identify places where we should replace
- [ ] Replace those with `ember-keyboard` instead

JoshSmith:
This has been implemented.

PR #627, opened 2016-10-20
Add ember-keyboard and utilize for user-skills-input

appleJax:
# What's in this PR?

Refactored user-skills-input component to use `ember-keyboard`
## References

Progress on: #608


JoshSmith:
@appleJax it looks like this broke the tests for the component.

appleJax:
Yeah, I'm trying to figure that out. It works in Phantom on my machine when I run `ember test --filter="Integration | Component | user skills input"`, but not in Chrome. I'm guessing it has something to do with this (from the `ember-keyboard` usage guide): ```Ember.TextField && Ember.TextArea

To prevent ember-keyboard from responding to key strokes while an input/textarea is focused, we've reopened Ember.TextField and Ember.TextArea and applied the EKOnInsertMixin and EKFirstResponderOnFocusMixin. This ensures that whenever an input is focused, other key responders will not fire. If you want to have responders associated with an input or textarea (such as a rich text editor with keyUp('ctrl+i') bindings), you need to extend these components from Ember.TextField or Ember.TextArea rather than Ember.component.```
Where can I extend Ember.TextField? I tried it from within the component, and it still didn't work. I may be doing it wrong though.

appleJax:
@JoshSmith Ah, ok, I think I got it working. Just have an ESLint error...   I added this code to the top of `user-skills-input.js`:
```javascript
Ember.TextField.extend(EKMixin, {
  activateKeyboard: on('init', function() {
    this.set('keyboardActivated', true);
  })
});
```
Here's the ESLint error: `Avoid accessing Ember.TextField directly (ember-suave/no-direct-property-access)`.  I'm not sure how to access the property indirectly.

appleJax:
Weird, the all tests pass on my local machine, except for the ESLint one. I am having the same problem with PR#632: tests pass on my machine, but fail on Circle. Any suggestions for getting Circle to pass would be appreciated.

app/components/user-skills-input.js
```diff
@@ -1,16 +1,28 @@
...
   isEmpty,
   run: { once },
-  observer
+  observer,
+  on,
+  TextField
 } = Ember;
 
-export default Component.extend({
+Ember.TextField.extend(EKMixin, {
```
sbatson5: @appleJax You destructure `Textfield` above from `Ember` but then explicitly call `Ember.Textfield` here. If you delete `Ember` here, your ESLinting error should pass.

app/components/user-skills-input.js
```diff
@@ -1,16 +1,28 @@
...
+  activateKeyboard: on('init', function() {
+    this.set('keyboardActivated', true);
+  })
+});
+
+export default Component.extend(EKMixin, {
+  activateKeyboard: on('init', function() {
```
sbatson5: Rather than calling `on('init', function() { ... } )` I believe it is now recommended to explicitly overwrite `init` and call `this._super`.  This could be updated to:

```
init() {
  this.set('keyboardActivated', true);
  this._super(...arguments);
},
```

For reference: http://notmessenger.com/proper-use-of-ember-on/

app/components/user-skills-input.js
appleJax: @sbatson5 Oh yeah, for some reason, the tests would fail if I deleted the `Ember.` prefix and used the destructured variable instead. Really bizarre. I added the `Ember.` prefix back to make the tests pass, but I forgot to delete the destructuring part.

app/components/user-skills-input.js
appleJax: Ok, thanks for that, will change...

appleJax:
@sbatson5 do you have time to clone this branch and test it on your machine? These tests are really getting frustrating. I tried using the destructured `TextField.extend` after refactoring to use `init()`. The tests pass in Phantom, fail in Chrome. But when I rerun the individual failing tests in Chrome, they magically pass.

sbatson5:
👋 @appleJax Just looking at this now. I'm not entirely sure I understand what we are trying to do. Why are extending `Textfield` in this component?

From `ember-keyboard`:

> you need to extend
these components from Ember.TextField or Ember.TextArea rather than Ember.component.

I think what they are saying is that you need to extend from `TextField` rather than `Component`.  In your PR, you have

```
TextField.extend(EKMixin, {
  init() {
    this._super(...arguments);
    this.set('keyboardActivated', true);
  }
});
```
But we actually never return or export this `TextField`, so I don't think it is doing what we expect. We may want to create a new component for `TextField` specifically, or perhaps not extend `TextField` at all here.

appleJax:
@sbatson5 ah ok, I was not sure where to extend `TextField`. So you're saying `ember g component textField` and use that within the {{user-skills-input}} component?

sbatson5:
@appleJax Let's move the conversation to #632 since it seems to be a simpler example. If we can get that one passing, it should hopefully make the solution here more obvious. I'm worried we have more than 1 issue here.

appleJax:
Updated PR to use built-in `keyDown` ember function and import `keyDown as keyDownEvent` from `ember-keyboard` to match PR #632. Also added EKOnFocusMixin. Tests are still failing 😞

appleJax:
@null-null-null can you take a look at this PR? According to the ember-keyboard docs, I think I'm supposed to extend this component from Ember.TextField, but I'm not sure exactly what that means or how to do it.

appleJax:
Ok, I created a new component to wrap the `<input>` element in `{{user-skills-input}}` so I could extend the input from `Ember.TextField`. I pass down all the keypress actions as closure-action-attributes and bubble them back up from the input component. All the tests are passing except the one for the escape key. All the escape key does is hide the dropdown, and I confirmed that the dropdown is hidden after pressing the escape key. The test for the escape key, however, only tests the length of the dropdown. I'm not sure why the length is still 1 after being hidden...

appleJax:
Ok, fixed it. I had to manually set `results` to `[ ]`. Not sure why this was not necessary before ember-keyboard.

appleJax:
Vastly simplified the action-passing of keyDown events from what I had before. I still can't think of a way to generalize the textfield component since certain keys need to `preventDefault()` that are specific to `user-skills-input`. I renamed the textfield component `user-skills-textfield` to reflect this coupling.

JoshSmith:
Simplification is looking pretty nice here @appleJax! Looks like a couple small conflicts.

@sbatson5 would you have a chance to review here?

PR #632, opened 2016-10-21
Add ember-keyboard and utilize for submittable-textarea

appleJax:
# What's in this PR?

Added ember-keyboard and refactored `submittable-textarea` component to use it. 
## References

Progress on: #608

### DIFF
--- a/app/components/submittable-textarea.js
+++ b/app/components/submittable-textarea.js
@@ -1,24 +1,16 @@
 import Ember from 'ember';
+import { keyDown as keyDownEvent } from 'ember-keyboard';
 
-const { TextArea } = Ember;
+const { TextArea, on } = Ember;
 
 export default TextArea.extend({
-  keyDown(event) {
-    if (this._isValidCombination(event)) {
-      event.preventDefault();
-      this.sendAction('modifiedSubmit');
-    }
+  init() {
+    this._super(...arguments);
+    this.set('keyboardActivated', true);
   },
 
-  _hasCorrectModifier(event) {
-    return event.ctrlKey || event.metaKey;
-  },
-
-  _isCorrectKeyCode(keyCode) {
-    return keyCode === 13;
-  },
-
-  _isValidCombination(event) {
-    return this._hasCorrectModifier(event) && this._isCorrectKeyCode(event.keyCode);
-  }
+  customSubmit: on(keyDownEvent('cmd+Enter'), keyDownEvent('ctrl+Enter'), function(e) {
+    e.preventDefault();
+    this.sendAction('modifiedSubmit');
+  })
 });
--- a/package.json
+++ b/package.json
@@ -69,6 +69,7 @@
     "ember-disable-proxy-controllers": "^1.0.1",
     "ember-exam": "0.5.0",
     "ember-export-application-global": "^1.0.5",
+    "ember-keyboard": "^2.1.9",
     "ember-load-initializers": "^0.5.1",
     "ember-metrics": "0.7.0",
     "ember-modal-dialog": "0.8.8",
--- a/tests/integration/components/submittable-textarea-test.js
+++ b/tests/integration/components/submittable-textarea-test.js
@@ -1,23 +1,13 @@
 import { moduleForComponent, test } from 'ember-qunit';
 import hbs from 'htmlbars-inline-precompile';
-import Ember from 'ember';
-
-const { $ } = Ember;
+import { triggerKeyDown, initialize } from 'ember-keyboard';
 
 moduleForComponent('submittable-textarea', 'Integration | Component | submittable textarea', {
-  integration: true
-});
+  integration: true,
 
-let pressCtrlEnter = $.Event('keydown', {
-  keyCode: 13,
-  which: 13,
-  ctrlKey: true
-});
-
-let pressCmdEnter = $.Event('keydown', {
-  keyCode: 13,
-  which: 13,
-  metaKey: true
+  beforeEach() {
+    initialize();
+  }
 });
 
 // This test is necessary because a new line after yield in the file will
@@ -36,7 +26,7 @@ test('it sends the modifiedSubmit action with ctrl+enter', function(assert) {
   });
   this.render(hbs`{{submittable-textarea modifiedSubmit="modifiedSubmit"}}`);
 
-  this.$('textarea').trigger(pressCtrlEnter);
+  triggerKeyDown('ctrl+Enter', this.$('textarea'));
   assert.equal(this.$('textarea').val().trim(), '');
 });
 
@@ -48,6 +38,6 @@ test('it sends the modifiedSubmit action with cmd+enter', function(assert) {
   });
   this.render(hbs`{{submittable-textarea modifiedSubmit="modifiedSubmit"}}`);
 
-  this.$('textarea').trigger(pressCmdEnter);
+  triggerKeyDown('cmd+Enter', this.$('textarea'));
   assert.equal(this.$('textarea').val().trim(), '');
 });


appleJax:
For this branch and the branch for PR#627, all tests pass on my machine. Does anyone have any idea what could cause the tests to fail on Circle?

sbatson5:
👋 I think we may be running into issues with `Ember.on` and triggering these key events.  Would you mind updating the component to this:

```
import Ember from 'ember';
import { EKMixin, keyDown as keyDownEvent } from 'ember-keyboard';

const { TextArea } = Ember;

export default TextArea.extend(EKMixin, {
  init() {
    this._super(...arguments);
    this.set('keyboardActivated', true);
  },

  keyDown() {
    if (keyDownEvent('cmd+Enter') || keyDownEvent('ctrl+Enter')) {
      this.sendAction('modifiedSubmit');
    }
  }
});
```

Now we can hook into the `keyDown` event offered by `Ember.Component` and then conditionally check to see if they are hitting `cmd+Enter`. Since `keyDown` is reserved by the component, we have to import `keyDown as keyDownEvent` otherwise, Ember will yell at us.

Let me know if you still get the behavior you want.

appleJax:
@sbatson5 ok, I just left for work, so I will try it as soon as I get home. Thanks for helping me out.

appleJax:
@sbatson5 🙌  you rock! You can see some of my failed attempts commented out in the code.

JoshSmith:
@sbatson5 if you have a second, since you're already familiar here, would you mind reviewing again and giving feedback on why you think this isn't working?

sbatson5:
@JoshSmith @appleJax is this failing? I thought it was passing now

appleJax:
@sbatson5 the tests currently pass, however I tried using `ember-keyboard` for `submittable-textarea-test.js` according to the `ember-keyboard` documentation, which does not work. Also, the current version of `submittable-textarea.js` uses `ember-keyboard` in a way that is not consistent with their documentation. Do have any idea why the implementation in their documentation doesn't work for Code Corps?  [ember-keyboard documentation](http://null-null-null.github.io/ember-keyboard)

sbatson5:
I can play around with this, but I think the documentation for the addon simply needs to be updated as it suggests using `on`, yet in the issues, you can see the owner of the repo suggesting the approach of calling the methods directly with `this._super` like we are:

https://github.com/null-null-null/ember-keyboard/issues/52#issuecomment-256208065

Looking into this further, I think this continues to be a product of using `on`.  If you look at the test helper that is imported, you can see that it calls triggers the keydown event on the document:

https://github.com/null-null-null/ember-keyboard/blob/1def7ee62d369f6533daf89f5d27bc4c8dc00578/addon/utils/trigger-event.js#L18

If you updated your test to be `$(document).trigger(pressCtrlEnter)` versus what you have (`this.$('textarea').trigger(pressCtrlEnter)`), you would see the same failure.

So, what is happening here? `ember-keyboard` observes the _entire_ document and not just the component that is using the mixin. To test, I mocked up a[ quick twiddle here](https://ember-twiddle.com/2c8d60135728df60da20898d06a7e2aa?openFiles=components.text-area.js%2Ctemplates.components.text-area.hbs).

Put the cursor in a text area and hit enter. You will see 2 things in the console--the first is the normal keyDown and the second is from `ember-keyboard`.  Then click out of a textArea and hit enter. You will see that the observed event from `ember-keyboard` still fires. This is because it observes the _whole_ documents.

Sadly, I think what you have is a lot better.  This seems like an odd oversight from `ember-keyboard`.

appleJax:
@sbatson5 wow, thanks so much for looking into this! I discovered that I can use EKOnFocusMixin to prevent the `on` event from firing when the text field is not in focus. Other than that, it looks like your solution is the best one. I've kind of lost faith a little bit in `ember-keyboard` though. @JoshSmith do you still feel `ember-keyboard` is a good solution?

JoshSmith:
@appleJax depends. Are there changes we can fold into `ember-keyboard` to improve the situation?

appleJax:
@JoshSmith do you mean can we contribute to the `ember-keyboard` project to fix it? I will take a look and see if there's anything I can do.

JoshSmith:
@appleJax absolutely! If it looks like there's something that's not working there, we certainly can help improve it.

appleJax:
@JoshSmith yeah, that's a great idea! I will see what I can do...

sbatson5:
@appleJax let me know if you want help opening an issue/PR against `ember-keyboard`.  Happy to help!

appleJax:
@sbatson5 yes, thank you, that would be great! It's gonna take me awhile to dig through the code and figure out what to do, so if you already have an idea, go for it!

JoshSmith:
@appleJax let us know if you need any help on the `ember-keyboard` side of things.

null-null-null:
Hey! I'm the maintainer for `ember-keyboard` and I just came across this issue because it was linked to another issue on the repo. Skimming over the thread, it looks like there's some uncertainty about why `ember-keyboard` observes the whole document, rather than individual components. This is to support cases where you'd want a component to respond to a keyboard event, even when it's not focused. For instance, imagine you have a search bar that you want to select when the user presses the `s` key. The focus will likely be elsewhere in the document when they press that `s` key, so `ember-keyboard` watches at the top to catch those events.

Looking over you code, you should be able to get it to work like so:

```js
import Ember from 'ember';
import { keyDown as keyDownEvent } from 'ember-keyboard';

const { TextArea } = Ember;

export default TextArea.extend({
  init() {
    this._super(...arguments);
    this.set('keyboardActivated', true);
  },

  customSubmit: on(keyDownEvent('cmd+Enter'), keyDownEvent('ctrl+Enter'), function(e) {
    e.preventDefault();
    this.sendAction('modifiedSubmit');
  })
});
```

Note that we can remove the EKMixin and EKOnFocusMixin here, since all TextFields and TextAreas automatically have that behavior. There an explanation for why at the bottom of the [usage](http://null-null-null.github.io/ember-keyboard/#/usage) page.

Anyway, I hope that helps! Please feel free to ask any further questions. Code Corps seems like a cool project, and I hope you get this working, regardless of whether or not you use `ember-keyboard`.

JoshSmith:
Hey @null-null-null! This is a seriously good example of doing open source right. Really awesome of you to jump in here and help.

Kevin, let me know if this clears things up for you.

appleJax:
@null-null-null thanks for chiming in! @JoshSmith sorry I've been MIA, my schedule has been pretty crazy lately. I should have some time tonight to take a look at it.

appleJax:
@null-null-null, your suggestion worked for `submittable-textarea.js`, but I tried adding ember-keyboard to trigger the keyDown event for `submittable-textarea-test.js`, which does not work. It seems pretty straightforward according to the ember-keyboard docs. I tried focusing the input before calling `triggerKeyDown`. Do you have any insight into what I'm doing wrong?

null-null-null:
@appleJax I believe the inconsistency you're experiencing is because when integration tests are run in isolation, the `ember-keyboard` initializer isn't run. However, if it's run as part of the whole test suite, the initializers get run during the acceptance tests and their effect carries on to the integration tests.

I just pushed up a new version of `ember-keyboard` that exposes the initializer so you can run it manually in your integration tests. For instance:

Since integration tests do not run initializers, you'll need to manually run the `ember-keyboard` initializer before each test:

```js
import { initialize } from 'ember-keyboard';

moduleForComponent('my component', 'Integration | Component | my component', {
  integration: true,

  beforeEach() {
    initialize();
  }
});
```

appleJax:
@null-null-null awesome! 🙌  Thanks for looking into that. I didn't see this in time to test it before I go to work, so I'll try it out tonight. 

Is it possible to modify `triggerKeyDown`, `triggerKeyUp`, and `triggerKeyPress` to take a second parameter (that defaults to `document`) so you can pass in the input or textarea where you want to trigger the key?

null-null-null:
@appleJax No problem. Check out the latest release (`v2.1.7`) and the updates to [the docs](http://null-null-null.github.io/ember-keyboard/#/testing).

appleJax:
@null-null-null 🙌 :tada: Fanfrickintastic! Thanks so much for that! I'm excited to try it out tonight!

appleJax:
@null-null-null the failing tests all say `_emberKeyboard.initialize is not a function`

null-null-null:
@appleJax That was a problem on my end. It's fixed, and I've verified that it now works with your codebase. Be sure to update to `v2.1.9`, and thanks for your quick report.

appleJax:
@null-null-null woot woot!!!! Thank you for all your help! :party_parrot:

appleJax:
@JoshSmith @sbatson5 this PR is good to review now!

appleJax:
@JoshSmith :shipit: ?

PR #633, opened 2016-10-23
Add yuidoc blocks to pager-control.js, see #223

pixelhandler:
# What's in this PR?

- Added yuidoc blocks in app/components/pager-control.js file

Progress on: #223

### DIFF
--- a/app/components/pager-control.js
+++ b/app/components/pager-control.js
@@ -6,18 +6,110 @@ const {
   computed: { alias, equal, gt }
 } = Ember;
 
+/**
+  The `PagerControlComponent` provides pagination controls, e.g. previous/next
+
+  ## default usage
+
+  Pass in the options that describe the paging needs, like size, total, current.
+
+  ```js
+  options: {
+    pageSize: 5,
+    totalRecords: 7,
+    currentPage: 1,
+    totalPages: 2
+  }
+  ```
+
+  ```handlebars
+  {{pager-control options=options}}
+  ```
+
+  @class pager-control
+  @module Component
+  @extends Ember.Component
+*/
 export default Component.extend({
   classNames: ['pager-control'],
+
+  /**
+    Number of pages to show
+
+    @property pagesToShow
+    @type Number
+  */
   pagesToShow: 5,
 
+  /**
+    Flag to check if there are at least one or more pages to show.
+
+    @property canShowPages
+    @type Boolean
+  */
   canShowPages: gt('totalPages', 1),
+
+  /**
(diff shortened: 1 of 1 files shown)


JoshSmith:
@rileytaylor @pixelhandler it's a good question as to what extent we're overdocumenting in other cases and if this is the right level of documentation here. Are some of these functions self-documenting enough just by naming convention? I'm not totally positive, but am certainly open to some competing ideas here to see what direction we should take.

rileytaylor:
@joshsmith @pixelhandler agreed. Some of the simple boolean values could go without description, but some could definitely use `@default` tags. Also the beginning header describing the entire component.

pixelhandler:
@JoshSmith

> Are some of these functions self-documenting enough just by naming convention? 

If you plan to use generated docs, only methods with doc blocks will show up. For example here are the generated docs I use for an addon project, http://ember-jsonapi-resources.com/docs/

pixelhandler:
@rileytaylor I'll add the defaults and the component doc block as well, thanks

JoshSmith:
@pixelhandler I wasn't clear: I meant that you didn't need much more than `@property canShowPages`, for example, given the self-documenting nature.

pixelhandler:
@JoshSmith cool thanks for the clarity, I do like seeing `@type` w/ `@property`

JoshSmith:
I was unclear again! I just meant that you didn't need extra English-language description about what it means.

pixelhandler:
with generated docs the English language description can be handy even when the property name is super targeted, there is usually some clarity that can be added. Also the generated docs too read nice with human descriptions vs variable names only, here is another example page of props, http://ember-jsonapi-resources.com/docs/classes/ApplicationAdapter.html

JoshSmith:
Okay, then I would definitely consider adding those descriptions here, then.

pixelhandler:
@rileytaylor @JoshSmith I updated this PR after rebasing on develop, Added doc block for the component as well as human friendly descriptions for the properties.

JoshSmith:
👍 please do rebase and squash when you can @pixelhandler!

pixelhandler:
@JoshSmith I already rebased (I figured maybe you'd use github's squash and merge feature).

I just squashed; so CI should run again :)

pixelhandler:
@rileytaylor do you need to 👍 for the requested changes ?

rileytaylor:
yes, looks like I do! Looks good now.

JoshSmith:
🙌

Issue #469, opened 2016-09-25
Add workflow to start accepting donations


JoshSmith:
# Problem

We need to add the rest of the UI to start accepting donations. Starts here: https://projects.invisionapp.com/share/RT8PS9FK2#/screens/191325346
## Subtasks
- [x] When the account does not yet have any fundraising goals, show add goal UI (custom amount, description)
- [x] Allow users to edit their goals
- [x] Allow users to add a new fundraising goal
- [ ] When one goal has been added, allow account to turn on donations (tells API to create Stripe plan for that account) - #670
## References

Progress on: https://github.com/code-corps/code-corps-api/issues/10

JoshSmith:
@begedin still not sure the best way to tell the API to actually turn donations on. Should that be what creates the Stripe Plan for that account?

begedin:
Good question. 

That would likely be the most straightforward to implement.

Other choice is a literal field on the project table that we set to true. We would still have to create a plan in that case, though. Maybe it doesn't have to be a restful action? 

Maybe we simply post to a `/projects/:id/enable_donations` endpoint and then that endpoints does what it needs to do?

JoshSmith:
I think creating `StripePlan` is best, so let's go with that. It makes the most sense, IMO.

JoshSmith:
Closing in favor of #670.

PR #634, opened 2016-10-24
Add donation goals

### LINKED ISSUE #470: Add a donation goal component
# Problem

As seen on https://projects.invisionapp.com/share/RT8PS9FK2#/screens/191325346 and https://projects.invisionapp.com/share/RT8PS9FK2#/screens/191325348
## Subtasks
- [ ] Have a field for adding an amount
- [ ] Have a field for adding a description
- [ ] Create goal button when it's new
- [ ] Edit link once saved
- [ ] When editing, can cancel or save goal
## References

Progress on: #469

begedin:
# What's in this PR?
## `donation-goal``

A component for viewing a donation goal. Can send an `edit` action to the parent, indicating that the "Edit" link rendered by the component was clicked.

``` handlebars
{{donation-goal 
  donationGoal=donationGoal 
  edit=(action externalEditHandler donationGoal)}}
```
## `donation-goal-edit`

A component for editing a donation goal

The component is not added anywhere into the app, but it's well tested with integration tests. I tried to finally use the "proper" data-down-actions-up approach here (or I hope I did, someone with more ember knowledge can correct me on it.

The component has two actions, `save` and `cancel`. In order to use it correctly, we do the following:

``` handlebars
{{donation-goal-edit
  cancel=(action externalCancelHandler)
  donationGoal=donationGoal
  save=(action externalSaveHandler donationGoal)}}
```
## `donation-goals`

A component for displaying a list of donation goals. Each can be edited/cancelled/saved individually

``` handlebars
{{donation-goals
  add=(action 'addDonationGoal')
  cancel=(action 'cancelDonationGoal')
  edit=(action 'editDonationGoal')
  project=project
  save=(action 'saveDonationGoal')}}
```

The action handlers will both receive the donation goal which is being saved/cancelled.
## Route behavior

Initially, all of the project's donation goals are listed. If there are none, a new record is initialized and added to the list. 
- Saving a new or existing records commits those changes to the API.
- Cancelling a new record removes that record from the list.
- Cancelling an existing record simply puts it back to display mode.
## TODO
- [x] `donation-goal-edit` component
- [x] `donation-goal` component
- [x] `donation-goals` component
- [x] `donation-goal` route
- [x] UI details (texts, etc.)
- [x] Hide the `edit` link and `add new buttons` for unauthorized users
- [x] add missing route and model unit tests (basic, but did not create them yet).
- [x] rewrite `donation-goal-edit` to bind to the donationGoal, not individual properties. Should then set initial values of individual properties on init. More explicit that way, but easier to use.
* [x] Merge #654 
* [x] Add it to the `donation-goal-edit` component here.
* [x] Update tests
* [x] Add project ability tests

## References

Progress on  #470
Progress on: #469

### DIFF
--- a/app/components/donation-goal-edit.js
+++ b/app/components/donation-goal-edit.js
@@ -1,0 +1,52 @@
+import Ember from 'ember';
+
+const {
+  Component
+} = Ember;
+
+/**
+ * `donation-goal-edit` used to edit new and existing donation goals
+ *
+ * ## default usage
+ *
+ * ```handlebars
+ * {{donation-goal-edit
+ *   canCancel=externalCancellableFlag
+ *   cancel=(action externalCancelHandler donationGoal)
+ *   donationGoal = donationGoal
+ *   save=(action externalSaveHandler donationGoal)}}
+ *
+ * Used as above, the `externalSaveHandler` function will receive a call with the actual
+ * `donationGoal` model as the first argument, and the properties set via the
+ * component as the second argument.
+ *
+ * Similarly, the `externalCancelHandler` function will recieve a call with the
+ * `donationGoal` as the first and only argument.
+ *
+ * @class donation-goal-edit
+ * @module  Component
+ * @extends Ember.Component
+ */
+export default Component.extend({
+  classNames: ['donation-goal-edit'],
+
+  /**
+   * Indicates if "cancel" button should render.
+   *
+   * Cancel button should only render if one of two cases
+   * - the record is already persisted and we are simply editing it
+   * - the record is new, but there are other persisted records, so cancelling this one
+   *   does not mean there will be no persisted records at all
+   *
+   * @property canCancel
+   * @type {Boolean}
+   */
+  canCancel: false,
+
+  init() {
+    this._super(...arguments);
+    let donationGoal = this.get('donationGoal');
+    let { amount, description } = donationGoal.getProperties('amount', 'description');
+    this.setProperties({ amount, description });
+  }
+});
--- a/app/controllers/project/settings/donations.js
+++ b/app/controllers/project/settings/donations.js
@@ -1,0 +1,67 @@
+import Ember from 'ember';
+
+const { Controller } = Ember;
+
+export default Controller.extend({
+  actions: {
+    /**
+     * Action which calls to initialize a new donation goal record
+     * for the current project.
+     *
+     * Triggered when user clicks a button to add a new donation goal
+     *
+     * @method addDonationGoal
+     * @param  {DS.Model} project A project record to initialize a new donation goal for.
+     */
+    addDonationGoal(project) {
+      project.get('donationGoals').createRecord().set('isEditing', true);
+    },
+
+    /**
+     * Action which switches a donation goal from edit to view mode
+     * If the donation goal is a new, unsaved record, it will be destroyed.
+     *
+     * Triggered when user clicks the cancel button,
+     * when editing or adding a donation goal.
+     *
+     * @method cancelDonationGoal
+     * @param  {DS.Model} donationGoal A donation goal record
+     */
+    cancelDonationGoal(donationGoal) {
+      donationGoal.set('isEditing', false);
+
+      if (donationGoal.get('isNew')) {
+        donationGoal.destroyRecord();
+      }
+    },
+
+    /**
+     * Action which switches a donation goal from view to edit mode.
+     *
+     * Trigged when user clicks the edit link.
+     *
+     * @method editDonationGoal
+     * @param  {DS.Model} donationGoal A donation goal record
+     */
+    editDonationGoal(donationGoal) {
+      donationGoal.set('isEditing', true);
+    },
+
+    /**
+     * Action which commits changes to a donation goal.
+     *
+     * Triggers when user clicks the save button while edditing or
+     * adding a new donation goal.
+     *
+     * @method saveDonationGoal
+     * @param  {DS.Model} donationGoal An unmodified or new donation goal record
+     * @param  {Object}   properties   A hash consisting of donation goal properties to update and save
+     */
+    saveDonationGoal(donationGoal, properties) {
+      donationGoal.setProperties(properties);
+      donationGoal.save().then((donationGoal) => {
+        donationGoal.set('isEditing', false);
+      });
+    }
+  }
+});
--- a/mirage/config.js
+++ b/mirage/config.js
@@ -63,7 +63,7 @@ function generatePreviewMentions(schema, preview) {
 
 // The set of routes we have defined; needs updated when adding new routes
 const routes = [
-  'categories', 'comment-user-mentions', 'comments', 'organizations',
+  'categories', 'comment-user-mentions', 'comments', 'donation-goals', 'organizations',
   'task-user-mentions', 'tasks', 'previews', 'projects', 'project-categories',
   'slugged-routes', 'user-categories', 'users'
 ];
@@ -134,6 +134,15 @@ export default function() {
 
     return comment;
   });
+
+  /**
+   * Donation goals
+   */
+
+  this.get('/donation-goals', { coalesce: true });
+  this.get('/donation-goals/:id');
+  this.patch('/donation-goals/:id');
+  this.post('/donation-goals');
 
   /**
   * Organization memberships
--- a/mirage/factories/donation-goal.js
+++ b/mirage/factories/donation-goal.js
@@ -1,0 +1,6 @@
+import { Factory, faker } from 'ember-cli-mirage';
+
+export default Factory.extend({
+  amount: faker.list.random(10000, 20000, 15020),
+  description: faker.lorem.paragraph
+});
--- a/tests/helpers/attributes.js
+++ b/tests/helpers/attributes.js
@@ -1,0 +1,17 @@
+import { test } from 'ember-qunit';
+import Ember from 'ember';
+import 'code-corps-ember/tests/helpers/has-attributes';
+
+const { get } = Ember;
+
+// modified from gist source for relationships
+// source: https://gist.github.com/he9qi/b6354a81a0672dc63294
+export function testForAttributes(name, expectedAttributes) {
+  test(`should have attributes [${expectedAttributes.join(', ')}]`, function(assert) {
+    assert.expect(1);
+    let Model = this.store().modelFor(name);
+    let actualAttributes = get(Model, 'attributes');
+
+    assert.hasAttributes(actualAttributes, expectedAttributes);
+  });
+}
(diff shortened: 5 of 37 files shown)


app/templates/components/edit-donation-goal.hbs
```diff
@@ -0,0 +1,16 @@
+<div class="amount-group">
+  <span class="currency">$</span>
+  {{input type="number" name="amount" value=amount placeholder="Custom amount"}}
+  <span class="period">per month</span>
+</div>
+<div class="input-group">
+  {{textarea name="description" value=description placeholder="Tell your donors what this goal will allow you to do. Be specific, but make it interesting!"}}
```
sbatson5: Thoughts on just breaking this onto multiple lines just for readability?

```
{{textarea
  name="description"
  value=description
  placeholder="Tell your donors what this goal will allow you to do. Be specific, but make it interesting!"}}
```

app/components/edit-donation-goal.js
```diff
@@ -0,0 +1,44 @@
...
+
+  ## default usage
+
+  ```handlebars
+  {{edit-donation-goal
+    isNew=donationGoal.isNew
+    amount=donationGoal.amount
+    description=donationGoal.description
+    save=(action externalSaveHandler donationGoal)
+    cancel=(action externalCancelHandler)}}
```
JoshSmith: Should we alphabetize here?

tests/integration/components/edit-donation-goal-test.js
```diff
@@ -0,0 +1,65 @@
...
+  this.set('cancelHandler', (donationGoal) => {
+    let expectedGoalProperties = this.get('donationGoal').getProperties('amount', 'description');
+    let actualGoalProperties = donationGoal.getProperties('amount', 'description');
+    assert.deepEqual(expectedGoalProperties, actualGoalProperties, 'donation goal is curried unchanged');
+  });
+
+  this.render(hbs`{{edit-donation-goal isNew=false amount=1 description='Test' cancel=(action cancelHandler donationGoal)}}`);
+
+  this.$('.cancel').click();
+});
```
JoshSmith: Can we use page objects here?

app/styles/components/edit-donation-goal.scss
```diff
@@ -0,0 +1,26 @@
...
+  .amount-group {
+
+    background: $background-gray;
+    border: 1px solid;
+    border-color: $gray;
+    border-radius: 4px;
+    float: left;
+    margin: 0 0 10px 0;
+
+    .currency, input[type=number], .period {
```
JoshSmith: Yeah, these will need changed.

app/components/edit-donation-goal.js
```diff
@@ -0,0 +1,44 @@
...
+
+  actions: {
+    save() {
+      let values = this.getProperties('amount', 'description');
+      this.sendAction('save', values);
+    },
+
+    cancel() {
+      this.sendAction('cancel');
+    }
```
JoshSmith: Looks like these are not alpha ordered. Also should add YUIdoc documentation.

JoshSmith: Documentation is for the internals, I mean.

app/templates/components/edit-donation-goal.hbs
JoshSmith: Agree here.

tests/integration/components/edit-donation-goal-test.js
begedin: It's an integration test. Are we using them in other integration tests?

JoshSmith:
Still in progress, right?

JoshSmith:
I've moved everything to `project/settings/donations` and rewrote some of the logic for when you can add/edit/cancel.

Also added information to the top of the page.

Removed the `currency` helper and replaced with `format-currency`; `currency` was fairly broken. Changed some logic in `format-currency`.

Nothing I've changed has tests for it. Not completely finished with the work here.

begedin:
@JoshSmith I was unable to completely finalize this, but I got close.

*I fixed the tests that got broken
*I added a project ability, with the option to call `can('manage donation goals in project)`, which returns true if user is the owner of the project's organization. The current credentials of "at least admin" for `can('manage organization')` were not enough, since we explicitly said API-side that only owners can do this.
* I updated the tests to match that a well. I also added an authentication helper which enables quickly creating an organization member of a specific role in an organization and authenticating with that member.
* I further updated `format-currency` so it works with an integer amount in cents, since that was my understanding we were going to use, API-side. If that's wrong, then I really think we should either rename `DonationGoal.amount` or `StripePlan.amount`, since having them be named the same and one be in dollars while the other is in cents is confusing.
* [ ] We still need tests for the project ability.

app/components/donation-goal-edit.js
```diff
@@ -0,0 +1,85 @@
...
+    /**
+     * Action that simply sends a `cancel` action outwards, to the parent
+     * container.
+     *
+     * Fired when user clicks the "Cancel" button
+     *
+     * @method cancel
+     */
+    cancel() {
+      this.sendAction('cancel');
```
sbatson5: We seem to be using closure actions with these components:
https://github.com/code-corps/code-corps-ember/pull/634/files#diff-e54aed91f3708411eb6f637d00ab5400R15

Thoughts on `this.get('cancel')()`?

JoshSmith:
@sbatson5 could use your opinion on `format-currency`. I like what you had previously and think it's okay to be dealing in either dollars or cents when just inputting data into a form field, but that we should do our conversion to cents just when we `POST` to the server. What do you think?

sbatson5:
@JoshSmith I don't know if I have a strong opinion on when the conversion should happen. When I wrote the helper, it made sense because it was coming from user input. Do we want to make that conversion when the request goes out or do we want the API to handle it?

What if we created a transform on the Ember side that took care of the serializing and deserializing for us?  That we can honor the user input in these helpers and that would handle the conversion to and from the API.

I can PR something real quick to see how we like it.

app/components/donation-goal-edit.js
begedin: I'm not sure I understand, so any advice is welcome. What's the advantage?

begedin:
@JoshSmith If we merge #654, then we will still need the `format-currency` helper, but we will have to revert it back to what you wrote. Just to make that clear. I do think we probably need both either way, though, since we're dealing with user input in the `donation-goal-edit` component, so the flow here should be:

* [ ] Merge #654 
* [ ] Add it to the `donation-goal-edit` component here.
* [ ] Update tests

begedin:
I added the missing unit tests. While doing this, I took a look at our `has-attributes` helper for model tests. I took a look at how `testBelongsTo` and `testHasMany` work and I repurposed `has-attributes` into a `testAttributes` helper, which works in a similar way.

I believe that, in a separate issue, we should get rid of `has-attributes` and move it's behavior into `testAttributes`, since it's less hacky that way (it would not override a `QUnit` prototype.

Additionally, we can then merge `helpers/relationships` and the now added `helpers/attributes` into `helpers/model.js`, for IMO slightly better code organization.

@JoshSmith Let me know if I should create an issue for that.

You can take a look at how the new stuff works in `tests/unit/models/donation-goal`

app/components/donation-goal-edit.js
begedin: I had a superficial understanding of it, so I made an effort to comprehend it a bit better. This video was the most helpful for me - https://www.emberscreencasts.com/posts/76-closure-actions-in-ember-20-components-passing-functions

I made use of closures and currying to the point where I now pass in the functions from controller level (closures are sadly not possible with route actions), which cleared a lot of "propagate this action up" code and made templates more explicit and easier to understand, in my opinion.

sbatson5: For route actions, we actually use `ember-route-action-helper` which allows us to pass actions from our route down to components https://github.com/code-corps/code-corps-ember/blob/develop/package.json#L78

You can do `cancel=(route-action "cancel")`.  It has been discussed that route-actions as closure actions should make it into an Ember release at some point.

sbatson5: Also, regarding that screencast, it is slightly outdated. It is not encouraged to call the actions from the `attrs` directly.  Locks has a good article about why you shouldn't call attrs directly: https://locks.svbtle.com/to-attrs-or-not-to-attrs

Basically, attrs is a Glimmer Components thing, which hasn't landed in a stable version of Ember quite yet.  In Classic Components (curly brace components), attrs have a slightly different meaning behind the scenes.

begedin:
This should finally be ready for final review and (hopefully) merge.

mirage/config.js
```diff
@@ -136,6 +136,15 @@ export default function() {
   });
 
   /**
+   * Donation goals
+   */
+
+  this.get('/donation-goals', { coalesce: true });
+  this.get('/donation-goals/:id');
+  this.patch('/donation-goals/:id');
+  this.post('/donation-goals');
```
begedin: I really don't think we need to comment on each and every mirage config shorthand. A comment like

```Javascript
// POST /categories
this.post('/categories');
```

really just tells us what the code is already telling us.

mirage/factories/donation-goal.js
```diff
@@ -0,0 +1,6 @@
+import { Factory, faker } from 'ember-cli-mirage';
+
+export default Factory.extend({
+  amount: faker.list.random(10000, 20000, 15020),
```
begedin: Mirage does not support custom attribute transforms, as far as I can tell, so we have to deal with cents. It makes sense, really. Mirage is there to mock the API, and the API deals with cents.

tests/helpers/attributes.js
```diff
@@ -0,0 +1,17 @@
...
+// source: https://gist.github.com/he9qi/b6354a81a0672dc63294
+export function testForAttributes(name, expectedAttributes) {
+  test(`should have attributes [${expectedAttributes.join(', ')}]`, function(assert) {
+    assert.expect(1);
+    let Model = this.store().modelFor(name);
+    let actualAttributes = get(Model, 'attributes');
+
+    assert.hasAttributes(actualAttributes, expectedAttributes);
+  });
+}
```
begedin: We can move this and `tests/helpers/relationhips` into a common `test/helpers/model` and then get rid of `has-attributes`, since that part can be mimicked from within this helper, without the need to monkeypatch the `QUnit` prototype.

app/controllers/project/settings/donations.js
```diff
@@ -0,0 +1,67 @@
+import Ember from 'ember';
+
+const { Controller } = Ember;
```
sbatson5: You should be able to use the `ember-route-action-helper` if you don't want to go through the controller.  Although, I don't see anything wrong with using a controller in this case.

PR #637, opened 2016-10-25
add documentation for the flash-messages component

### LINKED ISSUE #636: Add documentation for flash-messages component
Add documentation for the [flash-messages](https://github.com/code-corps/code-corps-ember/blob/develop/app/components/flash-messages.js) component

## References
- Progress on: #223

WenInCode:
# What's in this PR?
Adds documentation for the `flash-messages` component

## References
- Fixes #637 
- Progress on: #223

### DIFF
--- a/app/components/flash-messages.js
+++ b/app/components/flash-messages.js
@@ -5,8 +5,27 @@ const {
   inject: { service }
 } = Ember;
 
+/**
+  `flash-messages` is a container for displaying flash messages. It leverages
+  the flash-messages service from the
+  [ember-cli-flash](https://github.com/poteto/ember-cli-flash) add-on.
+
+  ## default usage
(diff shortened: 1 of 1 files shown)


app/components/flash-messages.js
```diff
@@ -5,8 +5,27 @@ const {
   inject: { service }
 } = Ember;
 
+/**
+  `flash-messages` is a container for the apps flash messages. It displays the
```
JoshSmith: Minor note, but just an apostrophe missing in "app's".

I might simplify though to:

> `flash-messages` is a container for displaying flash messages.

WenInCode: @JoshSmith I've updated this now - thanks for the feedback

PR #647, opened 2016-10-26
Simplify comments list behavior so API requests are consistent with other types of records.

### LINKED ISSUE #644: Fix comments endpoint (part of API changes)
# Problem

More details in https://github.com/code-corps/code-corps-api/pull/372, but basically, we want to simplify the comments endpoint, so it's more inline with the rest of our API

* Eliminate the filtering by task id from comments endpoint
* Add filtering by coalesced ids
* Simplify ember so the task route works similarly to other routes that handle relationships (as described above).

## References
Needs merge of API PR https://github.com/code-corps/code-corps-api/pull/372

begedin:
# What's in this PR?

This makes it so that, on the `project.tasks.task` route, instead of explicitly fetching comments for the task, we instead simply bind to `task.comments` instead.

This change causes the following
* Comments no longer need to be fetched via `projects/:id/tasks/:number/comments/` route. Instead, we simply fetch from `/comments/` using coalesced id requests.
* We no longer need the comments adapter at all
* The `model` and `setupController` hooks for the `project.tasks.task` route are much simpler.

## Further options
We could potentially further simplify things. 

The `setupController` hook could check if the session is authenticated and simply push a new comment into the `task.comments` collection. 

The `comment-list-component` could then be the one that displays either a `comment-item` component, or some sort of `comment-item-edit` component, which can be used both for new and existing comments, depending on an `isEditing` flag.

I believe it would reduce and dry up the code related to this section of the app a good deal, but it can be done in a separate PR. @JoshSmith I can create an issue if you think it makes sense to do so. The behavior would be similar to what we're doing with donation goals.

## References
Fixes #644

Progress on: https://github.com/code-corps/code-corps-api/pull/372


JoshSmith:
@begedin sure, create an issue, but make it low priority.

app/routes/project/tasks/task.js
```diff
@@ -2,40 +2,23 @@ import Ember from 'ember';
...
 
-  setupController(controller, models) {
-    this._super(controller, models);
-    controller.set('task', models.task);
-    controller.set('newComment', models.comment);
-    controller.set('comments', models.comments);
+  setupController(controller, task) {
+    let user = this.get('currentUser.user');
+    let newComment = this.store.createRecord('comment', { user });
+    return controller.setProperties({ newComment, task });
```
sbatson5: I think the `return` here isn't necessary.  `setupController` does not return a value by default:
http://emberjs.com/api/classes/Ember.Route.html#method_setupController

It should be fine to just do `controller.setProperties({ newComment, task });` without the `return`

app/routes/project/tasks/task.js
```diff
@@ -2,40 +2,23 @@ import Ember from 'ember';
...
+    return this.store.queryRecord('task', { projectId, number });
   },
 
-  setupController(controller, models) {
-    this._super(controller, models);
-    controller.set('task', models.task);
-    controller.set('newComment', models.comment);
-    controller.set('comments', models.comments);
+  setupController(controller, task) {
+    let user = this.get('currentUser.user');
```
sbatson5: This is out of the scope of this PR, but I never like fetching properties directly from Services. It would be nice to have a `getUser` method:
`this.get('currentUser').getUser();`

JoshSmith: @sbatson5 so we don't lose this, would you want to open an issue here or do a quick PR for this?

Issue #448, opened 2016-09-19
Add ember-i18n


JoshSmith:
# Problem

We want to avoid duplication of content and eventually allow for internationalization.
## Subtasks
- [ ] Add i18n
- [ ] Deduplicate content in head tags
- [ ] Research which parts of templates could be i18n-ed

LBKinson:
hey @JoshSmith, would love to contribute on this. I have a fair amount of experience with the ember-i18n addon. Can you clarify what "deduplicate content in head tags" means? I'm not seeing double info.

For templates: all hard coded verbage can be internationalized. I'll submit a PR for just one page so you can see and then I can move forward on converting all templates if that looks good.

JoshSmith:
@LBKinson it looks like what I meant by that is the duplication of `content` in the application route of say `meta-description` and `meta-og-description`, among others. Those could certainly be candidates for other hard-coded verbiage that's removed.

LBKinson:
Ohhh gotcha. Where `meta-description` and `meta-og-description` have the same content. Yes, those can definitely be replaced by one line item in the translation file.

JoshSmith:
Yeah, I'd almost forgotten what I meant, but that's certainly what I was hoping!

JoshSmith:
Per @LBKinson on the (now-merged) PR, we still need to do the following:

Needs:
- [ ] translations on all other templates
- [ ] translations on any dynamically generated messages (i.e. errors) within routes and components
- [ ] UI design for a user to change their locale (example below)

<img width="262" alt="screenshot 2016-10-26 22 39 58" src="https://cloud.githubusercontent.com/assets/4722967/19755826/33b427cc-9bcd-11e6-9c4d-577532e241f6.png">
<img width="288" alt="screenshot 2016-10-26 22 40 08" src="https://cloud.githubusercontent.com/assets/4722967/19755831/35d5bf16-9bcd-11e6-8ccf-23fb20f7ba09.png">

PR #648, opened 2016-10-27
issue #448: utilizing ember-i18n addon

LBKinson:
Work in progress.

Currently:
- installed ember-i18n addon
- declared locale detection in application route
- employed translations on `/about` template as example

Needs:
- [ ] translations on all other templates
- [ ] translations on any dynamically generated messages (i.e. errors) within routes and components
- [ ] UI design for a user to change their locale (example below)

screenshot 2016-10-26 22 39 58
screenshot 2016-10-26 22 40 08

### DIFF
--- a/app/locales/en-us/config.js
+++ b/app/locales/en-us/config.js
@@ -1,0 +1,16 @@
+// Ember-I18n includes configuration for common locales. Most users
+// can safely delete this file. Use it if you need to override behavior
+// for a locale or define behavior for a locale that Ember-I18n
+// doesn't know about.
+export default {
+  // rtl: [true|FALSE],
+  //
+  // pluralForm: function(count) {
+  //   if (count === 0) { return 'zero'; }
+  //   if (count === 1) { return 'one'; }
+  //   if (count === 2) { return 'two'; }
+  //   if (count < 5) { return 'few'; }
+  //   if (count >= 5) { return 'many'; }
+  //   return 'other';
+  // }
+};
--- a/app/locales/en-us/translations.js
+++ b/app/locales/en-us/translations.js
@@ -1,0 +1,27 @@
+export default {
+  'about-us': {
+    'we-help': 'We help people build simple software<br/>that solves hard problems.',
+    'we-were-promised': 'We were promised flying cars and we got targeted advertising.',
+    'where-are-the-apps': 'Where are the apps to challenge poverty or end climate change?',
+    'we-want-to-build': 'We want to build these things not because they are easy,<br/>but because they are hard. Here\'s our plan:',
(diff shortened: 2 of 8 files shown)


JoshSmith:
@LBKinson yay! This is so awesome to see here. Thanks so much for kicking this off.

One thing to consider is UI-related portions of the application that consume English-language responses from the server. I know that's outside the scope of Ember, but is certainly going to make things confusing to a user who thinks they've internationalized.

At the very, _very_ least, this does make it easier for copywriters to come in and tweak things without having to jump directly into HTML.

LBKinson:
@JoshSmith not sure what you mean, do you have an example of "UI-related portions of the application that consume English-language responses from the server"?

also, excellent linter, I'll be sure to fix those circle errors :)

JoshSmith:
@LBKinson I mean things like our JSON API errors. http://docs.codecorpsapidevelop.apiary.io/#reference/authentication/json-web-token-exchange/exchange-email-and-password-for-token like the 401 responses from this endpoint, for example.

begedin:
@JoshSmith We might want to, at some point in the future, consider not displaying error strings the server serves. Instead, we could rely on error IDs and use client-stored strings. It makes sense for the client do decide how to notify user of errors.

JoshSmith:
@LBKinson did we want to break this PR up into smaller pieces and have a tracking issue for those?

begedin:
I would also support merging this (if ready) as a first step.

We can then create issues to localize everything.

Additionally, I have an idea on writing a custom locale service for testing, with a set of custom assertions. This would allow us to

1. Check that there is a localized string present on the page/in the component template.
2. Check that there is indeed a key for it in the locale file.
3. Not check for the exact content of the key, just to avoid writing it twice.

JoshSmith:
Hey @LBKinson looks like the rebase brought in some unintentional changes?

LBKinson:
@JoshSmith I rebased off /develop, is there a different branch I should pull in?

JoshSmith:
@LBKinson no, that's the right one. But sometimes the rebase process for a fork isn't as straightforward as it should be. My guess is that you rebased on the upstream's origin develop, but your own develop was out of sync.

JoshSmith:
@LBKinson is this ready for a review right now or still in progress?

JoshSmith:
Hi @LBKinson I'd definitely like to bring this but wasn't sure if you were still planning to loop back more in the context of this PR. Just let me know so I know for sure what direction to take this!

JoshSmith:
Thanks for working on this! Sorry we weren't ever able to get more done in this PR, and would always love to do more here to incorporate this throughout the project.

Issue #507, opened 2016-09-29
Can hit submit button multiple times in various parts of the app


JoshSmith:
# Problem

Many instances of the app where you're `POST`-ing to a server you can hit the submit button multiple times. The button is not disabled and the form submission is a repeatable process.

This is bad because the app can get into a weird state, like in onboarding where multiple `PATCH` requests may come through and the `state_transition` is outdated by the time later requests hit the server.

## Subtasks
- [ ] Identify all instances where a button does not become disabled or show proper loading state to the user
- [ ] Create a proper async button (or use an external library)
- [ ] Replace instances of these buttons with the new behavior
- [ ] Test, test, test

WenInCode:
We had a similar issue in an app I worked on elsewhere. Their approach was to create a mixin to prevent double saves. It was pretty nice as you could plug it pretty much anywhere. But the saving was pretty consistent everywhere as well. Just wanted to mention it.

JoshSmith:
@WenInCode not a bad idea. If you have some thoughts on whether that's workable here (after we identify those instances so we know our baseline) that would be great.

sbatson5:
We had a similar issue on our app. A really simple/easy solution we found was binding the `disabled` attribute of our button to the `isSaving` property of the model. Our example only had 1 model on the page, but it seemed to work well.

JoshSmith:
@sbatson5 that's a pretty solid approach, too. Thanks for that!

sbatson5:
Happy to PR something this weekend 👍 . I love this hacktoberfest stuff! I just come across this repo and as an Elixir/Ember dev, this sounds great!

JoshSmith:
@sbatson5 feel free to stop by [our Slack](http://slack.codecorps.org) and :wave: hello, too!

begedin:
# Identified instances where we don't handle it:

* `start/{hello, interests, skills, expertise}.hbs` - can spam click next step, will fail due to trying to transition into invalid state. Should probably bind `disable` to `model.isSaving`, or we can just have the continue action return if model is currently saving. Other option might even be better, since short flashes of a disabled state on the button can be perceived badly. The `continue` action is defined in `mixins/onboarding-controller.js`
* `components/create-comment-form` - has just a class binding: `class="default small right {{if comment.isSaving "disabled"}}"` which is not enough. User can still click, it just looks like they cant. Pretty sure you can end up creating multiple identical comments by spam-clicking here.
[...]

PR #650, opened 2016-10-28
use ember-concurrency to handle multiple signup requests

sbatson5:
# What's in this PR?

We use `ember-concurrency` in a couple of places in our app already. We should leverage tasks from `ember-concurrency` to solve the problem of trying to submit data multiple times.  This PR addresses the signup form and ensures the user can't spam the submit button.
## References

Progress on: #507 
Progress on: #336

### DIFF
--- a/app/components/signup-form.js
+++ b/app/components/signup-form.js
@@ -1,10 +1,10 @@
 import Ember from 'ember';
+import { task } from 'ember-concurrency';
 
 const {
   Component,
   computed,
   computed: { alias, and, gte },
-  inject: { service },
   run: { later }
 } = Ember;
 
@@ -14,9 +14,6 @@ export default Component.extend({
   hasError: false,
   usernameValid: false,
 
-  session: service(),
-  store: service(),
-
   canSubmit: and('emailValid', 'passwordValid', 'usernameValid'),
   passwordLength: alias('password.length'),
   passwordValid: gte('passwordLength', 6),
@@ -25,11 +22,6 @@ export default Component.extend({
     return this.get('user.password') || '';
   }),
 
-  init() {
-    this._super(...arguments);
-    this.set('user', this.get('store').createRecord('user'));
-  },
-
   actions: {
     emailValidated(result) {
       this.set('emailValid', result);
@@ -37,7 +29,7 @@ export default Component.extend({
 
     signUp() {
       if (this.get('canSubmit')) {
-        this._submit();
+        this.get('_submit').perform();
       } else {
         this._shakeButton();
       }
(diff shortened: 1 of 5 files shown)


app/components/signup-form.js
```diff
@@ -67,20 +53,18 @@ export default Component.extend({
...
-
-      if (!payloadContainsValidationErrors) {
-        this.controllerFor('signup').set('error', error);
-      }
+      this.get('handleErrors')(error);
     });
-  }
+
+    yield promise;
+  }).drop()
```
sbatson5: From the `ember-concurrency` documentation:

> The drop modifier `drops` tasks that are `.performed` while another is already running. Dropped tasks' functions are never even called.

app/components/signup-form.js
```diff
@@ -25,19 +22,14 @@ export default Component.extend({
     return this.get('user.password') || '';
   }),
 
-  init() {
-    this._super(...arguments);
-    this.set('user', this.get('store').createRecord('user'));
```
sbatson5: Moved this to the route so we don't have to inject the store in the component anymore.

sbatson5:
@JoshSmith let me know if this looks ready and I will squash

tests/unit/routes/signup-test.js
```diff
@@ -1,11 +1,39 @@
...
-  needs: ['service:metrics']
+  needs: ['service:metrics', 'service:session']
 });
 
 test('it exists', function(assert) {
   let route = this.subject();
   assert.ok(route);
 });
+
+test('it renders erros through controller', function(assert) {
```
JoshSmith: `erros` needs fixed.

app/components/signup-form.js
JoshSmith: @sbatson5 I'm a little confused by this. Does this mean that a request will still continue on to the server?

app/templates/signup.hbs
```diff
@@ -1,3 +1,5 @@
 {{title "Sign up"}}
 
-{{signup-form}}
+{{signup-form user=model
```
JoshSmith: Do we want to place `user=model` on a separate line? Alpha order these?

app/components/signup-form.js
sbatson5: @JoshSmith Yes. It means that any additional clicks (after the first one) will be "dropped." So we should never see more than 1 request going out.

app/templates/signup.hbs
sbatson5: Can do!

sbatson5:
That's a good question @begedin. Not sure how we could test concurrency in something like this. And even if we did, aren't we just testing `ember-concurrency` and not our code?

Also, rebased and squashed.

PR #654, opened 2016-10-31
add transform for cent value currency

sbatson5:
# What's in this PR?

Spurred by the conversation here: https://github.com/code-corps/code-corps-ember/pull/634#issuecomment-257439956

If we want to deal with integer/cent values from our API but still have them reflect like decimals on the frontend, it may be easiest to create a custom transform to handle the currency conversation.  This would allow us to do:

`amount: attr('dollar-cents')`

And not have to do any cleaning of user input.

### DIFF
--- a/app/transforms/dollar-cents.js
+++ b/app/transforms/dollar-cents.js
@@ -1,0 +1,13 @@
+import DS from 'ember-data';
+
+const { Transform } = DS;
+
+export default Transform.extend({
+  deserialize(serialized) {
+    return serialized  / 100;
+  },
+
+  serialize(deserialized) {
+    return Math.floor(deserialized * 100);
+  }
+});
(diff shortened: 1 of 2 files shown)


app/transforms/dollar-cents.js
```diff
@@ -0,0 +1,13 @@
...
+
+const { Transform } = DS;
+
+export default Transform.extend({
+  deserialize(serialized) {
+    return serialized  / 100;
+  },
+
+  serialize(deserialized) {
+    return Math.floor(deserialized * 100);
```
JoshSmith: Are there any potential issues we can run into with the math here? Potential rounding errors?

begedin: `.floor` doesn't leave much room for it. Anything smaller than a cent is simply removed when posting to API. API deals with integers, so there will never be anything smaller than a cent sent to us.

PR #659, opened 2016-11-02
Added donation-progress component

### LINKED ISSUE #471: Add donation progress component
# Problem

The donation progress component will be a sub-component of the project's donation UI seen here https://projects.invisionapp.com/share/SP8PSQGNM
## Subtasks
- [ ] Take in the project's current goal
- [ ] Display the monthly amount for the current goal
- [ ] Calculate the percentage for the current goal
- [ ] Display the percentage for the current goal
- [ ] Display a progress bar using the percentage amount
- [ ] Display the current goal's description

begedin:
# What's in this PR?

Work in progress, but almost there. Still need to show description.

* [x] Display description of current goal

When we have amounts with zero cents, our `format-currency` helper shows those zeros which does not look nice. For that purpose, I added an option to the helper to trim those zeros if there are any. We might want to get rid of the option and just make it default, though.

Similarly, I implemented a `format-percentage` helper that behaves the same. Both helpers need testing for this feature

* [x] Tests for changes in `format-currency`
* [x] Tests for `format-percentage`

The implementation makes use of the existing `progress-bar-container` component.

## References
Closes #471
Progress on: #469 

https://projects.invisionapp.com/share/SP8PSQGNM

## Subtasks (taken from issue)
- [x] Take in the project's current goal
- [x] Display the monthly amount for the current goal
- [x] Calculate the percentage for the current goal
- [x] Display the percentage for the current goal
- [x] Display a progress bar using the percentage amount
- [x] Display the current goal's description

### DIFF
--- a/app/helpers/format-currency.js
+++ b/app/helpers/format-currency.js
@@ -1,10 +1,19 @@
 import Ember from 'ember';
+import prettyFloat from 'code-corps-ember/utils/pretty-float';
 
 const {
   Helper: { helper }
 } = Ember;
 
 const COMMAS_EVERY_THREE_DIGITS = /(\d)(?=(\d{3})+(?!\d))/g;
+
+function applyFormatting(dollars) {
+  if (dollars.length == 0) {
+    return '';
+  } else {
+    return `$${dollars.replace(COMMAS_EVERY_THREE_DIGITS, '$1,')}`;
+  }
+}
 
 /**
  * Used to display an amount (in dollars) as currency in the format of
@@ -15,20 +24,16 @@ const COMMAS_EVERY_THREE_DIGITS = /(\d)(?=(\d{3})+(?!\d))/g;
  * - dot as decimal separator
  * - fixed 2-decimal notation
  *
- * @param  {Array[String]} params An array of paramaters provided to the helper in the template
- * @return {String}        Amount formated as $20,000.99
+ * In order to not display zero decimals, provide a hash, consisting of
+ * `{ trimZero: true }` as the second parameter.
+ *
+ * @param  {String}  dollarsAsString   The value to be formatted as currency
+ * @param  {Boolean} options.trimZero  Indicates if zero decimals should be removed
+ * @return {String}                    The value, formatted as currency
  */
-export function formatCurrency(params) {
-  let [dollarsAmountAsString] = params;
-  let dollarsAmount = parseFloat(dollarsAmountAsString);
-
-  if (dollarsAmount) {
-    return `$${dollarsAmount.toFixed(2).replace(COMMAS_EVERY_THREE_DIGITS, '$1,')}`;
-  } else if (isNaN(dollarsAmount)) {
-    return '';
-  } else {
-    return dollarsAmount;
-  }
+export function formatCurrency([dollarsAsString], { trimZero = false } = {}) {
+  let dollarsAmount = prettyFloat(dollarsAsString, { trimZero });
+  return applyFormatting(dollarsAmount);
 }
 
 export default helper(formatCurrency);
--- a/app/helpers/format-percentage.js
+++ b/app/helpers/format-percentage.js
@@ -1,0 +1,32 @@
+import Ember from 'ember';
+
+import prettyFloat from 'code-corps-ember/utils/pretty-float';
+
+const {
+  Helper: { helper }
+} = Ember;
+
+function applyFormatting(percentage) {
+  return percentage.length == 0 ? '' : `${percentage}%`;
+}
+
+/**
+ * Helper formats a float or a string representing a float into a percentage value.
+ *
+ * By default, the value will be formatted as `'22.5%'`.
+ * If the number is an integer, the format will be `'22.0%'`
+ *
+ * To remove the zero deciamals, a "trimZero" named argument can be provided.
+ * If this argument is set to `true`, the zero decimals
+ * will be removed, so the format will be `'22%'`
+ *
+ * @param  {String}   percentageAsString The value to be formatted as percentage
+ * @param  {Boolean}  options.trimZero   Option to remove the decimal part if the value of it is zero
+ * @return {String}                      Value, formatted as percentage
+ */
+export function formatPercentage([percentageAsString], { trimZero = false } = {}) {
+  let percentage = prettyFloat(percentageAsString, { numDecimals: 1, trimZero });
+  return applyFormatting(percentage);
+}
+
+export default helper(formatPercentage);
--- a/app/styles/components/donations/donation-progress.scss
+++ b/app/styles/components/donations/donation-progress.scss
@@ -1,0 +1,24 @@
+.donation-progress {
+  .information {
+    display: flex;
+    justify-content: space-between;
+
+    .percentage {
+      text-align: right;
+    }
+
+    .descriptive-text {
+      color: $gray;
+    }
+
+    .value {
+      font-size: $header-font-size-large;
+      font-weight: 600;
+    }
+  }
+
+  .progress-bar-container {
+    height: 10px;
+    margin-top: 10px;
+  }
+}
--- a/app/templates/components/donations/donation-progress.hbs
+++ b/app/templates/components/donations/donation-progress.hbs
@@ -1,0 +1,12 @@
+<div class="information">
+  <div class="amount">
+    <div class="value">{{format-currency amountDonated trimZero=true}}</div>
+    <div class="descriptive-text">given each month</div>
+  </div>
+  <div class="percentage">
+    <div class="value">{{format-percentage percentage trimZero=true}}</div>
+    <div class="descriptive-text">of goal</div>
+  </div>
+</div>
+{{progress-bar-container percentage=percentage}}
+<div class="description">{{donationGoal.description}}</div>
--- a/tests/integration/components/donations/donation-progress-test.js
+++ b/tests/integration/components/donations/donation-progress-test.js
@@ -1,0 +1,46 @@
+import { moduleForComponent, test } from 'ember-qunit';
+import hbs from 'htmlbars-inline-precompile';
+
+import PageObject from 'ember-cli-page-object';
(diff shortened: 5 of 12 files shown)


begedin:
This should be good to go. The component receives a donation goal and a value indicating the amount donated towards that goal. I'm not really sure where to get that value from. I'm guessing it may even eventually be a computed field on the donation goal itself. Either way, that would be a quick change once we integrate the component into the app.

app/helpers/format-currency.js
```diff
@@ -15,20 +24,16 @@ const COMMAS_EVERY_THREE_DIGITS = /(\d)(?=(\d{3})+(?!\d))/g;
...
-  let dollarsAmount = parseFloat(dollarsAmountAsString);
-
-  if (dollarsAmount) {
-    return `$${dollarsAmount.toFixed(2).replace(COMMAS_EVERY_THREE_DIGITS, '$1,')}`;
-  } else if (isNaN(dollarsAmount)) {
-    return '';
-  } else {
-    return dollarsAmount;
-  }
+export function formatCurrency([dollarsAsString, { trimZero = false } = {}]) {
```
begedin: I actually like that way more than a hash. However, how would be pass in a named argument when using the helper in a template, such as we do right now? Are there any difficulties there, or can I still continue to use a hash helper. 

Is there a better way to have named arguments for helpers in general?

sbatson5: You wouldn't use the hash helper when using `namedArgs`. It would be `{{format-currency amount trimZero=true}}`. Ember will put all named arguments into a hash for us and pass that as the second argument for us

begedin:
Should be good for another go at review @sbatson5  @JoshSmith

app/helpers/format-percentage.js
```diff
@@ -0,0 +1,32 @@
...
+
+function applyFormatting(percentage) {
+  return percentage.length == 0 ? '' : `${percentage}%`;
+}
+
+/**
+ * Helper formats a float or a string representing a float into a percentage value.
+ *
+ * By default, the value will be formatted as `'22.25%'`.
+ * If the number is an integer, the format will be `'22.00%'`
```
JoshSmith: I know this adds some complexity, but I would prefer the percentage to have only a single decimal place., so `22.3%` instead of `22.25%`. The latter is just too much detail.

app/styles/components/donations/donation-progress.scss
```diff
@@ -0,0 +1,24 @@
+.donation-progress {
+  .information {
+    display: flex;
+    justify-content: space-between;
+
+    .value {
+      font-weight: 600;
+      font-size: 200%;
```
JoshSmith: A `200%` size is really unclear here. What does this mean?

app/styles/components/donations/donation-progress.scss
```diff
@@ -0,0 +1,24 @@
...
+    .percentage {
+      text-align: right;
+    }
+  }
+
+  .progress-bar-container {
+    margin-top: 10px;
+    height: 10px;
+  }
+}
```
JoshSmith: There are some alpha-ordering issues here. Would do a general sweep yourself of some of these.

app/templates/components/donations/donation-progress.hbs
```diff
@@ -0,0 +1,12 @@
...
+    <div class="value">{{format-currency amountDonated trimZero=true}}</div>
+    <div class="descriptive-text">given each month</div>
+  </div>
+  <div class="percentage">
+    <div class="value">{{format-percentage percentage trimZero=true}}</div>
+    <div class="descriptive-text">of goal</div>
+  </div>
+</div>
+{{progress-bar-container percentage=percentage}}
+<div class="description">{{donationGoal.description}}</div>
```
JoshSmith: We're using what I think are too many classes here. Are you able to rely on other primitives like HTML5 tags that are good candidates here?

tests/integration/components/donations/donation-progress-test.js
```diff
@@ -0,0 +1,34 @@
...
+  assert.expect(3);
+
+  let mockGoal = { amount: 1000, description: 'Lorem ipsum' };
+  this.set('donationGoal', mockGoal);
+
+  this.render(hbs`{{donations/donation-progress donationGoal=donationGoal amountDonated=500.50}}`);
+
+  assert.equal(this.$('.amount .value').text().trim(), '$500.50', 'Correct amount is rendered');
+  assert.equal(this.$('.percentage .value').text().trim(), '50.05%', 'Correct percentage is rendered');
+  assert.equal(this.$('.progress-bar').length, 1, 'Progress bar component is rendered');
```
JoshSmith: I think some of these should be wrapped into a page object for the component.

begedin: We already briefly mentioned this in another PR. We do not, currently, use page objects in any integration test. They are acceptance tests only. I get the advantage of using it, but it would be a precedent, so I'm not sure if it should be done here or as a separate PR. We would also have to consider if such page objects need a different structure, or if we can just reuse component page objects used in acceptance tests.

app/templates/components/donations/donation-progress.hbs
begedin: We could make the value an `<h{1,2,3}>` tag and the description a `<p>` tag. I personally prefer classes, though. Our HTML is not that complex that using classes would slow selection down, and the advantage is that we need to update tests less.

If we make the value an `<h2>` now, but decide that it needs to be an `<h3>` or even a `<div>` later, if that element keeps the class, the test will till pass. If we rely on a tag, though, the test will fail and have to be updated, even though the app itself still works.

Additionally, it's not really a header, so it doesn't make much sense to use an h tag just for the sake of not using a class. Really, I can't think of a tag closer fitting than a div. It's a text, displayed on it's own line, styled in a certain way. It's not a paragraph, or a header. We can make it, maybe a `<strong>`, but then we additionally need to make it display as block. If anything, it would be more appropriate to make it a `<strong>` wrapped in a `<div>`, but that would just add to the complexity.

tests/integration/components/donations/donation-progress-test.js
begedin: I looked further into this and tried to create a solution that makes sense to me.

This is the documentation for using page objects in integration tests:

http://ember-cli-page-object.js.org/docs/v1.6.x/quickstart#integration-tests

As you can see, the idea is that we have a page object for the acceptance tests and then reuse it in integration tests.

Thing is, we do not have such a page object yet, and to me, it seems like, if we define a full-fledged page object for acceptance testing of the page that would contain this component, it's misleading to use it for testing the component alone.

Instead, my idea was to define `pages/components/donations/donation-progress.js` as a page object component. We then import that component into our test and create a page object using `PageObject.create(donationProgressComponent)`.

Additionally, we have to set and remove it's context before and after each test. That being said, the tests are much cleaner that way, so I like the solution. It's just that I would also love if there was a way to wrap the whole process into a helper of some sort, possibly something like

```JavaScript
moduleForComponentUsingPageObject(path, moduleName, pageObjectComponent) {
  let page = PageObject.create(pageObjectComponent);
  moduleForComponent(path, moduleName, {
    integration: true,
    beforeEach: function() { page.setContext(this); },
    afterEach: function() { page.removeContext(this) }
  })

  return page;
}
```

Note that we would have to return the page, since we then need to access it in our tests.

We would use this helper along the lines of 

```Javascript
import moduleForComponentUsingPageObject from '../../../helpers/module-for-component-using-page-object';
import donationProgressComponent from '../../../pages/components/donations/donation-progress';

let page = moduleForComponentUsingPageObject('donations/donation-progress', 'Integration | Component | donations/donation progress', donationProgressComponent);
```

Compared to the current 

```Javascript
import PageObject from 'ember-cli-page-object';
import donationProgressComponent from '../../../pages/components/donations/donation-progress';

let page = PageObject.create(donationProgressComponent);

moduleForComponent('donations/donation-progress', 'Integration | Component | donations/donation progress', {
  integration: true,
  beforeEach() {
    page.setContext(this);
  },

  afterEach() {
    page.removeContext();
  }
});
```

I'm not sure it's that much cleaner, but it slightly cleaner, and it definitely is shorter (3 lines, compared to 13). 

That being said, there would be additional complications with such a helper, if we want to use custom before and after contexts.

app/helpers/format-percentage.js
begedin: It is a tad more complicated that way, but while implementing this, I found a bug, so added gain!

I was checking if there are any decimals **before** rounding. This means that

`22.001` would display as `22.00`, even though we trim decimals, because `22.001 - 22 > 0`. Now we check if there are decimals after rounding to determine if we should trim.

This means that `prettyFloat(22.001, { trimZero: true }) == '22'`, while `prettyFloat(22.006, { trimZero: true }) == '22.01'`

app/templates/components/donations/donation-progress.hbs
JoshSmith: Maybe it's the naming structure itself that I find problematic. Something feels wrong here every time I look at the conventions we're using in our styling.

sbatson5: Thoughts on following something like BEM for styling where child elements clearly adopt from their parents?  So, rather than a general `description` tag, we could have its class name based on its wrapper?

```
<div class="donation-goal">
  <h1 class="donation-goal__header">Lorem Ipsum</h1>
  <div class="donation-goal__description">some description</div>
</div>
```
Then in our sass styles, we can follow the convention of:

```
.donation-goal {
  color: whatever;

  &__header {
    font-size: 50px
  }

  &__description {
     font-family: some-font, sans-serif
  }
}
```

It adds explicitness to the styles and clarity into where elements are used.

JoshSmith: @sbatson5 what is BEM?

sbatson5: http://getbem.com/introduction/

Issue #472, opened 2016-09-25
Add donation project sidebar component


JoshSmith:
# Problem

We need to add the component seen on https://projects.invisionapp.com/share/SP8PSQGNM, https://projects.invisionapp.com/share/SP8PSQGNM#/screens/191333611, and https://projects.invisionapp.com/share/SP8PSQGNM#/screens/191332617
## Subtasks
- [ ] Show the sub-component of donation progress #471 
- [x] Show success state with amount you give each month: `You give $10 each month`.
- [x] Display "Become a donor" button by default
- [x] Clicking the "Become a donor" button slides down to reveal donation UI
- [x] Four buttons: "Donate $10", $15, $25, $50
- [x] Custom amount per month field that only accepts integers
- [x] Continue button that takes the amount and passes it onto the next page `/projects/donate?amount=number`

JoshSmith:
We can use `accounting.js` for changing the dollar numbers into well-formatted cents integers to post to the API. http://openexchangerates.github.io/accounting.js/

begedin:
I'm thinking we have multiple components here, each of which renders based on what the model state is.

* If we have no donation model loaded, then we render  `{{donations/become-a-donor}}`
* If we have a donation model, but it's a new record, means we just clicked "become a donor", so we render `{{create-donation}}`
* If we have a donation model and it's a persisted record, mean we already donated, so we render `{{donation-item}}` or something.

JoshSmith:
@begedin do you mean an active subscription in your last point?

begedin:
Somehow edited your reply instead of replying myself. 

Really, I mean "active subscription" on all 3 points. I'll correct.

JoshSmith:
Closing in favor of #679

PR #662, opened 2016-11-04
Add donation status to sidebar

begedin:
# What's in this PR?

Still a work in progress.

* [x] Add parent `donations/donation-status` component which renders different child component based on model.
  * If there is no donationl, render `{{donations/become-a-donor}}`
  * If there is an unsaved donation, render `{{donations/create-donation}}`
  * If there is a persisted donation, render `{{donations/show-donation}}`
* [x] Implement `{{donations/become-a-donor}}` 
* [x] Implement `{{donations/create-donation}}`
* [x] Implement  `{{donations/show-donation}}`
* [x] Add to route template
* [x] Bind to route/controller actions
* [x] Slide down when clicking "Become a Donor"
* [ ] Add icon for "you plegded x amount"
* [x] Go through tests and add missing assert descriptions
* [ ] Disable continue button if amount is blank, + tests for that behavior

## References
Progress on: #472
Progress on: #469 

## ToDos copied from issue

- [ ] Show the sub-component of donation progress #471 
- [x] Display "Become a donor" button by default
- [x] Clicking the "Become a donor" button slides down to reveal donation UI
- [x] Four buttons: "Donate $10", $15, $25, $50
- [x] Custom amount per month field that only accepts integers
- [x] Continue button that takes the amount and passes it onto the next page `/projects/donate?amount=number`
- [x] Show success state with amount you give each month: `You give $10 each month`.

### DIFF
--- a/app/controllers/project/index.js
+++ b/app/controllers/project/index.js
@@ -1,0 +1,22 @@
+import Ember from 'ember';
+
+const {
+  Controller,
+  computed,
+  inject: { service }
+} = Ember;
+
+export default Controller.extend({
+  store: service(),
+  currentUser: service(),
+
+  user: computed.alias('currentUser.user'),
+
+  actions: {
+    saveSubscription(amount) {
+      let project = this.get('project');
+      let queryParams = { amount };
+      this.transitionToRoute('project.donate', project, { queryParams });
+    }
+  }
+});
--- a/app/models/stripe-subscription.js
+++ b/app/models/stripe-subscription.js
@@ -1,0 +1,10 @@
+import Model from 'ember-data/model';
+import attr from 'ember-data/attr';
+import { belongsTo } from 'ember-data/relationships';
+
+export default Model.extend({
+  amount: attr('dollar-cents'),
+
+  project: belongsTo('project', { async: true }),
+  user: belongsTo('user', { async: true })
+});
--- a/app/routes/project/index.js
+++ b/app/routes/project/index.js
@@ -1,0 +1,36 @@
+import Ember from 'ember';
+
+const {
+  inject: { service },
+  Route,
+  RSVP
+} = Ember;
+
+export default Route.extend({
+  currentUser: service(),
+
+  model() {
+    let project = this.modelFor('project');
+
+    return RSVP.hash({
+      project,
+      subscription: this._fetchCurrentUserSubscriptionFor(project)
+    });
+  },
+
+  setupController(controller, models) {
+    controller.setProperties(models);
+  },
+
+  _fetchCurrentUserSubscriptionFor(project) {
+    let user = this.get('currentUser.user');
+
+    if (user) {
+      return user.get('subscriptions').find((subscription) => {
+        subscription.belongsTo('project').id == project.get('id');
+      });
+    } else {
+      return null;
+    }
+  }
+});
--- a/app/templates/components/donations/donation-status.hbs
+++ b/app/templates/components/donations/donation-status.hbs
@@ -1,0 +1,7 @@
+{{#if subscription}}
+  {{donations/show-donation amount=subscription.amount}}
+{{else if processStarted}}
+  {{donations/create-donation amount=15 continue=(action createDonation)}}
+{{else}}
+  {{donations/become-a-donor becomeADonor=(action (mut processStarted) true)}}
+{{/if}}
--- a/tests/integration/components/donations/become-a-donor-test.js
+++ b/tests/integration/components/donations/become-a-donor-test.js
@@ -1,0 +1,49 @@
+import { moduleForComponent, test } from 'ember-qunit';
+import Ember from 'ember';
+import hbs from 'htmlbars-inline-precompile';
+import PageObject from 'ember-cli-page-object';
+
+import becomeADonorComponent from '../../../pages/components/donations/become-a-donor';
+
+const {
+  K
+} = Ember;
+
+let page = PageObject.create(becomeADonorComponent);
+
+function setHandler(context, becomeADonorHandler = K) {
+  context.set('becomeADonorHandler', becomeADonorHandler);
+}
+
+moduleForComponent('donations/become-a-donor', 'Integration | Component | donations/become a donor', {
+  integration: true,
+  beforeEach() {
+    setHandler(this);
+    page.setContext(this);
+  },
+  afterEach() {
+    page.removeContext();
+  }
+});
+
+test('it renders button', function(assert) {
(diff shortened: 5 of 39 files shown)


tests/pages/components/donations/create-donation.js
```diff
@@ -0,0 +1,21 @@
+import { clickable, fillable, isVisible, value } from 'ember-cli-page-object';
+
+export default {
+  scope: '.create-donation',
+
+  rendersSetTo10Button: isVisible('button.preset-amount.10'),
+  rendersSetTo15Button: isVisible('button.preset-amount.10'),
+  rendersSetTo25Button: isVisible('button.preset-amount.10'),
+  rendersSetTo50Button: isVisible('button.preset-amount.10'),
```
JoshSmith: These need fixed.

begedin:
Started integrating this into routes and encountered a difficulty.

When we go to the project about page, we want to see if the current user has an active subscription for the current project. With our API the way it is right now:

```
# models/project
# not sure if has_many :plans or belongs_to :plan, we didn't define it yet

# models/stripe_plan
belongs_to :project

# models/user
has_many: :stripe_subscriptions

# models/stripe_subscription
belongs_to :user
belongs_to :plan
```

What we would have to do is

* get current user
* get their subscriptions
* see if any of them has a plan that belongs to the loaded project

The way I see it, the API doesn't make sense to me. I'm thinking we have a single, global plan which everyone uses. There is no need to link it to anything else in our app, other than a subscription. This would be a 1$ plan.

We then link the subscription to a project and a user, as well as an x amount of our global plan.

Once and if we want to allow for cent donations, we simply create a second, 1 cent plan, which we use alongside our $1 plan.

JoshSmith:
I removed the scrolling behavior in here. I think it was based on a misunderstanding and the originally intended UX is far from MVP. Also have simplified the CSS to use some existing styles and rely more heavily on flexbox.

JoshSmith:
Let's try to get merged ASAP and handle the rest in separate PRs if possible. This has become unwieldly.

begedin:
@JoshSmith This should be good to go. Parts that need to be done separate:

* [ ] Add icon for `show-donation-goal` component
* [ ] Add `disabled` behaviour test for when custom amount is 0. We also need a style for the default button disabled 
* [ ] Add sanitizing of the custom amount input, so it only allows integers.

Not sure if these should go into a whole new issue, or if we should close this PR and just add these items to the existing issue. Your call.

app/controllers/project/index.js
```diff
@@ -0,0 +1,18 @@
+import Ember from 'ember';
+
+const { Controller, computed, inject: { service } } = Ember;
```
sbatson5: Think we could break this onto multiple lines? I find the `inject: { service }` hard to read when it is on one line with everything else.

app/controllers/project/index.js
```diff
@@ -0,0 +1,18 @@
...
+  store: service(),
+  currentUser: service(),
+
+  user: computed.alias('currentUser.user'),
+
+  actions: {
+    saveSubscription(amount) {
+      let project = this.get('project');
+      let queryParams = { amount };
+      this.transitionToRoute('project.donate', project, { queryParams });
```
sbatson5: Do we need `queryParams` to be in a hash? Isn't it already a hash with `amount`?  This would send it as:
```
{
  queryParams: {
    amount: x
  }
}
```

app/models/stripe-subscription.js
```diff
@@ -0,0 +1,10 @@
+import Model from 'ember-data/model';
+import attr from 'ember-data/attr';
+import { belongsTo } from 'ember-data/relationships';
+
+export default Model.extend({
+  amount: attr('dollar-cents'),
```
sbatson5: 👏 for using this transform

app/routes/project/index.js
```diff
@@ -0,0 +1,32 @@
+import Ember from 'ember';
+
+const { inject: { service }, RSVP, Route } = Ember;
```
sbatson5: Same comment as above. Do you prefer doing these on multiple lines?

app/templates/components/donations/donation-status.hbs
```diff
@@ -0,0 +1,7 @@
+{{#if subscription}}
+  {{donations/show-donation amount=subscription.amount}}
+{{else if processStarted}}
+  {{donations/create-donation subscription=subscription continue=(action createDonation) amount=15}}
```
sbatson5: Curious, how can we pass `subscription` here if it isn't present? If `subscription` exists, then the first `if` condition will be met and this component won't be rendered.  This component is only shown if _don't_ have a subscription.

tests/integration/components/donations/become-a-donor-test.js
```diff
@@ -0,0 +1,51 @@
...
+
+import becomeADonorComponent from '../../../pages/components/donations/become-a-donor';
+
+const {
+  K
+} = Ember;
+
+let page = PageObject.create(becomeADonorComponent);
+
+function setHandlers(context, { becomeADonorHandler = K } = {}) {
```
sbatson5: Not sure that the second param needs to be a hash here. Can we just do 
`setHandlers(context, becomeADonorHandler = K) { ... }`

Then just update line 46 to be `setHandlers(this, becomeADonorHandler)`, rather than wrapping it?

tests/integration/components/donations/create-donation-test.js
```diff
@@ -0,0 +1,84 @@
...
+
+import createDonationComponent from '../../../pages/components/donations/create-donation';
+
+const {
+  K
+} = Ember;
+
+let page = PageObject.create(createDonationComponent);
+
+function setHandlers(context, { continueHandler = K } = {}) {
```
sbatson5: Same as the other test. Not sure the second param here needs to be a hash since we only ever call one property from it.

app/controllers/project/index.js
begedin: Yeah, last parameter of `transitionToRoute` needs to be a hash with a `queryParams` property, which is also a hash.

http://emberjs.com/api/classes/Ember.Controller.html#method_transitionToRoute

It's at the end of that section

app/templates/components/donations/donation-status.hbs
begedin: Good catch, It's a remnant from the start of this PR, when the implementation was different.

begedin:
Ready for another pass. Bed for me soon, so if we want this merged, someone feel free to take over.

Issue #477, opened 2016-09-25
Finish thank you page


JoshSmith:
# Problem

We need a thank you page as seen here: https://projects.invisionapp.com/share/SP8PSQGNM#/screens/191332621

## Subtasks
- [x] Show #476 project header
- [x] Show thank you icon and text
- [x] Have a link back to the project
- [ ] Add src to the icon image
- [ ] Change alt text to be about the success
- [ ] Match text color/size/spacing everywhere
- [ ] Blocked by #678 – Show 12 random team members (who have actual photos, not the default icons)

JoshSmith:
I removed the "photos of contributors" to make this shippable more quickly.

begedin:
Mostly solved in #668 

All that's left is to add the icon src. Will update original description

WenInCode:
I'll take on finishing up this task

JoshSmith:
Closed by #786.

PR #668, opened 2016-11-10
Add new card component

### LINKED ISSUE #473: Add new credit card component

JoshSmith:
# What's in this PR?

Should implement:
- [x] Month/year dropdown
- [x] Button state changes to disabled/clear vs enabled/default based on valid parameters
- [x] Generate Stripe token for card when pressing donate https://github.com/code-corps/ember-stripe-service#creating-stripe-tokens-for-cards
- [x] Pressing donate takes Stripe token and POSTs to server to create subscription
- [x] Shows Stripe token creation and server errors in a list in a paragraph (not client-side validation errors) 
- [x] Pressing donate button shows loading state; changes on error or success
- [x] Create credit card when one is added. List them if user has any

**Copied stuff to add later into #669** so we do not have it listed under "incomplete subtasks" on the milestone.

## References
Fixes #473


begedin:
I got further on this one, but I had trouble with testing and handling it properly. Turns out, Ember does not detect the promise `stripe.card.createToken` generates, so it doesn't wait for it. It took me a while to figure out how to wait and test for it and I'm not extremely happy with the result, but it works for now. Simply put, I use `Ember.Test.registerWaiter`, to wait for an `isLoading` flag on the controller to be set to false. This means tighter coupling between the test and the code than I would like, but the alternative is to completely override the stripe service in our acceptance test, which sort of defeats the purpose of the acceptance test.

Then again, I'm up for trying that as well.

Either way, I got the process to complete, so now it would be a matter of rendering errors and deciding what to do on success. Do we want to just display a message, or redirect to a whole new route (ex. `project.thankyou`)?

JoshSmith:
@begedin do we need to be fixing the Stripe service? Open an issue there?

JoshSmith:
@begedin in particular we were already having conversations over there about this https://github.com/code-corps/ember-stripe-service/issues/12

begedin:
@JoshSmith I took a look at that issue already. Really, it's how ember behaves. I'm not really sure if we can do anything about it. Ember handles tracks some type of async stuff correctly, but not all of it. I stubbed out the stripe service for now, and that does the trick, but the downside is, if we change this addon, we will have to update the stub.  

The upside is, we're maintaining the addon.

begedin:
Updated the to-do list. It doesn't seem like much was checked, but I think we're past the hard work, so now it's a matter of writing the templates for errors and deciding what to do when the user successfully donates.

begedin:
@JoshSmith This should finally be good to review.

The "Thank you page just displays a message for now". We can add a template, which will be mostly static, separately.

begedin:
I've mostly implemented the thank you page. The layout matches the provided wireframe and all that's missing is setting the src of the icon image. Referring #477 here, since it's about 95% solved by this.

begedin:
Got as far as I could tonight.

* [ ] Add card ui is always shown. Should probably show it only if there are no cards and/or if the user clicks "add card"
* [ ] Add card UI doesn't go away after card was added. Should go away.
* If there are cards present, user can click to select one and then donate. This creates a subscription and redirects to thank you page.
* If there are no cards, user can add one. Once added, card is preselected and user can click donate to create subscription.

The process of adding cards needs more work:

* [ ] We create the token, but we do not use it anywhere. I think we should use it to create a customer, then use the customer to create a card?
* [ ] We create a stripe customer, but they are just assigned the current user and the current user's email, nothing else. Should probably assign token as `source`
* [ ] We always create a customer, need to make this conditional.
* [ ] We create a card after creating a customer, we assign it the user, the exp month, year, last 4 digits, brand, country, name, from stripe token response. Not sure if we should assign anything else.
* [ ] We add the card to the card list. User can now click to select it, then click donate. Should probably have some preselection.

* Most of the stuff is tested, but still need a good amount of acceptance testing.
* The process of adding a card is a long chain of promises. I tried to make it as simple as possible, but it's still complicated. I find that it's easier to debug if I add a console log to `controllers/project/donate/_handleError`. 

I hope this helps.

JoshSmith:
Adding these flow notes here for extra reference:

- Create a platform customer
- Create a customer on the connect account with the same info
- Create a plan on the connect account
- Create a card for the platform customer
- Create a token with the platform card and platform customer with the Stripe-Account header set to the connect account
- Create a subscription with the connect account's customer, the connect account's plan, and the token as the source

JoshSmith:
I've gone about as far as I can tonight. Still a number of failing tests due to changes.

I'm wondering how much we should try to close this PR out soon and put things into separate issues. It's not as though `develop` really works, so I'm not opposed to getting this passing, merging, and then breaking out into parallelizable issues.

app/components/donation/credit-card.js
```diff
@@ -2,11 +2,89 @@ import Ember from 'ember';
...
+  stripe: service(),
+
+  cardInvalid: not('cardValid'),
+  cardValid: and('isCardNumberValid', 'isCVCValid', 'isExpiryValid'),
+
+  preventSubmit: or('isSubmitting', 'cardInvalid'),
+
+  isSubmitting: false,
+
+  date: computed('month', 'year', function() {
```
sbatson5: Thoughts on simply renaming this to `dateString` so it is clear we expect it to be a string and not a date object?

app/components/donations/create-donation.js
```diff
@@ -1,9 +1,14 @@
 import Ember from 'ember';
 
-const { Component } = Ember;
+const { Component, computed, isEmpty } = Ember;
 
 export default Component.extend({
   classNames: ['create-donation'],
   presetAmounts: [10, 15, 25, 50],
-  amount: 15
+  amount: 10,
```
sbatson5: Could we add `customAmount: null` here as well, so it is clear it's a property set on this component?

app/controllers/project/donate.js
```diff
@@ -1,8 +1,107 @@
...
+  _tokenParams(cardParams) {
+    return {
+      number: cardParams.cardNumber,
+      cvc: cardParams.cvc,
+      exp_month: cardParams.month,
+      exp_year: cardParams.year
+    };
+  },
+
+  _updateAddingCardState() {
```
sbatson5: For clarity (not to be too nitpicky), could we do 'disableAddingCardState`?  `update` implies that it could be anything--whereas this function will only set it to false.

app/templates/components/donation/credit-card.hbs
```diff
@@ -7,37 +8,49 @@
...
+  {{#if isSubmitting}}Adding...{{else}}Add Card{{/if}}
+</button>
+
+{{#if canCancel}}
+  <a {{action "cancel"}}>Cancel</a>
+{{/if}}
+
+{{#unless isSubmitting}}
+  {{!validation errors can be passed in from parent, using a block}}
+  {{yield}}
```
sbatson5: 👍 I like this.

tests/acceptance/project-donate-test.js
```diff
@@ -0,0 +1,227 @@
...
+    amount: 10,
+    organization: organization.slug,
+    project: project.slug
+  });
+
+  andThen(() => {
+    projectDonatePage.creditCard.cardNumber('4242-4242-4242-4242');
+    projectDonatePage.creditCard.cardCVC('123');
+    projectDonatePage.creditCard.cardMonth('12');
+    projectDonatePage.creditCard.cardYear('2020');
```
sbatson5: Thoughts on extracting this block to a helper function since it is re-used?

tests/integration/components/donation/credit-card-test.js
```diff
@@ -1,29 +1,64 @@
...
+
+import creditCardComponent from '../../../pages/components/donation/credit-card';
+
+let page = PageObject.create(creditCardComponent);
+
+const {
+  K
+} = Ember;
+
+let setHandler = function(context, submitHandler = K) {
```
sbatson5: Thoughts on extracting this to a helper file since we use it so often?

tests/integration/components/donation/credit-card-test.js
```diff
@@ -1,29 +1,64 @@
...
+  };
 
-  let inputs = this.$().find('input').get();
+  setHandler(this, submitHandler);
 
-  inputs.forEach((input) => {
-    assert.notOk($(input).prop('disabled'));
+  stubService(this, 'stripe', {
+    card: {
+      validateCardNumber: () => true,
```
sbatson5: Since we're importing `Ember.K` already, we can change these to `validateCardNumber: K`, as it does the same thing.

sbatson5: Resolved in https://github.com/code-corps/code-corps-ember/pull/668/commits/37a6024037b25a58e0cd6c4d0abc92c1667f3cce as `Ember.K` is no longer needed.

tests/integration/components/donation/credit-card-test.js
```diff
@@ -50,6 +54,4 @@ test('it sends submit with credit card fields when button is clicked', function(
       .cardYear(expectedProps.year)
       .cardCVC(expectedProps.cvc)
       .clickSubmit();
-
-  assert.ok(page.submitDisabled, 'Submit button got disabled.');
```
sbatson5: The component is destroyed at this point, so the `_afterSubmit` method returns early and does not disable the button.

tests/integration/components/donation/donation-container-test.js
```diff
@@ -1,35 +1,73 @@
...
+  assert.equal(
+    page.paymentInformationText,
+    'Your payment method will be charged $100.00 per month to support CodeCorps.',
+    'Payment information renders correctly.'
+  );
+  assert.ok(page.creditCardIsRendered, 'Credit card component is rendered.');
 });
 
-test('it renders donation amount and frequency', function(assert) {
+test('it handles clicking card submit button correctly', function(assert) {
```
sbatson5: @JoshSmith I'm questioning what we are testing here.  Most of the actions in this particular test are for the credit card component.  This test fails because of the submit action on `donation/credit-card` (the child component).  I think the test coverage for submitting should be handled there and we should only worry about rendering what is appropriate here, as this is really just the container of the components that hold the logic.

JoshSmith: @sbatson5 agreed. Feel free to modify at your leisure.

app/components/donation/donation-container.js
```diff
@@ -2,13 +2,25 @@ import Ember from 'ember';
...
   Component,
-  computed: { bool }
+  computed,
+  computed: {
+    and, gt, not
+  }
 } = Ember;
 
 export default Component.extend({
+  // canAddCard: true,
```
sbatson5: Can we delete this comment?

PR #670, opened 2016-11-11
Allow account to turn on donations

begedin:
# What's in this PR?

Added the "activate-donations" action to the donation-goals page. Bound it to create a stripe plan for the specified project. 

## ToDo

Needs written.

## References

Progress on: #469


JoshSmith:
@begedin flow should work like this:

- Click Stripe Connect button
- Go to Stripe, fill out page
- Webhook creates `StripeAccount`, but is not yet associated with `Organization`
- Redirect to Code Corps
- Use OAuth token from redirect URL to connect `StripeAccount` to `Organization`, redirect to the project's donation goals
- Need Stripe to have approved account to create donation goals
- Need to create at least one donation goal to active Stripe
- Activating Stripe creates the default `StripePlan` for the `Project`, enabling donations, donation progress, etc on the project page
- You can't deactivate Stripe

JoshSmith:
@begedin based on notes above, this needs work to match that flow.

tests/helpers/relationship.js
```diff
@@ -3,25 +3,25 @@ import Ember from 'ember';
 
 const { get } = Ember;
 
+function runAssertions(assert, store, modelName, relationshipName, relationshipKind) {
```
sbatson5: Thoughts on renaming this function to better reflect that it is looking for relationships? `runAssertions` seems pretty broad.

begedin:
@JoshSmith Not sure if I'm missing anything here and correct me if I'm wrong, but I think parsing the results from the 'connect' process is going to be difficult.

The stripe connect process requires specifying the "redirect back" URL directly from the platform holder's stripe dashboard only. The most dynamic we can get is to specify a comma-separated list of URLs and then append a dynamic `redirect_to` parameter to our connect button URL, which is one of the urls from the list. 

I see several options on what we can do, but none of them are very nice. In all cases, we would redirect back to a specific route. Let's, for the purposes of this, just assume it's the `application/connect-stripe` route, which we don't have yet.

On clicking the stripe connect button, we store the current path as a cookie/local storage value. Once we hit the new route, that route loads the cookie/local storage value and navigates to that URL.

We premake the stripe account record for the organization and put it in a "not connected" state. We use that record to store our URL for redirecting.

We use the token created by our API for stripe-auth, to store a redirect back URL into it. We decode it locally, since stripe sends that back as well. We would have to share our Guardian secret key with our client app in that case. This feels the cleanest and least hacky, so for now, I did that while setting my test account to simply redirect to the application index route. Depending on how we set this up for everyone, it can be some other route, but it has to be one without dynamic segments.

JoshSmith:
This is blocked now by https://github.com/code-corps/code-corps-api/pull/460

JoshSmith:
Last comment by @begedin is out of date with changes made since.

JoshSmith:
Failing tests here are just linting and some missing model attributes in one unit test.

JoshSmith:
Last commit should fix everything we needed to make this work.

JoshSmith:
We still need to handle errors better when part of the process fails.

There are a few different errors:

- `400` when the code has already been used
- The organization already has a `StripeConnectAccount`
- Others I'm not thinking of?

JoshSmith:
There is an intermediate loading state, redirection flow appears to work pretty well, and we show the right things depending on what happens.

The `oauth-stripe.hbs` has the loading logic, but depending on what went wrong, we should just update that with some helpful information rather than redirect.

In the case of the code being used already, we should probably check if there is an account created and if not, render them a button to connect again?

If it already has a Stripe account, probably just redirect.

Other errors, maybe retry? Not sure.

Issue #688, opened 2016-11-17
Finalize donate page logic


begedin:
# Problem

The flow and framework are there, but it doesn't really work yet.

* Still need to not create customer when one already exists
* Still need to set proper properties for each record being created
* ...

JoshSmith:
@begedin can you update the to-do list here?

PR #689, opened 2016-11-17
Finalize donate page logic

begedin:
# What's in this PR?

This should be (hopefully), the final PR for the donation page to work properly. It will probably require merging of several API PRs

# Done until now

* Rework donate controller so it only tries to create customer when necessary 
* ...

# References 

Progress on #688


app/templates/components/donation/card-item.hbs
```diff
@@ -1 +1,5 @@
-{{card.brand}} ending in {{card.last4}}
\ No newline at end of file
+{{#if card.isLoading}}
+  Loading card...
+{{else}}
+  {{card.brand}} ending in {{card.last4}}
+{{/if}}
```
begedin: A quick way to resolve the case where the record is still loading, so we just get `<blank> ending in <blank>`

app/templates/components/donation/donation-container.hbs
```diff
@@ -11,19 +11,19 @@
   per month to support {{projectTitle}}.
 </p>
 
-{{#if canShowCardList}}
-  {{donation/card-list cards=cards selectedCard=defaultCard}}
-{{/if}}
 
-{{#if canAddCard}}
-  {{#donation/credit-card canSubmit=canSubmitAddCard submit=addCard canCancel=canShowCardList}}
+{{#if showCardForm}}
```
begedin: Form is now shown initially, if there are no cards defined. Past that point we toggle between the form and the list via the "add new card" button. We still need the cancel button.

PR #699, opened 2016-11-20
Remove coalescing from Stripe, update unit tests

### LINKED ISSUE #680: Remove coalescing from stripe- models
# Problem

All of our `stripe-` models should be written so that they do not coalesce records on fetch using `ember-data`. Records will need to be individually fetched.

## Subtasks
- [ ] Write test for the coalescing behavior
- [ ] Implement the right behavior
- [ ] Add endpoints to Mirage that mirror this behavior; ensure no `index` endpoints in Mirage

JoshSmith:
# What's in this PR?

Makes it so that Stripe-related models do not coalesce find requests.

Also updated tests to use `testForAttributes` helper and wrote proper tests for Stripe models.

Also updates Mirage config slightly.

## References
Fixes #680


app/adapters/stripe-subscription.js
```diff
@@ -0,0 +1,5 @@
+import ApplicationAdapter from './application';
+
+export default ApplicationAdapter.extend({
```
sbatson5: What are your thoughts on just creating a `StripeAdapter` that extends from the application adapter and sets `coalesceFindRequests` to false, and then all of these adapter could extend from that?

app/models/stripe-account.js
```diff
@@ -0,0 +1,16 @@
+import Model from 'ember-data/model';
+import attr from 'ember-data/attr';
+import { belongsTo } from 'ember-data/relationships';
+
+export default Model.extend({
+  businessName: attr(),
```
sbatson5: Is there a reason we don't specify these as `string` transforms but we do specify for `boolean` and `date`?

tests/unit/adapters/stripe-account-test.js
```diff
@@ -0,0 +1,12 @@
+import { moduleFor, test } from 'ember-qunit';
+
+moduleFor('adapter:stripe-account', 'Unit | Adapter | stripe account', {
+  // Specify the other units that are required for this test.
+  // needs: ['serializer:foo']
+});
+
+// Replace this with your real tests.
+test('it exists', function(assert) {
```
sbatson5: Thoughts on deleting these boilerplate tests?

tests/unit/models/stripe-account-test.js
```diff
@@ -0,0 +1,19 @@
...
+moduleForModel('stripe-account', 'Unit | Model | stripe account', {
+  // Specify the other units that are required for this test.
+  needs: [
+    'model:organization'
+  ]
+});
+
+test('it exists', function(assert) {
+  let model = this.subject();
+  // let store = this.store();
```
sbatson5: Can we delete this comment?

app/models/stripe-account.js
JoshSmith: @sbatson5 do they not need to be specified at all in either case?

app/adapters/stripe-subscription.js
JoshSmith: @sbatson5 that makes perfect sense. Can update.

tests/unit/adapters/stripe-account-test.js
JoshSmith: Ideally I would love to find a way to actually test that it does not coalesce find requests.

app/models/stripe-account.js
sbatson5: I don't think it matters.  `DS.StringTransform` will convert integers to strings.  The `boolean` transform will convert strings (other than `"true"` or `"t"`) to false.  But you really shouldn't need the transforms, since we control our API and how our information is serialized.

tests/unit/adapters/stripe-account-test.js
sbatson5: I don't know if that can happen at a unit test level as we have to test the request itself. Even then, I may argue against it since it seems like we are testing `Adapter` class.

PR #708, opened 2016-11-23
Rewrite integration tests for categories-list component to use page o…

### LINKED ISSUE #707: Rewrite categories-list integration tests with page objects
# Problem

Need to rewrite integration tests for the `categories-list` component to make use of page objects.

## References

Falls under umbrella issue #697

dmoon1221:
## References
Fixes #707

### DIFF
--- a/tests/integration/components/categories-list-test.js
+++ b/tests/integration/components/categories-list-test.js
@@ -2,15 +2,26 @@ import { moduleForComponent, test } from 'ember-qunit';
 import hbs from 'htmlbars-inline-precompile';
 import Ember from 'ember';
 import stubService from 'code-corps-ember/tests/helpers/stub-service';
+import PageObject from 'ember-cli-page-object';
+
+import categoriesListComponent from '../../pages/components/categories-list';
 
 const { K } = Ember;
 
+let page = PageObject.create(categoriesListComponent);
+
 moduleForComponent('categories-list', 'Integration | Component | categories list', {
-  integration: true
+  integration: true,
+  beforeEach() {
+    page.setContext(this);
+  },
+  afterEach() {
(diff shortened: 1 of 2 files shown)


tests/integration/components/categories-list-test.js
```diff
@@ -30,11 +41,11 @@ test('it renders the categories and sorts them by name', function(assert) {
...
   this.set('categories', categories);
-  this.render(hbs`{{categories-list categories=categories}}`);
+  page.render(hbs`{{categories-list categories=categories}}`);
 
-  assert.equal(this.$('.categories-list').length, 1);
-  assert.equal(this.$('.category-item').length, 3);
-  assert.equal(this.$('.category-item:eq(0)').text().trim(), 'Alphabets');
-  assert.equal(this.$('.category-item:eq(1)').text().trim(), 'Society');
-  assert.equal(this.$('.category-item:eq(2)').text().trim(), 'Zoology');
+  assert.ok(page.isRendered);
```
JoshSmith: I know this was in the tests before, but seems like we can remove if we're testing other things. What do you think @dmoon1221?

JoshSmith: I think the way the other selectors work they can't work if the elements are hidden, but I could be wrong. The main point of those tests was not really to check for visibility, but as a smoke test.

JoshSmith:
@dmoon1221 looks good but you need to remove `isVisible` now for the linter not to complain. When you do that, can you rebase from upstream and squash?

dmoon1221:
Ah, I'll be more careful about tests next time. Rebased, squashed, tests passing.

JoshSmith:
🙌 thanks @dmoon1221!

PR #710, opened 2016-11-23
Finalize process to turn on donations

### LINKED ISSUE #709: Finalize logic to "enable donations"
# Problem

We should finish the last parts of enabling donations for a project

## Subtasks
- [x] If the project has no `stripeConnectAccount` associated, stripe connect button is visible
- [x] Stripe connect button works and the end result is that it creates a `stripeConnectAccount`
- [ ] If there is a stripe connect account and the project has donation goals, but no associated `stripeConnectPlan`, the "turn on donations button" is visible 
- [ ] If there is no plan or donation goals, a message indicating donation goals are needed is show instead of the button
- [ ] clicking the button creates a plan for the project
- [ ] if there is a plan, a message indicating donations are active is visible

begedin:
# What's in this PR?

Final steps of the logic to turn on donations

- [x] If there is a stripe connect account and the project has donation goals, but no associated `stripeConnectPlan`, the "turn on donations button" is visible 
- [x] If there is no plan or donation goals, a message indicating donation goals are needed is show instead of the button
- [x] clicking the button creates a plan for the project
- [x] if there is a plan, a message indicating donations are active is visible

## References
Fixes #709


app/components/donation-goals.js
```diff
@@ -48,7 +48,7 @@ export default Component.extend({
    * @property canActivateDonations
    * @type {Boolean}
    */
-  canActivateDonations: computed.alias('hasExistingDonationGoals'),
+  canActivateDonations: computed.and('hasExistingDonationGoals'),
```
begedin: Missed it during a revert. Will fix.

app/templates/components/donation-goals.hbs
```diff
@@ -1,4 +1,6 @@
-{{#if canActivateDonations}}
+{{#if project.stripeConnectPlan}}
+  Donations are active for this project!
+{{else if canActivateDonations}}
```
JoshSmith: I would remove the `if` case. We won't show anything here.

JoshSmith: Would change to `if canActivateDonations` only.

PR #703, opened 2016-11-21
Styles for card-item from tickets #675 and #676

rlueder:
# What's in this PR?

- updated name of class added to cards on the list from "selected" to "card-item--selected" so it follows BEM conventions;
- created new stylesheet: components/donation/card-item.scss;
- adjusted font-weight of h3 elements to match Sketch design (semi-bold instead of bold);
- added font-size to html element (14px $body-font-size-normal) so that em units can use it as a reference;
- added title and button to donation-container.hbs (see reference to #676 below)

card-item

## References

Fixes #675 
- radio buttons still need styling to match Sketch mockups

Closes #676 
- "Add Card" is static at the moment;
- style for this button has to be moved to _buttons.scss (and all buttons should be updated to BEM naming convention)

Issue #672, opened 2016-11-11
Use BEM for Styles


rileytaylor:
# Problem

We want to standardize our stylesheets using [BEM](http://getbem.com/introduction/). This is a master issue to track all needed refactoring.

The lists below correspond to stylesheets. Each stylesheet (except those with `_` in front of the name) correspond to a `.hbs` template file. General styles will need to be renamed to fit the naming conventions, which then need to be changed everywhere they exist in templates. Take one `_name.scss` file at a time. Create a separate issue and reference this master issue. Take advantage of `git grep` to find all references to that style. If you are making changes to styles within components or templates, feel free to create a grouped issue and pull request (as in #673) Once each issue and it's related PR's are complete, check off the files refactored on this list. Make sure you make changes in both the `.scss` and `.hbs`! 

# Files to Refactor
## General
- [ ] _alerts.scss
- [ ] _animations.scss
- [ ] _buttons.scss
- [ ] _code.scss
- [ ] _colors.scss
- [ ] _fonts.scss
- [ ] _forms.scss
- [ ] _header.scss
- [ ] _icons.scss
- [ ] _images.scss
- [ ] _inputs.scss
- [ ] _misc.scss
- [ ] _pace.scss
- [ ] _retina.scss
- [ ] _settings.scss

## Templates
- [ ] app.scss
- [ ] dropdown-menu.scss
- [ ] index.scss
- [ ] main.scss
- [ ] start.scss
- [ ] task.scss
- [ ] task.scss
- [ ] task-new.scss
- [ ] templates/about.scss
- [ ] templates/team.scss
- [ ] templates/project/index.scss
- [ ] templates/project/settings/contributors.scss
- [ ] templates/start/expertiese.scss
- [ ] templates/start/hello.scss
- [ ] templates/start/skills.scss

## Components
- [ ] donations/donation-progress.scss
- [ ] donations/donation-status.scss
- [ ] categories-list.scss
- [ ] category-item.scss
- [ ] code-theme-selector.scss
- [ ] create-comment-form.scss
- [ ] donation-goal-edit.scss
- [ ] donation-goal.scss
- [ ] donation-payment.scss
- [ ] drag-zone.scss
- [ ] editor-with-preview.scss
- [ ] error-wrapper.scss
- [ ] image-drop.scss
- [ ] login-form.scss
- [ ] member-list-item.scss
- [ ] navigation-menu.scss
- [ ] organization-header.scss
- [ ] organization-members.scss
[...]

PR #786, opened 2016-11-27
Update thank you page

WenInCode:
# What's in this PR?
- Updated `alt` text for project's icon on thank you page
- Updated `src` of project's icon on thank you page
- Created `thankyou-container` component
- Created `thankyou-container` component page object
- Added minor tests for the component.
- I tried to update the styles to match the desired specification of the related issue ( #477 ), but I am not at all confident in that. Especially hoping for pointers here, maybe I'll ping it out in the design channel.

code_corps

## References
Progress on: #477


app/styles/templates/project/thankyou.scss
```diff
@@ -1,14 +1,27 @@
...
+    img {
+      display: inline-block;
+    }
+    margin-bottom: 0px;
   }
 
   .text {
-    margin-bottom: 18px;
+    color: $light-text;
+    font-size: 14px;
```
rlueder: on #703 I added "font-size: $body-font-size-normal" to the html element, that sets the baseline across the whole app to 14px so there's no need to set it here.

For future reference, we have four font sizes defined on _fonts.scss: 12px (small), 14px (normal), 16px (large) and 18px (x-large).

app/styles/templates/project/thankyou.scss
```diff
@@ -1,14 +1,27 @@
...
+  .header {
+    font-size: 1.5em;
+    margin-top: 0px;
+  }
+
+  .icon {
+    img {
+      display: inline-block;
+    }
+    margin-bottom: 0px;
```
rlueder: small tweak but 0 doesn't take units so "margin-bottom: 0" would be valid CSS

app/styles/templates/project/thankyou.scss
```diff
@@ -1,14 +1,27 @@
...
 
-  .icon img {
-    display: inline-block;
+  .header, .text {
+    margin-bottom: 20px;
+  }
+
+  .header {
+    font-size: 1.5em;
+    margin-top: 0px;
```
rlueder: small tweak but 0 doesn't take units so "margin-bottom: 0" would be valid CSS

app/styles/templates/project/thankyou.scss
```diff
@@ -1,14 +1,27 @@
...
   }
 
-  .icon img {
-    display: inline-block;
+  .header, .text {
+    margin-bottom: 20px;
+  }
+
+  .header {
+    font-size: 1.5em;
```
rlueder: I like that you're using ems and I think that's probably the way going forward when the app becomes responsive.

With our baseline at 14px 1.5em is rendered as 21px, the Sketch reference has headers set at 24px (1.7em approx.).

Minor conversation at this point but hey, every pixel counts. 😛

app/styles/templates/project/thankyou.scss
```diff
@@ -1,14 +1,27 @@
...
-  .icon, .header, .text, .project-link{
+  .icon, .header, .text, .project-link {
     width: 100%;
     text-align: center;
   }
 
-  .icon img {
-    display: inline-block;
+  .header, .text {
+    margin-bottom: 20px;
```
rlueder: I'd set margins as ems as well, 14px is the baseline multiplier so 20px would translate to "margin-bottom: 1.4em"

app/templates/components/thankyou-container.hbs
```diff
@@ -0,0 +1,4 @@
+<div class="icon"><img src={{project.iconLargeUrl}} alt={{projectIconAltText}} data-test-selector="project icon" /></div>
+<h3 class="header">Thank you!</h3>
+<div class="text" data-test-selector="thank you message">From all the volunteers on the {{project.title}} team.</div>
+<div class="project-link">{{#link-to "project" project}}Back to project{{/link-to}}</div>
```
rlueder: This comment applies to all html elements above (and it's really just a suggestion). To follow the BEM approach describe on #672 I'd make the following changes:

`<div class="thankyou-container">`
`<div class="thankyou-message">`
`<div class="project-icon"></div>` (more descriptive than just "icon" and could be used on other templates. If a class "icon" already exists you can extend project-icon with: @extend icon)
`<h3>Thank you!</h3>`
`<p>From all the volunteers on the {{project.title}} team.</p>`
`{{#link-to "project" project class="project-link"}}Back to project{{/link-to}}` (you can add a class directly to the view helper)
`</div>`
`</div>`

This would of course affect the css above and also the tests, I think it'd make it easier to style the component, most of the styling would be inherited.

JoshSmith:
I've got this.

Issue #783, opened 2016-11-25
Change donations to take a single card instead of a list


JoshSmith:
# Problem

First, we need to take the design changes from #703 and narrow them down to styling of the individual card only.

If the user has a card:
- [x] show the card item
- [x] hit donate
- _NOT MVP - TODO LATER_ you can edit or replace the card
> Visa 4242 [edit] [replace]
[edit] lets you update expiration dates
[replace] gives you card form with cancel link or donate button

If the user doesn't have a card:
- [x] show the new card form
- [ ] hit donate
- [ ] creates customer (if needed)
- [ ] creates card
- [ ] creates subscription

rlueder:
I'll work on this one.

rlueder:
@JoshSmith: most of what's needed seems to be in place already with only a few small tweaks needed, here's my understanding of the task:

1. rename "Add Card" button to "Donate";

2. on donation page change header from "Payment information" to "Review Payment", I'll also review the styles since there's no concept of selecting a card anymore;

3. I'll also consider the Edit and Replace buttons, it won't take me much longer for me to style the static buttons. We can hide the elements and add links to them on a separate ticket;

### Future "Nice-to-Have"

- Sign Up should be the default option for an anonymous user (with Sign In as an option on the page itself and not just main nav "already registered? Sign In to donate");

- if a user had previously decided to donate by clicking on "Become a Donor" and then choosing a specific dollar amount it'd be nice to take them directly to the "Payment" page (formerly "Add a Card") with that amount already selected ($xx given each month) so they don't have to repeat the steps after signing up;

- allow users to donate without signing up, ask for an email with payment information and send them into selecting interests, expertise and skills only after they've already donated (direct Sign Up flow is not affected)

JoshSmith:
Edit and replace have difficult actions associated with them, so if we can leave them out for now and circle back later, I think that's better.

JoshSmith:
I'm not sure that there should be a duplicated "Donate" button.

I think there's more here than what you listed above, for example we need to remove the cancel link. `canCancel` similar is not a thing.

When the card gets added, we also probably don't want it to be displaying the card, which I imagine it might do when that promise resolves.

JoshSmith:
Closed by #793.

PR #793, opened 2016-11-28
minor changes to copy and removing cancel action

rlueder:
# What's in this PR?

- minor text changes to credit-card.hbs to reflect change in flow
- removed reference to cancel action on credit-card.js

## References

Progress on: #783 

## Todo

- it seems like pull request #780 has impacted rendering of card.brand, card.expMonth, card.expYear and card.last4 on the payment review step (prev. card list) I did some investigation but didn't get to a solution.

### DIFF
--- a/app/components/donation/credit-card.js
+++ b/app/components/donation/credit-card.js
@@ -23,9 +23,7 @@ export default Component.extend({
   cardInvalid: not('cardValid'),
   cardValid: and('isCardNumberValid', 'isCVCValid', 'isExpiryValid'),
 
-  preventSubmit: or('isSubmitting', 'cardInvalid'),
-
-  isSubmitting: false,
+  isBusyOrInvalid: or('isBusy', 'cardInvalid'),
 
   date: computed('month', 'year', function() {
     let month = this.get('month');
@@ -81,27 +79,13 @@ export default Component.extend({
     this.setFreshDate();
   },
 
-  _afterSubmit() {
-    if (this.get('isDestroyed')) {
-      return;
-    }
-    this.set('isSubmitting', false);
-  },
-
   actions: {
     submit() {
       this.set('isSubmitting', true);
       let cardAttrs = this.getProperties('cvc', 'cardNumber', 'year', 'month');
       let onSubmit = this.get('submit');
 
-      onSubmit(cardAttrs).finally(() => this._afterSubmit());
-    },
-
-    cancel() {
-      let cancelAddCard = this.get('cancelAddCard');
-
-      this.clearForm();
-      cancelAddCard();
+      onSubmit(cardAttrs);
     }
   }
 });
--- a/app/components/donation/donation-container.js
+++ b/app/components/donation/donation-container.js
@@ -1,33 +1,11 @@
 import Ember from 'ember';
 
 const {
-  Component,
-  computed,
-  computed: {
-    and, empty, not, or
-  }
+  Component
 } = Ember;
 
 export default Component.extend({
   classNames: ['donation-container'],
   donationAmount: 0,
-  projectTitle: null,
-
-  canAddCard: computed('hasCard', 'isAddingCard', function() {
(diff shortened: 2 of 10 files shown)


tests/integration/components/donation/credit-card-test.js
```diff
@@ -19,22 +19,21 @@ let setHandler = function(context, handlerName, handler = K) {
...
+// let getFreshDates = function() {
+//   let date = new Date();
+//   let currentMonth = `0${date.getMonth() + 1}`.slice(-2);
+//   let currentYear = date.getFullYear();
+//
+//   return {
+//     month: currentMonth,
+//     year: currentYear
+//   };
+// };
```
begedin: Looks like this is unused now that "cancel" is gone. Feel free to remove the code as well.

begedin:
Just to review requirements:

* [ ] Show card form if there is no card - donate button on card form triggers the entire process
* [ ] Show card item and donate button if there is a card (donate button only creates subscription using existing card)
* [ ] Prevent flashing of card form while card is still loading.

begedin:
@rlueder Hope you don't mind, made a bunch of these changes myself. @JoshSmith could use a review from both of you guys.

app/components/donation/credit-card.js
```diff
@@ -23,9 +23,7 @@ export default Component.extend({
   cardInvalid: not('cardValid'),
   cardValid: and('isCardNumberValid', 'isCVCValid', 'isExpiryValid'),
 
-  preventSubmit: or('isSubmitting', 'cardInvalid'),
-
-  isSubmitting: false,
+  isBusyOrInvalid: or('isBusy', 'cardInvalid'),
```
begedin: Route simply has an `isLoading` flag now, when anything is happening, which is passed down into the components. Much simpler that way, with the exception being the card form, which can also be disabled if the card is invalid. Still, simpler than the previous approach.

app/components/donation/credit-card.js
```diff
@@ -81,27 +79,13 @@ export default Component.extend({
     this.setFreshDate();
   },
 
-  _afterSubmit() {
-    if (this.get('isDestroyed')) {
-      return;
-    }
-    this.set('isSubmitting', false);
-  },
```
begedin: No need to do this anymore, since the busy state is set from outside.

app/components/donation/credit-card.js
```diff
@@ -81,27 +79,13 @@ export default Component.extend({
...
-
   actions: {
     submit() {
       this.set('isSubmitting', true);
       let cardAttrs = this.getProperties('cvc', 'cardNumber', 'year', 'month');
       let onSubmit = this.get('submit');
 
-      onSubmit(cardAttrs).finally(() => this._afterSubmit());
```
begedin: No need to wait for a promise anymore, either.

app/controllers/project/donate.js
```diff
@@ -25,26 +25,26 @@ export default Controller.extend({
...
       this._clearErrors();
+      this._updateIsLoading(true);
 
       return this._createCreditCardToken(cardParams)
                  .then((stripeResponse) => this._createCardForPlatformCustomer(stripeResponse))
+                 .then((stripeCard) => this._createSubscription(amount, stripeCard))
+                 .then(() => this._transitionToThankYou())
                  .catch((reason) => this._handleError(reason))
-                 .finally(() => this._updateAddingCardState(false));
+                 .finally(() => this._updateIsLoading(false));
```
begedin: I could extract the first part into a function to avoid repeating, but really, that would just make it more complicated. This is more explicit, IMO.

app/components/donation/donation-container.js
```diff
@@ -1,33 +1,11 @@
...
-  isNotAddingCard: not('isAddingCard'),
-
-  showCardForm: or('hasNoCard', 'isAddingCard'),
-
-  actions: {
-    addCard() {
-      this.set('isAddingCard', true);
-    }
-  }
+  projectTitle: null
```
begedin: Especially love the changeset here.

PR #794, opened 2016-11-28
#784 - Redirect to Project Page if User Already Has a Subscription

### LINKED ISSUE #784: Don't let users who have a subscription create a new one
# Problem

If you go to `http://localhost:4200/code-corps/code-corps/donate?amount=25` manually when you already have a subscription, you would be able to create a new one.

## Subtasks
- [ ] Figure out a better alternative for this (updating it? redirecting?)

alexwilkinson:
This is still a work in progress, but the idea is to put in a check in the `afterModel` hook of the `donate` route for whether the current user has a stripe subscription for a project already. If so, redirect to the project page.

So far, the `user.stripeConnectSubscriptions` collection seems to be available, but I am having trouble traversing the object graph when inspecting an individual subscription in the find loop. Maybe mirage isn't setting up the obejcts properly?

Also updated errors in the mirage models. 

Will Fix #784

### DIFF
--- a/app/routes/project/donate.js
+++ b/app/routes/project/donate.js
@@ -2,7 +2,20 @@ import AuthenticatedRouteMixin from 'ember-simple-auth/mixins/authenticated-rout
 import Ember from 'ember';
 
 const {
+  computed: { alias },
+  inject: { service },
   Route
 } = Ember;
 
-export default Route.extend(AuthenticatedRouteMixin, {});
+export default Route.extend(AuthenticatedRouteMixin, {
+  currentUser: service(),
+  user: alias('currentUser.user'),
+  project: alias('model'),
+
+  afterModel() {
+    let subscriptions = this.get('user.stripeConnectSubscriptions');
+    let subscribedToProject = subscriptions.find(function(subscription) {
+      return subscription.get('stripeConnectPlan.project') == this.get('project');
+    });
+  }
+});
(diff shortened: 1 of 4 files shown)


JoshSmith:
@alexwilkinson maybe we want to put all this into a service object and take an approach similar to some of the other services there (like the user skills one).

alexwilkinson:
@JoshSmith Definitely fond of that idea—I was thinking of doing that from the start, once the logic had been figured out.

begedin:
@alexwilkinson I'm sorry not to have realised this sooner, but I was actually working on the same solution. We seem to have duplicate issues, #784 and #798, one of which was assigned to you and another to myself. 

Mine is just about ready to merge and contains other fixes, but it only deals with redirection from the `project.donate` route.

Once it's merged, I was thinking we should revisit your issue (since based on title alone, it's wider in scope) and figure out if anything else is needed in relation to a user already having a subscription. 

I apologize for the inconvenience. It was an honest mistake on my part, since you clearly started on your issue sooner. We'll work on avoiding these sort of misunderstandings in the future.

alexwilkinson:
@begedin That's ok!! You got much further along anyways, and if my little work on this Issue just ended up helping me get acclimated to the project, that's useful too : ). I'll just note on your PR another mirage model that you can fix while you're at it—something @JoshSmith discovered while I we were discussing my mirage struggles.

The only thing that could be revisited around this Issue would be potentially showing a more useful page to the user in the situation where they already have a subscription, instead of just redirecting. That conversation is in #784. I'm going to make a new Issue about that.

JoshSmith:
@alexwilkinson not sure if you made a new issue, but please do when you get a chance!

PR #803, opened 2016-11-29
Redirect user to projects.index from projects.donate, if donation already exist

### LINKED ISSUE #798: You should not be able to load /donate?amount if the project doesn't have donations active
# Problem

If you go to load `/donate?amount` and the project has no donations active, then the page still works. We should fix this by redirecting to the project page in this event.

begedin:
# What's in this PR?

Changes the flow on the donate page to redirect to main project route if donation already exists. 

Since it reuses the same logic as the `project.index` route to find the user's subscription for the current project, I moved that logic into a service.

I also added a flash message, but it doesn't seem to work for some reason. Flash messages do not seem to be broken, but I couldn't find a problem. Any pointers are welcome.

## References
Fixes #798

### DIFF
--- a/app/routes/project/donate.js
+++ b/app/routes/project/donate.js
@@ -2,7 +2,37 @@ import AuthenticatedRouteMixin from 'ember-simple-auth/mixins/authenticated-rout
 import Ember from 'ember';
 
 const {
-  Route
+  get,
+  inject: { service },
+  Route,
+  RSVP
 } = Ember;
 
-export default Route.extend(AuthenticatedRouteMixin, {});
+const ALREADY_A_SUBSCRIBER = "You're already supporting this project.";
+
+export default Route.extend(AuthenticatedRouteMixin, {
+  flashMessages: service(),
+  userSubscriptions: service(),
+
+  model() {
+    let project = this.modelFor('project');
+
+    return RSVP.hash({
+      project,
+      subscription: this.get('userSubscriptions').fetchForProject(project)
+    });
+  },
+
+  afterModel({ project, subscription }) {
+    if (subscription) {
+      get(this, 'flashMessages').success(ALREADY_A_SUBSCRIBER);
+      this.transitionTo('project', project);
+    } else {
+      this._super.call(...arguments);
+    }
+  },
+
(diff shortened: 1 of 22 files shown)


app/routes/project/donate.js
```diff
@@ -2,7 +2,35 @@ import AuthenticatedRouteMixin from 'ember-simple-auth/mixins/authenticated-rout
...
+
+    return RSVP.hash({
+      project,
+      subscription: this.get('userSubscriptions').fetchForProject(project)
+    });
+  },
+
+  afterModel({ project, subscription }) {
+    if (subscription) {
+      get(this, 'flashMessages').success('We are grateful for eagerness, but you are supporting us already.');
```
JoshSmith: We should buff this language:

> You're already supporting this project.

JoshSmith:
@begedin I think flash messages aren't showing maybe because we clear them at the router transition level. I'm not sure if we are doing that, but it seems non-obvious means that it was a fix for something that went undocumented and is now not working as we would expect. Too hacky.

begedin:
@JoshSmith We actually do clear messages on transition, which means two things.

1. They will clear when we switch routes.
2. In hooks such as `afterModel`, even if we stay on the same route, `didTransition` is called after, so it will clear any message added through such a hook.

My guess is, we are clearing messages because of most of the flashes we actually display are "sticky", meaning they never go away until the user dismisses them via  click. Since all of these cases where there is a sticky also specify a timeout, my guess would be, we were adding them incorrectly to begin with.

My suggestion is, we 
* remove the sticky from all of them
* remove the explicit timeout and use a default timeout
* use helpers (`.success(message)`, `.warning(message)`, etc.) instead of the default `add(optionsHash)` to keep things simpler.
* remove the clearing on `didTransition`

begedin:
@JoshSmith I rewrote the flash messages and added some convenience test helpers for them, so we don't have to stub the service anymore.

From now on, every time we display a message, we explicitly clear the message queue first. It didn't look like we need to not clear it in any place, but it's easily switchable now.

JoshSmith:
@alexwilkinson addressed that change.

PR #804, opened 2016-11-29
Replace credit card input with ember-credit-cards

### LINKED ISSUE #799: Add rendering of split credit card numbers in input when typing
# Problem

We should add an ember-fied version of https://github.com/stripe/jquery.payment#fnpaymentformatcardnumber (minus the jQuery). Can use this for inspiration.

https://github.com/stripe/jquery.payment/blob/master/lib/jquery.payment.js#L224

## Subtasks
- [ ] Includes a space between every 4 digits
- [ ] Restricts input to numbers
- [ ] Limits to 16 numbers
- [ ] Supports American Express formatting

begedin:
# What's in this PR?

Replaces our credit card number and cvc input with components provided by `ember-credit-cards`.

The addon also supports an expiration date input, but this one is a format restricted `input` field, instead of a pair of `select` elements, so I did not make that replacement.

## References
Fixes #799

### DIFF
--- a/app/templates/components/donation/credit-card.hbs
+++ b/app/templates/components/donation/credit-card.hbs
@@ -12,11 +12,12 @@
 <div class="input-group">
   <label>
     <span>Card number</span>
-    {{input name="card-number" type="text" value=cardNumber}}
+    {{input-credit-card-number name="card-number" number=cardNumber}}
   </label>
 </div>
 
 <div class="input-group">
+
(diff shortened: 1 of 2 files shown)


begedin:
I know the suggestion was to wrap `jquery.payments`, but this addon seems pretty well-rounded and is inspired by `jquery.payments`. It feels like less work and easier to manage.

PR #809, opened 2016-11-30
Update thank you image with high five

JoshSmith:
# What's in this PR?

Updates the thank you page to use a high five instead of the fireworks.

This could probably be simplified some. Would love feedback on how to simplify further.


app/components/animated-high-five.js
```diff
@@ -0,0 +1,70 @@
...
+    let followOnTimer = run.later((function() {
+      set(that, 'followOnAnimation', false);
+    }), 500);
+
+    set(this, 'followOnTimer', followOnTimer);
+    set(this, 'reset', true);
+
+    let resetTimer = run.later((function() {
+      set(that, 'reset', false);
+    }), 1);
```
begedin: Could probably use an arrow function instead of storing context into `that`. That should preserve the `this` from within the arrow function.

begedin: Also, I believe `run.later` can receive the context as an argument to, so that might be an option.

PR #810, opened 2016-11-30
Improve donation progress styles

JoshSmith:
# What's in this PR?

Just minor changes to CSS.

@begedin if this breaks tests would you mind fixing? This was my last one before bed.


JoshSmith:
@begedin they were broken but I fixed. :ton

PR #814, opened 2016-11-30
Render donation goal related errors properly

### LINKED ISSUE #813: Render form errors when donation goal creation fails
# Problem

We should render form errors when donation goal creation fails.

Right now if you get any kind of 422 error (or any error), you don't know why.

We may want to add in "Saving..." to the button here while saving, too (which makes cancel impossible to do while it's saving).

begedin:
# What's in this PR?

Adds field-based validation error rendering for the donation goals page.

Also adds displaying of other generic server-side errors as a friendly error message.

I also realised our `isValidationError` helper on the `project.donate` page was wrong in assuming our server payload would contain a title set to `VALIDATION_ERROR`. Instead, it is now checking if any of the errors in the payload contains a `source` property. 

The helper has also been moved into a separate module, since it's now also being used on the `project.donations` page.

## References
Fixes #813

### DIFF
--- a/app/controllers/project/donate.js
+++ b/app/controllers/project/donate.js
@@ -1,5 +1,6 @@
 import Ember from 'ember';
 import FriendlyError from 'code-corps-ember/utils/friendly-error';
+import { isValidationError } from 'code-corps-ember/utils/error-utils';
 
 const {
   Controller,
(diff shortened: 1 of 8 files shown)


JoshSmith:
🙌  LGTM

PR #818, opened 2016-12-01
Prevent UI popping during subscription process

### LINKED ISSUE #811: Fix rendering of pieces of the credit card page as things slowly load in
# Problem

@begedin when I talked about "flashing" of the components here, I meant that if something loads in, then we do not want to show the components loading piecemeal.

Right now, though, if you create your customer/card/etc and that loads slowly, you'll watch as the form changes to load in the card, etc.

It's pretty unexpected and we would prefer to prevent re-rendering the UI until the entire operation completed, not until each atomic promise returns.

Understandably, we may have to render the card or something after an error.

Maybe we need to rethink how this looks in order to make things clearer, but as it stands, things are simply kind of confusing from the user's perspective.

begedin:
# What's in this PR?

Closes #811

This makes it so while anything is happening in the background on the donate page, both the card form and the card item display are replaced by a "Processing..." indicator, which includes a loading spinner.

Since the change eliminates the need for some logic in the credit-card component (namely, the busy state), commit 260044f also removes all of that code.

If this behavior is not an acceptable solution, the only alternative I can think of is to show the form **until** the whole process is done, provided the form was there at the start, or show the card item until the process is done, provided the card item was there to begin with.

The problem with that is, it would significantly complicate the logic. In order to achieve that, we should

* revert 260044f
* remove the `isBusy` wrapping from the template
* on `templates/project/donate`, replace `{{#donation/donation-container
  card=user.stripePlatformCard`with just `card=card`
* in `routes/project/donate`,  add a `model` hook that explicitly loads the project and the user's card. Then add a `setupController` hook that sets the two properties on the controller. That way, the card wont be set automatically and the UI wont switch from the form to the card half-way through.

If the current solution is acceptable, finalizing this PR is a matter of setting the right style (in case the current spinner with "Processing..." is not good).

If the proposed alternative is the way to go, then the process id described above. I don't think it would break any tests to do it that way, but as explained, it slightly complicates things.

### DIFF
--- a/app/components/donation/credit-card.js
+++ b/app/components/donation/credit-card.js
@@ -81,7 +81,6 @@ export default Component.extend({
 
   actions: {
     submit() {
-      this.set('isSubmitting', true);
       let cardAttrs = this.getProperties('cvc', 'cardNumber', 'year', 'month');
       let onSubmit = this.get('submit');
 
--- a/app/components/donation/credit-card.js
+++ b/app/components/donation/credit-card.js
@@ -3,7 +3,7 @@ import Ember from 'ember';
 const {
   Component,
   computed,
-  computed: { and, not, or },
+  computed: { and, not },
   inject: { service }
 } = Ember;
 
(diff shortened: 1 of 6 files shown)


app/components/donation/credit-card.js
```diff
@@ -81,7 +79,6 @@ export default Component.extend({
 
   actions: {
     submit() {
-      this.set('isSubmitting', true);
```
begedin: This was actually useless property we don't need anymore, so I would keep this one change in no matter what.

JoshSmith:
We can fix this by simply doing:

```
{{#if (and card (not isBusy))}}
```

Will need to install this https://github.com/jmurphyau/ember-truth-helpers

PR #821, opened 2016-12-01
Fix card loading (WIP)

JoshSmith:
# What's in this PR?

WIP

### DIFF
--- a/app/components/donation/donation-container.js
+++ b/app/components/donation/donation-container.js
@@ -1,11 +1,31 @@
 import Ember from 'ember';
 
 const {
-  Component
+  Component,
+  computed: {
+    and, empty, or
+  }
 } = Ember;
 
 export default Component.extend({
   classNames: ['donation-container'],
   donationAmount: 0,
-  projectTitle: null
+  projectTitle: null,
+  wasNewCard: true,
+
+  /**
+   * If there is no card for the user, this might be an ObjectProxy
+   * that has `null` content. This `card.id` approach was the least
+   * hacky way we could think to deal with this late one night.
+   *
+   * TODO: Find a better approach!
+   */
+  isNewCard: empty('card.id'),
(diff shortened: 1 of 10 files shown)


JoshSmith:
Various states:

new card processing - show the form and “Processing…” button (action on form)
existing card processing - show the card and “Processing…” button (action on container)
no card processing - show the form and “Processing...” button (action on form)
new card not processing - show the card and “Donate” button (action on container)
existing card not processing - show the card and “Donate" button (action on container)
no card not processing - show the form and “Donate” button (action on form)

noslouch:
which is the desired button to show? the button on `donation/donation-container` or `donation/credit-card`?

app/components/donation/donation-container.js
```diff
@@ -1,11 +1,24 @@
...
 } = Ember;
 
 export default Component.extend({
   classNames: ['donation-container'],
   donationAmount: 0,
-  projectTitle: null
+  projectTitle: null,
+  wasNewCard: true,
+
+  isNewCard: empty('card.id'),
```
begedin: Someone might wonder whiz this isn't simply `alias(cad.isNew)`, not realizing it's because `card` might be `null` to begin with. Maybe a comment would help?

At least, that's my assumption here. Maybe `card.isNew` would actually work? Does init get called if the component is not rendered?

JoshSmith: @begedin this one is tricky because it's an `ObjectProxy` which has `content: null` if there is none. This was the least hacky way I could think to deal with this.

Should definitely comment, but would prefer a better fix.

I'm not super satisfied with my implementation over all, but hey: it works.

begedin: Maybe add a request for a better fix in the comment. Maybe someone gets inspired.

JoshSmith: @begedin can you make an issue refing this PR for us to fix? And put under this milestone, I guess.

PR #825, opened 2016-12-03
Add ember serve line to INSTALLING.md

gdcohan:
# What's in this PR?

Added a line to the installation documentation about how to run the ember server to verify that everything is set up correctly.


JoshSmith:
🙌 @gdcohan 

@WenInCode @rileytaylor think we should improve the docs to point more from INSTALLING => USAGE when done here? Can cut some duplication, I think with no less to flow of docs.

PR #830, opened 2016-12-03
adds civic.json

noslouch:
# What's in this PR?
Adds a civic.json file

## References
[civic.json](http://open.dc.gov/civic.json/) is an open spec for civc tech projects. The idea looks to be provide a way for civic tech aggregators to spread the word about projects of these kinds.

### DIFF
--- a/civic.json
+++ b/civic.json
@@ -1,0 +1,24 @@
+{
+  "name": "Code Corps",
+  "description": "Help build and fund software projects with massive social impact.",
+  "status": "Beta",
+  "homepage": "https://www.codecorps.org",
+  "license": "https://github.com/code-corps/code-corps-ember/blob/develop/LICENSE.txt",
+  "repository": "https://github.com/code-corps/code-corps-ember",
+  "geography": [],
+  "contact": {
+    "email": "josh@codecorps.org",
+    "name": "Josh Smith"
+  },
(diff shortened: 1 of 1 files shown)


civic.json
```diff
@@ -0,0 +1,23 @@
...
+  "politicalEntity":"",
+  "type":"Web App",
+  "needs": [
+    {"need": "Ember developer"},
+    {"need": "Phoenix developer"},
+    {"need": "Open source evangelist"}
+  ],
+  "categories": [
+    {"category": "Open Source"},
+    {"category": "Developer productivity"}
```
JoshSmith: @noslouch maybe change to "Developer Tools"?

civic.json
```diff
@@ -0,0 +1,23 @@
...
+  ],
+  "tags": [
+    "open source",
+    "project management",
+    "tech for good"
```
JoshSmith: Are these tags and the categories above already in use on their site?

noslouch: no, they're placeholder

noslouch: if you had any, that would be great. I'm trying to carve out some time for CC between work and life, but I can add a few thoughts if you get the ball rolling.

JoshSmith: @noslouch not off the top of my head. Can circle back, but those look good honestly.

PR #834, opened 2016-12-05
Fix issues with organization membership reloading

### LINKED ISSUE #824: Organization memberships don't load correctly on sign in
# Problem

A user's organization memberships do not appear to refresh on sign in, and load in inconsistently at other times.

You can see this when:

- Signing in and trying to join a project; the button will not change state because the memberships have not updated or computed
- Signing in as an organization admin and not seeing the right tabs on a project

## Subtasks
- [ ] Determine what is causing this issue
- [ ] Write a regression test that catches up
- [ ] Fix the issue

JoshSmith:
# What's in this PR?

Fixes #824.

The issue is directly related to https://github.com/minutebase/ember-can/issues/20, which I missed somehow when trying to see if there were related issues.

The complexity of this service grew to rely on being promise-aware due to `beforeModel()` hooks in routes that prevent transitions. Those worked fine, but the computed properties throughout the application's templates did not.

This also works with Mirage now. 👏 

There are a few things I'm not satisfied with in the current implementation, that can maybe go into a separate issue to refactor later now that this is working, in order from smallest to largest refactorings:

- We could maybe stand to rewrite the promise logic and see if we can collapse the API from `membership` and `membershipPromise` into a single API. I'm not sure this is possible given the `ember-can` issue above.
- This service is poorly named. `credentials` does not really make clear what we're doing here, which is checking for organization permissions. Maybe even just `organization-permissions` would be a better name.
- We're acting like this is something akin to the `onboarding` service, which is application-wide, a design decision driven by being on a project route or organization route. On the route? Set the current organization in the service. This doesn't reflect other use cases, though, like checking for organization permissions in a list of organizations. Suggest that we consider thinking more functionally here.
- I'm not entirely positive that the Ember application should even be responsible for any of the complexity here. I would like to consider hiding most of the complex permission logic inside of the user via the JWT. Updates to the user's data would then reflect in organization permissions.

### DIFF
--- a/app/services/credentials.js
+++ b/app/services/credentials.js
@@ -20,7 +20,7 @@ export default Service.extend({
 
   membership: computed('memberships', 'user', {
     get() {
-      get(this, 'membershipPromise').then((membership) => {
+      this.fetchMembership().then((membership) => {
         set(this, 'membership', membership);
       });
       return null;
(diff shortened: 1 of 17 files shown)


app/services/credentials.js
```diff
@@ -13,64 +18,64 @@ export default Service.extend({
...
     }
   }),
 
-  membershipPromise: computed('memberships', function() {
-    let memberships = this.get('memberships');
-    let fulfilled = this.get('memberships.isFulfilled');
-
+  // Memberships may not be loaded, so we need to wait for them to resolve
+  // Can be useful in routing when we need it loaded to prevent a transition
+  membershipPromise: computed('memberships', 'user', function() {
```
green-arrow: It feels like this should just be a function like `fetchMembership` instead of a computer property. Since the `membership` property is already computed based on the same values, it should still be fine

JoshSmith:
@begedin followed up here with #836.

Issue #698, opened 2016-11-20
Rewrite page objects to be decoupled from CSS


JoshSmith:
# Problem

https://dockyard.com/blog/2015/09/25/ember-best-practices-acceptance-tests

> The first approach we’re going to talk about is decoupling your acceptance tests from the HTML structure and CSS of your application. If your tests are breaking due to semantic HTML or class name changes but the end experience to the user is nearly exactly the same, this should be a red flag that your tests are too closely coupled to the HTML and CSS of your application. Our acceptance tests should be from the perspective of a user interacting with the application. A user doesn’t care about whether the element they are clicking on is a `<button>` or a `<span>` or whether we use a class name of `post` or `blog-post`. They care that they can read a blog post, comment on it, and like it.

> However, we can’t perfectly decouple our tests from presentation, as we still need to be able to find the elements that we need to interact with and make assertions against. One approach to minimizing the coupling is to add a data attribute to elements that we want to target during our tests. Data attributes are attributes that are prefixed with data- and are intended to store meta data on elements

## Subtasks
- [ ] Create individual issues for page objects / components / etc that need decoupled
- [ ] Submit PRs for those page objects / components / etc
- [ ] ???
- [ ] Profit!

## References

Related to: #697

WenInCode:
I like this idea. At my work we use aria-labels a lot and select from them as they are pretty descriptive and don't change often.

JoshSmith:
@WenInCode yeah I'm not as big a fan of `aria` since that's for accessibility, but the principle is sound.

PR #846, opened 2016-12-08
672 use BEM for styles

rlueder:
# What's in this PR?

This PR is taking care of most of the refactoring needs described in #672 

The approach I'm taking is (somewhat) documented here: [Code Corps SASS Ideas and Discussion](https://github.com/code-corps/code-corps-ember/wiki/Code-Corps-SASS---Ideas-and-Dicussion)

## References

Progress on: #672


JoshSmith:
Wow, this is a lot of work. Okay, I need time to sit down and go through this. Looks like there are conflicts due to my changes now too @rlueder, and some failing tests.

Would you be able to fix conflicts/tests and then I can make sure I sit down to review this and understand it fully.

WenInCode:
Hey @rlueder - I saw that you mentioned you were a bit confused about decoupling the tests from css. It so happens that your tests are failing for that very reason. In this PR, you've updated some classes to follow the BEM style guide. Some of the page objects (used in tests to represent/select from pages and/or components) used those css classes as selectors. Because those css classes no longer exist, they are not returning the expected values, and in result the tests are failing. 

Going forward we want to use `data-test-selector` attributes on elements that need to be accessed in tests. This way we can make descriptive selectors that won't change during styling changes. 

A good example of this [failing test](https://github.com/code-corps/code-corps-ember/blob/672_use-BEM-for-styles/tests/integration/components/categories-list-test.js#L23). It is failing because [this class name was replaced](https://github.com/code-corps/code-corps-ember/pull/846/files#diff-6515172b6e428e006dc3a691b925404dL22). The page object for that component [scopes by that class](https://github.com/code-corps/code-corps-ember/blob/672_use-BEM-for-styles/tests/pages/components/categories-list.js#L7). For this reason, the test fails as it cannot find the element. The solution to this is to apply a new `data-test-selector` attribute to the corresponding component, and scoping by that in the page object.

To add a `data-` attribute to the component you'll need to follow [the guides](https://guides.emberjs.com/v2.5.0/templates/binding-element-attributes/#toc_adding-data-attributes)

basically you'll be doing the following:

```hbs
{{!-- app/templates/start/interests.hbs --}}
{{categories-list categories=model data-test-selector="Categories List"}}
```

and in the component
```js
// app/components/categories-list.js
...
  attributeBindings: ['data-test-selector'],  // <-- you'll need to add the attributeBindings
  classNames: ['start__interests'],
...
```

Now that the `data-test-selector` is set up we can scope by it in the page object
```js
// tests/pages/components/categories-list.js
...
export default {
  scope: '[data-test-selector="Categories List"]',
...
```

This is just an example of one of the changes, but I hope it sheds some light on what was meant by the decoupling of tests from css. Let me know if you have any questions, I'd be more than happy to help.

JoshSmith:
@WenInCode is this documented in our tasks about how to do this, or somewhere else? If not, think good to add so those tasks can be legitimately labeled `good for new contributors`.

WenInCode:
@JoshSmith It is not currently, I just wrote it up last night for this. I can add it to #698 though. Might be good to mention somewhere in contributing docs as well for new folks contributing.

JoshSmith:
@WenInCode agreed

JoshSmith:
🙌  @rlueder

PR #847, opened 2016-12-09
Change to can accept donations from charges enabled

JoshSmith:
# What's in this PR?

- Adds in `canAcceptDonations` to `stripe-connect-account`
- Changes acceptance behavior for creating donations to `charges_enabled` being `true` _or_ this being a non-production Stripe account (where `charges_enabled` can _never_ be `true` on Stripe's API, since this means for a _live_ account)
- Ember now relies on the new behavior


JoshSmith:
🙌  thanks @appleJax

PR #857, opened 2016-12-12
Add task-list model

### LINKED ISSUE #855: Create Model for Task-List
# Problem
From the [corresponding API issue](https://github.com/code-corps/code-corps-api/issues/554)

> Currently, a project in code corps has many tasks. We need to create a new layer, 'task list', that allows ordered grouping of those tasks within a project. The task list should have the following properties:
> 
> * name: string
> * position: integer
> * tasks: has_many tasks
> * project: belongs_to project

We'll also need to create/update mirage models & endpoints

## Other model changes
**Project:**
- hasMany TaskLists

**Tasks:**
- belongsTo TaskList

## Subtasks
- [x] Create task-list model & mirage model
- [x] Update project & tasks models & mirage models
- [x] Create task-list mirage endpoint

## References
- [project model](https://github.com/code-corps/code-corps-ember/blob/develop/app/models/project.js)
- [project mirage model](https://github.com/code-corps/code-corps-ember/blob/develop/mirage/models/project.js)
- [task model](https://github.com/code-corps/code-corps-ember/blob/develop/app/models/task.js)
- [task mirage model](https://github.com/code-corps/code-corps-ember/blob/develop/mirage/models/task.js)
- [corresponding API issue](https://github.com/code-corps/code-corps-api/issues/554)

WenInCode:
# What's in this PR?
- Adds the `task-list` model
- Updates `project` & `task` models
- Updates related mirage config

## References
Fixes #855

### DIFF
--- a/app/models/task-list.js
+++ b/app/models/task-list.js
@@ -1,0 +1,55 @@
+import Model from 'ember-data/model';
+import attr from 'ember-data/attr';
+import { belongsTo, hasMany } from 'ember-data/relationships';
+
+/**
+  Task-list is a collection of categorized tasks that belong to a project.
+
+  @class task-list
+  @module Model
+  @extends Ember.Model
+ */
+export default Model.extend({
+
+  /**
+    Task-lists name
+
+    @attribute name
+    @type string
+   */
+  name: attr(),
+
+  /**
+    Order is a read-only attribute computed from the `position` attribute
+
+    @attribute order
+    @readonly
+    @type number
+   */
+  order: attr(),
+
+  /**
+    Position is a virtual (write-only) attribute used to compute the `order` of the task-list
+
+    @attribute position
+    @virtual
+    @type number
+   */
+  position: attr(),
+
+  /**
+    The project that the task-list belongs to.
+
+    @attribute project
+    @type Ember.computed
+   */
+  project: belongsTo('project', { async: true }),
+
+  /**
+    The tasks that belong to the task-list.
+
+    @attribute tasks
+    @type Ember.computed
+   */
+  tasks: hasMany('task', { async: true })
+});
--- a/app/models/task.js
+++ b/app/models/task.js
@@ -3,18 +3,134 @@ import attr from 'ember-data/attr';
 import { belongsTo, hasMany } from 'ember-data/relationships';
 import ContainsCodeMixin from '../mixins/contains-code';
 
+/**
+  Task is a (issue/idea/task) created by a user for a project.
+
+  @class task
+  @module Model
+  @extends Ember.Model
+ */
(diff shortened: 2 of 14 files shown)


WenInCode:
This is my first time working with mirage, so I mostly just based it off of the other models. Ready for code review though.

app/models/task-list.js
```diff
@@ -0,0 +1,11 @@
+import Model from 'ember-data/model';
+import attr from 'ember-data/attr';
+import { belongsTo, hasMany } from 'ember-data/relationships';
+
+export default Model.extend({
+  name: attr('string'),
+  position: attr('number'),
```
JoshSmith: I think you can just do `attr()` on both without arguments.

app/models/task.js
```diff
@@ -8,12 +8,14 @@ export default Model.extend(ContainsCodeMixin, {
   insertedAt: attr('date'),
   markdown: attr(),
   number: attr('number'),
+  position: attr('number'),
```
JoshSmith: Same here.

app/models/task-list.js
JoshSmith: We may want to wait on naming convention until this settles in https://github.com/code-corps/code-corps-api/pull/557

@green-arrow are discussing `rank` vs `order`.

One thing that is missing here, though, is the distinction, which will require some manual work in the mirage routes for the task. We'll need to make it so that `position` can be set on the model, but it's a virtual attribute on the server that then gets turned into `rank` or `order`, which is a completely different number.

This is a little odd but setting and getting are not the same thing, or even the same property.

WenInCode:
@JoshSmith updated, not sure if these were the changes you were looking for

mirage/factories/task-list.js
```diff
@@ -1,6 +1,9 @@
 import { Factory } from 'ember-cli-mirage';
 
 export default Factory.extend({
+  order() {
+    return (this.position || 0) * 3;
+  },
```
JoshSmith: I would make this multiplier significantly larger, like 100.

app/models/task.js
```diff
@@ -8,7 +8,8 @@ export default Model.extend(ContainsCodeMixin, {
   insertedAt: attr('date'),
   markdown: attr(),
   number: attr('number'),
-  position: attr('number'),
+  order: attr(),
+  position: attr(),
```
JoshSmith: I would set `position` below the others separately (see `stripe-platform-card`) and write that it is a virtual, write-only attribute. Would also document that `order` is read-only.

JoshSmith:
@WenInCode LGTM!

WenInCode:
great - thanks @JoshSmith

PR #877, opened 2016-12-15
Add birthdate select

JoshSmith:
# What's in this PR?

Work in progress.

Adds a birth date selector component to encapsulate setting the user's birth date.


JoshSmith:
The business type is not needed.

Issue #853, opened 2016-12-11
Create Stripe account form that handles the various states for creating a managed account


JoshSmith:
# Problem

We need a form to create a Stripe account. This form should be on a separate page from the `settings/donations` and should probably be under `settings/donations/payments` instead. The current `settings/donations` should actually probably be `settings/donations/goals`.

A project should be able to create donation goals at any time so that a project can keep setting up even while their identity is being verified. But if the account isn't yet able to take donations, there should be at least some information on the goals page with a link to take you to the payments page.

Some screenshots that show the ideal setup flow, using Kickstarter as an example (with some minor differences from the flow laid out below):

- Selected individual (form) https://cl.ly/1m0B2H1h0Z3Q
- Selected legal entity (form) https://cl.ly/2M2z0i2v373n
- Verifying details https://cl.ly/1u1r3e070N0G
- Document verification needed (file upload) https://cl.ly/281p1U28193j
- Verifying documents https://cl.ly/3I1R151f442S
- Bank account details (form) https://cl.ly/2p3l2F3J1e05
- Completely verified https://cl.ly/2Q1p2x3Z0q20

First, a legend for the form:

- Conditional logic
> o Radio button
> [Button]

## Form

> Create a Stripe account for {{organization name}}.
> The location of either your permanent residence or your business.

Create dropdown of countries, can hard-code US now as only country with country code

> [Continue]

### Contact details
- Email verified / not verified (can skip for now, not MVP)
We'll show an email field with a reverify button.

#### Individual or legal entity
- Needs identity - `verification[fields_needed]` `legal_entity.type`
If the `legal_entity.type` isn't set yet, it means we probably have not allowed you to pick your identity. Here you can set Individual/Business, then fill in all the required form fields for both.

Radio button to indicate whether individual or legal entity (`legal_entity.type`)
> o Individual
> o Legal entity (company or organization)

#### Business information
(show if company is selected)
business name, tax id

#### Personal information
(show always)
first name, last name, date of birth (day, month, year), personal address line 1, personal address line 2, personal address city, personal address state, personal address postal code, personal address country (hardcoded to 2 digit iso), ssn last 4
> [Submit]

- Needs identify verification
check if verified with `legal_entity.verification` - `status` (`unverified`, `pending`, `verified`)
if `unverified` and `verification[fields_needed]` with:
	- `legal_entity.personal_id_number` - display full SSN field and encrypt with PII token (with ember-stripe-service)
[...]

PR #878, opened 2016-12-15
added country dropdown

### LINKED ISSUE #864: Change country dropdown
# Problem

In `components/payments/funds-recipient.hbs` we need to change the country dropdown to not be disabled and only have the US.

rileytaylor:
# What's in this PR?

This enables the country dropdown and makes it a proper select dropdown. The only option, for now, is 'US'

## References
Fixes #864 

Progress on: #853

### DIFF
--- a/app/components/select/country-select.js
+++ b/app/components/select/country-select.js
@@ -1,0 +1,12 @@
+import Ember from 'ember';
+
+const {
+  Component
+} = Ember;
+
+export default Component.extend({
+  classNames: ['select-country'],
+  countryOptions: [
+    { name: 'United States', id: 'US' }
+  ]
+});
--- a/app/templates/components/payments/funds-recipient.hbs
+++ b/app/templates/components/payments/funds-recipient.hbs
@@ -56,9 +56,7 @@
   </div>
   <div class="input-group">
     <label for="country">Country</label>
-    <select value=country name="country">
-      <option value="US">United States</option>
-    </select>
+    {{select/country-select country=country}}
   </div>
 </div>
 <div class="input-group">
--- a/tests/integration/components/select/country-select-test.js
+++ b/tests/integration/components/select/country-select-test.js
@@ -1,0 +1,31 @@
+import { moduleForComponent, test } from 'ember-qunit';
+import Ember from 'ember';
+import hbs from 'htmlbars-inline-precompile';
+import selectCountry from '../../../pages/components/select/country-select';
+import PageObject from 'ember-cli-page-object';
+
+const {
+  get,
+  set
+} = Ember;
+
+let page = PageObject.create(selectCountry);
+
+moduleForComponent('select/country-select', 'Integration | Component | select/country select', {
+  integration: true,
+  beforeEach() {
+    page.setContext(this);
+  },
+  afterEach() {
+    page.setContext(this);
+  }
+});
+
+test('it sets the correct country', function(assert) {
+  assert.expect(1);
+  set(this, 'country', 'US');
+  page.render(hbs`{{select/country-select country=country}}`);
+  page.country.fillIn('US');
+  let country = get(this, 'country');
+  assert.equal(country, 'US');
+});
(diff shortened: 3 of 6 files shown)


rileytaylor:
I'm going to redo this using the x-select component, display the country as a full name in the select ui. and I also need to add tests.

JoshSmith:
@rileytaylor you can look to my other open PR for ideas on how to test with selectable.

tests/integration/components/select/country-select-test.js
```diff
@@ -0,0 +1,25 @@
...
+  beforeEach() {
+    page.setContext(this);
+  },
+  afterEach() {
+    page.setContext(this);
+  }
+});
+
+test('it sets the country strings correctly', function(assert) {
+
```
JoshSmith: Think you can use the existing tests in the `select/birth-date`

tests/integration/components/select/country-select-test.js
```diff
@@ -0,0 +1,31 @@
...
+});
+
+test('it sets the correct country', function(assert) {
+  assert.expect(1);
+
+  this.set('country', 'US');
+
+  page.render(hbs`{{select/country-select country=country}}`);
+
+  page.country.fillIn('US');
```
JoshSmith: Can you only `fillIn` with the `value` and not the full name.

tests/integration/components/select/country-select-test.js
```diff
@@ -0,0 +1,31 @@
...
+
+  let country = this.get('country');
+
+  assert.equal(country, 'US');
```
JoshSmith: I would definitely condense all of these down without gaps between lines.

tests/integration/components/select/country-select-test.js
```diff
@@ -0,0 +1,31 @@
...
+test('it sets the correct country', function(assert) {
+  assert.expect(1);
+
+  this.set('country', 'US');
+
+  page.render(hbs`{{select/country-select country=country}}`);
+
+  page.country.fillIn('US');
+
+  let country = this.get('country');
```
JoshSmith: Prefer `get(this, 'prop')` and `set(this, 'prop')`. `const { get, set } = Ember;`

app/components/select/country-select.js
```diff
@@ -0,0 +1,12 @@
+import Ember from 'ember';
+
+const {
+  Component
+ } = Ember;
+
+export default Component.extend({
+  classNames: ['select-country'],
+  countryOptions: [
+    { name:'United States', abbr:'US' }
```
JoshSmith: Would change `abbr` to `id`. Also would space after each `:`

app/components/select/country-select.js
```diff
@@ -0,0 +1,12 @@
+import Ember from 'ember';
+
+const {
+  Component
+ } = Ember;
```
JoshSmith: Extra space before `}`.

tests/integration/components/select/country-select-test.js
rileytaylor: I'm not sure what you mean by this, isn't `US` is the value?

tests/integration/components/select/country-select-test.js
rileytaylor: Should `birth-date` also use this? I agree this is better I just didn't do it initially since i didn't see any other tests using it

JoshSmith: Yeah, it should.

tests/integration/components/select/country-select-test.js
JoshSmith: I'm asking if we can pass "United States" instead, more out of curiosity.

app/templates/components/payments/funds-recipient.hbs
```diff
@@ -55,10 +55,7 @@
     {{input type="text" name="zip" value=zip}}
   </div>
   <div class="input-group">
-    <label for="country">Country</label>
```
JoshSmith: Would add this `label` back.

JoshSmith:
🙌

PR #880, opened 2016-12-16
update state drop

### LINKED ISSUE #863: Create a state dropdown
# Problem

#849 added a basic state dropdown that needs to be populated with actual data.

https://github.com/code-corps/code-corps-ember/pull/849/files#diff-fadc9970a9b163a9321d697cbdd42755R63

rileytaylor:
# What's in this PR?

This updates the state dropdown to include a list of states and their corresponding postal codes.

## References
Fixes #863 

Progress on: #853

### DIFF
--- a/app/components/select/state-select.js
+++ b/app/components/select/state-select.js
@@ -1,0 +1,65 @@
+import Ember from 'ember';
+
+const {
+  Component
+} = Ember;
+
+export default Component.extend({
+  classNames: ['select-state'],
+  // Supported states/territories:
+  // https://support.stripe.com/questions/can-i-use-stripe-if-my-business-is-based-in-an-outlying-territory-of-a-supported
+  stateOptions: [
+    { name: 'Alabama', id: 'AL' },
+    { name: 'Alaska', id: 'AK' },
+    { name: 'Arizona', id: 'AZ' },
+    { name: 'Arkansas', id: 'AR' },
+    { name: 'California', id: 'CA' },
+    { name: 'Colorado', id: 'CO' },
+    { name: 'Connecticut', id: 'CT' },
+    { name: 'Delaware', id: 'DE' },
+    { name: 'District of Columbia', id: 'DC' },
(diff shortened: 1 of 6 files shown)


app/components/select/state-select.js
```diff
@@ -0,0 +1,63 @@
+import Ember from 'ember';
+
+const {
+  Component
+} = Ember;
+
+export default Component.extend({
+  classNames: ['select-state'],
+  stateOptions: [
+    { name: 'Alabama', id: 'AL' },
```
JoshSmith: I would note this article link in here for documentation: https://support.stripe.com/questions/can-i-use-stripe-if-my-business-is-based-in-an-outlying-territory-of-a-supported

tests/integration/components/select/state-select-test.js
```diff
@@ -0,0 +1,31 @@
...
+
+let page = PageObject.create(selectState);
+
+moduleForComponent('select/state-select', 'Integration | Component | select/state select', {
+  integration: true,
+  beforeEach() {
+    page.setContext(this);
+  },
+  afterEach() {
+    page.setContext(this);
```
JoshSmith: This should be `page.removeContext();`. We may have this wrong in some other tests. Like country, it looks like. Can fix in this PR.

tests/integration/components/select/state-select-test.js
```diff
@@ -0,0 +1,31 @@
...
+
+let page = PageObject.create(selectState);
+
+moduleForComponent('select/state-select', 'Integration | Component | select/state select', {
+  integration: true,
+  beforeEach() {
+    page.setContext(this);
+  },
+  afterEach() {
+    page.removeContext(this);
```
JoshSmith: `page.removeContext()`

rileytaylor: oh duh.

JoshSmith:
@rileytaylor LGTM. Merge when passing.

PR #888, opened 2016-12-17
Template, styles and behavior for bank-account component

### LINKED ISSUE #860: Update payments/bank-account component logic
# Problem

The `payments/bank-account` component will take in a `stripeConnectAccount` model and a `stripeExternalAccount` model and render several states based on its `bank_account_status`:

- `pending_requirement` - the "Bank account" box with no other information, no fields, nothing; only the title
- `required` - we need your bank account info w/ form and submit
- `verified` - show the success state with the bank account last 4 number and the full routing number

When it's `required`, we'll have a form for adding your  bank account routing number and account number, and use that in combination with `ember-stripe-service` to create a bank account token. We need to rewrite `onBankAccountInformationSubmitted` from #849 to be something more like `addBankAccount`, which will be a route action that will update the `StripeConnectAccount` with the token. This action will need to return a promise that can be used for changing loading status of the button.

The token will be used to set the `external_account` field of `StripeConnectAccount` with the token.

There is no `verifying` status for this component.

rlueder:
# What's in this PR?

Updates to bank-account component. 

## Notes

- I used font-awesome just to get something quickly in place instead of editing our sprites
- I added a click action to the first state of the page just so we can test the different states

## TODO

- [x] Implement green check icon in icons sprite
- [x] Remove font-awesome
- [x] Remove test js
- [x] Add `ember-stripe-service` token generation
- [ ] Add route action
- [x] Add button loading state

## References
Fixes #860


JoshSmith:
@begedin I didn't get to it but the `.account-setup` classes should go in the `account-setup.scss` component template instead of in here. Those classes and the general HTML structure here is intended to be reused across the other components.

I can go back through and make the green check icon work. Don't replicate that across the other issues. I'll fix it here, and introduce the new styles elsewhere.

JoshSmith:
These will also need the actual values bound to the real values, and the account number last 4 will have to be used here instead, along with minor things I'm probably not mentioning.

begedin:
I did not use the route action for now, reason being I didn't want to potentially waste time and figuring out how to test that it's triggering properly. Instead, for now, the action is just passed out using closures. 

We can switch to the route action helper when we have the time.

begedin:
This took me longer than I thought due to me initially not realizing I need to implement the route action as well. I thought it would be just the component, not the whole logic chain, so I went for this before doing the recipient part in the other issue.

However, I got it as close as possible. The process consists of two steps - creating a token and adding the external account to the stripe connect account. Each step catches and handles it's own errors (a friendly string is displayed, similar to how it's done on the donate page.

Token creation was tested manually. Adding the account to our api record is untested and might need adjustments, but seems pretty straightforward for now.

An acceptance test will be need it. We can write it once we figure out the API parts and know the names of properties. Not sure if that's going to be part of this PR, or separate.

Issue #885, opened 2016-12-16
Create a payments/funds-recipient/verification-document component


JoshSmith:
# Problem

**EDIT**: The PR #892 itself has more up-to-date requirements.

The `payments/funds-recipient/verification-document` component will take in a `stripeConnectAccount` model and render several states based on its `verification_document_status`:

- `pending_requirement` - nothing
- `required` - we need you to upload a scan of your ID (form + submit)
- `verifying` - we're verifying your id scan
- `verified` - nothing, parent responsible now

When it's `required`, we'll have a form for uploading a file, and use that to send a route action which will then create a `StripeFileUpload`. This action will need to return a promise that can be used for changing loading status of the button.

When it's `verifying`, we'll say something like:

> Please be patient while we review the document you provided.

The exact document type may vary depending on whether the account was an `individual` or `legal_entity`.

The allowed file types are JPEG and PNG.

JoshSmith:
Closed by #892.

PR #892, opened 2016-12-17
Introduce verification-doc component with 4 states

sarupbanskota:
Progress on: #885

In the `identity-document-file-upload` component we'll:

- [x] set the upload URL to `https://uploads.stripe.com/v1/files`
- [x] change [`paramName`](https://github.com/benefitcloud/ember-uploader#change-parameters) to `file`
- [x] add the `purpose` param with value of `"identity_document"`
- [x] add a `Stripe-Account` [header](https://github.com/benefitcloud/ember-uploader#modifying-the-request) to accept the `stripeConnectAccount.idFromStripe`
- [x] add a `Basic` auth [header](https://github.com/benefitcloud/ember-uploader#modifying-the-request) to have a username of our `ENV.stripe.publishableKey` with no password – this probably needs to be base64 encoded (e.g. `pk_test_123:` would be encoded)

Ideally we should:

- [x] validate the file size is less than 8MB
- [x] validate it is either a JPEG or PNG

When [the upload is finished](https://github.com/benefitcloud/ember-uploader#finished-uploading), we'll need to:

- [x] send the `id` value found in the response in an action to the `verification-document` component
- [x] the `verification-document` component will then call an action to set the `identityDocumentId` field on the `StripeConnectAccount` model

### DIFF
--- a/app/components/payments/funds-recipient/identity-document-file-upload.js
+++ b/app/components/payments/funds-recipient/identity-document-file-upload.js
@@ -1,0 +1,106 @@
+import Ember from 'ember';
+import EmberUploader from 'ember-uploader';
+import ENV from 'code-corps-ember/config/environment';
+
+const {
+  computed,
+  get,
+  getProperties,
+  isEmpty
+} = Ember;
+
+const { Uploader } = EmberUploader;
+
+/**
+ * `payments/funds-recipient/identity-document-file-upload` provides a file input
+ * for uploading an identity verification document to stripe.
+ *
+ *
+ * @class identity-document-file-upload
+ * @module Component
+ * @extends EmberUploader.FileField
+ */
+export default EmberUploader.FileField.extend({
+  classNames: ['identity-document-file-upload'],
+
+  /**
+   * Object containing additional properties to be passed
+   * in as part of the form data being uploaded
+   * @type {Object}
+   */
+  additionalUploadData: {
+    // required to authorize the upload request
+    key: ENV.stripe.publishableKey,
+    purpose: 'identity_document'
+  },
+
+  maxFileSize: 1024 * 1024 * 8, // 8mb
+  multiple: false,
+  supportedFileTypes: ['image/jpeg', 'image/jpg', 'image/png'],
+  url: 'https://uploads.stripe.com/v1/files',
+
+  /**
+   * A computed property containing settings for the ajax request
+   * Used to set stripe account id in the request header
+   * @return {Object}
+   */
+  ajaxSettings: computed('stripeConnectAccount', function() {
+    let headers = {
+      'Stripe-Account': get(this, 'stripeConnectAccount.idFromStripe')
+    };
+
+    return { headers };
+  }),
+
+  /**
+   * Triggers when the file selection for the rendered file input changes
+   * @param {[File]} files An array of files selected by the user.
+   *                        Since the `multiple` setting is set to false, only 1 file
+   *                        is in the array.
+   */
+  filesDidChange(files) {
+    if (!isEmpty(files) && this._validate(files[0])) {
+      this._performUpload(files[0]);
+    }
+  },
+
+  _validate({ size, type }) {
+    let { maxFileSize, supportedFileTypes } = getProperties(this, 'maxFileSize', 'supportedFileTypes');
+    let isValid = (size <= maxFileSize) && (supportedFileTypes.indexOf(type) >= 0);
+
+    if (!isValid) {
+      this.sendAction('validationError');
+    }
+
+    return isValid;
+  },
+
+  _performUpload(file) {
+    this.sendAction('uploadStarted');
+
+    let params = getProperties(this, 'url', 'ajaxSettings');
+    let uploader = Uploader.create(params);
+
+    uploader.on('progress', (event) => this._handleUploadProgress(event));
+
+    let additionalUploadData = get(this, 'additionalUploadData');
+
+    uploader.upload(file, additionalUploadData)
+            .then((event) => this._handleUploadDone(event))
+            .catch((reason) => this._handleUploadError(reason));
+  },
+
+  // error handlers
+
+  _handleUploadDone({ id }) {
+    this.sendAction('uploadDone', id);
+  },
+
+  _handleUploadError(reason) {
+    this.sendAction('uploadError', reason);
+  },
+
+  _handleUploadProgress(event) {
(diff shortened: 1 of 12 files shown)


app/templates/components/payments/funds-recipient/verification-document.hbs
```diff
@@ -0,0 +1,9 @@
+{{#if (eq stripleConnectAccount.verification_document_status 'pending_requirement')}}
+  {{!-- Show nothing --}}
+{{else if (eq stripleConnectAccount.verification_document_status 'required')}}
+  we need you to upload a scan of your ID (form + submit)
+{{else if (eq stripleConnectAccount.verification_document_status 'verifying')}}
+  Please be patient while we review the document you provided.
+{{else if (eq stripleConnectAccount.verification_document_status 'verified')}}
```
JoshSmith: Just a small typo here `stripleConnectAccount`

tests/integration/components/payments/funds-recipient/verification-document-test.js
```diff
@@ -0,0 +1,25 @@
...
+  assert.equal(this.$().text().trim(), '');
+
+  // Template block usage:
+  this.render(hbs`
+    {{#payments/funds-recipient/verification-document}}
+      template block text
+    {{/payments/funds-recipient/verification-document}}
+  `);
+
+  assert.equal(this.$().text().trim(), 'template block text');
```
JoshSmith: You'll need to introduce a new page object for these tests, when you get there.

app/templates/components/payments/funds-recipient/verification-document.hbs
JoshSmith: I would just do `if (eq stripleConnectAccount.verification_document_status 'required')` and `if (eq stripleConnectAccount.verification_document_status 'verifying')` without the branching. Otherwise, it will by default render nothing.

sarupbanskota: mm, didn't quite understand; why would it render nothing @JoshSmith?

tests/integration/components/payments/funds-recipient/verification-document-test.js
sarupbanskota: sure, could you point me to an example on another test file?

JoshSmith: Sure, the `bank-account-test` which is a sister component is a good example.

app/templates/components/payments/funds-recipient/verification-document.hbs
JoshSmith: @sarupbanskota I'm saying we only want to handle the two cases of `required` and `verifying`. No need to branch then with your conditionals. This component renders something only in those two cases.

sarupbanskota: ah ok!

begedin:
@sarupbanskota Good job up to this point. I'll take over and try to wrap up before you're back.

begedin:
This checks all the boxes now, but there are some notes that I would like to add:

* We only have some basic tests that test bindings and various `if` switches in the templates. We need to figure out how to test uploads themselves, and handling of events triggered by the `ember-uploader` component
* The whole thing needs styling (all the way up to the `funds-recipient` component.
* Identifying is not done via an authorization header. Instead, it's done by adding a `key` field to the form data being sent by the upload, which is set to the publishable key. Is there documentation available indicating that a basic authorization header would be preferable? I Couldn't find anything.

Either way, the process here does work. It will upload the file to stripe, then pass out the action with a stripe id parameter all the way to the route, where the route will assign it to the connect account and try to save that connect account. 

This fails right now, because the API has no such endpoint, but all the failure points that I could think of are handled in a basic way - by rendering a friendly error. The errors rendered are:
* file validation fails
* upload fails
* assigning file to account fails

app/components/payments/funds-recipient/identity-document-file-upload.js
```diff
@@ -0,0 +1,89 @@
...
+  }),
+
+  /**
+   * Object containing additional properties to be passed
+   * in as part of the form data being uploaded
+   * @type {Object}
+   */
+  additionalUploadData: {
+    // required to authorize the upload request
+    key: ENV.stripe.publishableKey,
```
JoshSmith: This actually works? I only saw the basic auth header in the docs.

begedin: I couldn't find that part at all. Only found this one.

begedin:
It's not just that we aren't testing internals. The addon doesn't allow much in appearance stuff, so we have to basically wrap it up into an external addon, which handles the event hooks. That external addon would then be the one that displays progress/errors/etc., so we ought to test that one to ensure all of the features work. I'm not sure how, though.

In an integration test, we would have to mock a web request (to upload), which sends a progress report, as well as mocking file selection.

In a unit test, those things could be mocked by calling the handlers manually, but then we aren't testing the whole surface area.

That being said, I think it should be mergeable as is, but we need to create an issue to figure that out later.

JoshSmith:
Haven't had a chance to code review but there are some conflicts that need resolving now. Would prefer keeping the `account` => `stripeConnectAccount` changes that were made.

JoshSmith:
I might be able to fix these changes when I get home this evening.

begedin:
This should be good to go now.

app/components/payments/funds-recipient/identity-document-file-upload.js
```diff
@@ -0,0 +1,89 @@
...
+    purpose: 'identity_document'
+  },
+
+  filesDidChange(files) {
+    if (!isEmpty(files) && this._validate(files[0])) {
+      this._performUpload(files[0]);
+    }
+  },
+
+  _validate({ type, size }) {
```
JoshSmith: Would alpha order these.

app/controllers/project/settings/donations/payments.js
```diff
@@ -11,60 +11,99 @@ const {
...
+
+    return account.save()
+                  .then((account) => RSVP.resolve(account))
+                  .catch((reason) => this._handleIdentityVerificationDocumentError(reason));
+  },
+
+  _handleIdentityVerificationDocumentError() {
+    let friendlyError = new FriendlyError(VERIFICATION_DOCUMENT_ERROR);
+    return RSVP.reject(friendlyError);
+  },
```
JoshSmith: Would also alpha-order here.

PR #893, opened 2016-12-17
Add stripe-file-upload model, mirage; small component fiix

JoshSmith:
# What's in this PR?

Add file upload model and tests.

Tiny fix on a component, removing something not needed.


JoshSmith:
Closed since we don't need this.

PR #894, opened 2016-12-18
Update funds recipents form

### LINKED ISSUE #884: Create a payments/funds-recipient/details-form component
# Problem

The `payments/funds-recipient/details-form` component will live inside of the `funds-recipient` component.

It will have the recipient type (business or individual), form fields for both types, and a submit button.

It can send a route action up to the route upon submission. Ideally this action would expect a promise to be returned from the route so that the button can load asynchronously until we get a response.

Depending on the `stripeConnectAccount` model's `recipient_status`:

- `required` - show the form fields, including a radio button for selecting individual or legal entity
  - if `individual` - show the personal information fields
  - if `legal_entity` - show the personal _and_ legal entity fields (with legal entity fields shown first)
- `errored` - show the fields with any relevant errors as to what's missing; can come from the `fields_needed` hash and used to match to those fields

marineb:
# What's in this PR?
 
- didn't complete the org / individual toggle at the top. I wasn't sure how to extend the target area of the radio button with ember
- I know my class names and the general markup structure aren't great :(
- suggested some language changes (not sure what the requirements are here, just thought i'd share some diff options) 
- in the address section, I was wondering if all these fields are required for cc processing? (or could you have only the zipcode instead?) 

## References
Fixes #884
Design https://cl.ly/2o3C0p1Y1H2T/o.

### DIFF
--- a/app/components/payments/account-setup.js
+++ b/app/components/payments/account-setup.js
@@ -5,33 +5,5 @@ const {
 } = Ember;
 
 export default Component.extend({
-  classNames: ['account-setup'],
-
-  // set by binding
-  email: null,
-
-  // Preset values for testing
-  fundsRecipient: {
-    recipientType: 'company',
-    firstName: 'John',
-    lastName: 'Doe',
-
-    businessType: 'sole_prop',
-    businessName: 'Managed LLC',
-    businessEin: '1234-managed',
-
-    dob: '06-12-1986',
-
-    dobDay: '06',
-    dobMonth: '12',
-    dobYear: '1986',
-
-    address1: 'Some street 22',
-    address2: 'PO 23',
-    city: 'Los Angeles',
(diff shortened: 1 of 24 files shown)


JoshSmith:
@marine this is just for accepting donations, not for paying. These are strict requirements under know your customer laws and Stripe's needs.

begedin:
Good work so far. I'll take over for now, since I'm available, and it's high priority.

marineb:
Thank you @begedin. Happy to style the radio buttons later. Lmk!

begedin:
This ought to be good to review. The form has all the fields and submits correctly based on recipient type. The two integration tests I wrote ensure that.

There are minor details on how the radio buttons look (wasn't sure about correct margins/paddings), but overall, it follows the provided designs, I think.

PR #896, opened 2016-12-19
Minor design fixes for payment forms

JoshSmith:
# What's in this PR?

- Fixes width/height overriding in check mark


begedin:
Is this good for review? Seems OK to me, but not sure if we wanna wait and add more potential fixes.

PR #906, opened 2016-12-22
payments/funds recipient/personal id number component

### LINKED ISSUE #886: Create a payments/funds-recipient/personal-id-number component
# Problem

The `payments/funds-recipient/personal-id-number` component will take in a `stripeConnectAccount` model and render several states based on its `personal_id_number_status`:

- `pending_requirement` - nothing
- `required` - we need your full personal ID number (form + submit)
- `verifying` - we're verifying your id number
- `verified` - nothing, parent responsible now

When it's `required`, we'll have a form for adding your full personal ID number, and use that in combination with `ember-stripe-service` to create a PII token. We'll then send a route action which which will update the `StripeConnectAccount` with the token. This action will need to return a promise that can be used for changing loading status of the button.

When it's `verifying`, we'll say something like:

> Please be patient while we verify your ID number.

begedin:
# What's in this PR?

Should be good to go. I expanded the funds-recipient tests a bit as well, and sadly, got stuck at failing to produce a useful test for the document selection process.

## References
Fixes #886

### DIFF
--- a/app/components/payments/funds-recipient/personal-id-number.js
+++ b/app/components/payments/funds-recipient/personal-id-number.js
@@ -1,0 +1,11 @@
+import Ember from 'ember';
+
+const {
+  Component,
+  computed: { alias }
+} = Ember;
+
+export default Component.extend({
+  classNames: ['personal-id-number'],
+  status: alias('stripeConnectAccount.personalIdNumberStatus')
+});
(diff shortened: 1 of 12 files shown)


begedin:
@JoshSmith Should be good to review

PR #907, opened 2016-12-22
Update tasks

JoshSmith:
# What's in this PR?

WIP. Update the tasks lists.

There's an issue with dragging/dropping tasks that we need to resolve because of siblings.


JoshSmith:
This needs fixes for tests since it looks like some async behavior was added in that didn't exist previously.

PR #918, opened 2016-12-29
Fix issue with ember-page-title causing tests to fail

begedin:
# What's in this PR?

Cloning the repo and running an NPM install was causing two issues for me, which I spent most of the day resolving

1. We explicitly use phantomjs 1.9.20, which is failing to start on my MacBook (sierra). Removing that dependency resolves that issue. 1.9.20 is generally not supported any more, so I don't think we should be using it.
2. Regardless of phantomjs, tests are failing due to an issue I opened in ember-page-object (https://github.com/tim-evans/ember-page-title/issues/48). In the issue description, I described a workaround, which I applied here.

### DIFF
--- a/package.json
+++ b/package.json
@@ -76,7 +76,7 @@
     "ember-metrics": "0.7.0",
     "ember-modal-dialog": "0.8.8",
     "ember-moment": "7.2.0",
-    "ember-page-title": "^3.0.10",
+    "ember-page-title": "^3.1.2",
     "ember-radio-button": "^1.0.7",
     "ember-resolver": "^2.0.3",
     "ember-route-action-helper": "^2.0.0",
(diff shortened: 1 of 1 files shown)


JoshSmith:
@begedin given that https://github.com/tim-evans/ember-page-title/pull/49 is merged, can this be closed completely?

begedin:
I would still merge in the removal of explicit dependency on phantomjs 1.9.20. Anything 1.9.x crashes on sierra for me and from what I can tell from googling, everyone else. I'll update this PR so it upgrades ember-page-title and keeps in the removal of 1.9.20

begedin:
I changed the version number in package.json and removed the fix, but the actual v3.1.2 has not been published on npm yet, so the build will fail. I'll contact tim so he can publish.

begedin:
I contacted @tim-evans, so as soon as 3.1.2 is published on NPM, this build should pass and the PR can be merged.

begedin:
I'll rebase and rename the commit once build passes, just for some nicer, more accurate commit history.

begedin:
Blocked until 3.1.2 is the latest version at https://www.npmjs.com/package/ember-page-title

begedin:
@JoshSmith New version is up and I performed a rebuild without cache, just in case. 

Could you also make sure this works for you locally (be sure to nuke `node_modules` or `node_modules/phantomjs` at least) and then merge?

PR #928, opened 2017-01-02
Add Terms of Use parameters and UI to account creation

### LINKED ISSUE #925: Add terms acceptance
# Problem

On the payments settings page we need to add a Stripe terms acceptance. I believe we should add this at the "create account" step.

We'll need to collect the IP address, browser user agent, and the time the form was submitted as a Unix timestamp.

The text will display a link to our terms of service and to Stripe's terms of service.

> By registering your account, you agree to our Terms of Use and the [Stripe Connected Account Agreement](https://stripe.com/connect-account/legal).

begedin:
# What's in this PR?

Adds UI to the account create component indicating that, by creating the account, the user accepts our own and Stripe's terms of use.

Also adds behavior to the component, so that a `tosAcceptanceDate` parameter is sent while creating the account.

## References
Fixes #925

### DIFF
--- a/app/templates/components/payments/create-account.hbs
+++ b/app/templates/components/payments/create-account.hbs
@@ -6,7 +6,8 @@
     </div>
   </div>
   <div class="input-group">
-    <button class="default" {{action onCreateStripeConnectAccount country}} disabled={{isBusy}}>
+    <p>By registering your account, you agree to our {{link-to "Terms of Use" "terms"}} and the <a href="https://stripe.com/connect-account/legal" target="_blank">Stripe Connected Account Agreement.</a></p>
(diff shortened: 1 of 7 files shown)


app/templates/components/payments/create-account.hbs
```diff
@@ -6,7 +6,8 @@
     </div>
   </div>
   <div class="input-group">
-    <button class="default" {{action onCreateStripeConnectAccount country}} disabled={{isBusy}}>
+    <p>By registering your account, you agree to our {{link-to "Terms of Use" "terms"}} and the <a href="https://stripe.com/connect-account/legal" target="_blank">Stripe Connected Account Agreement.</a></p>
```
JoshSmith: Really good thinking on `_blank` here. You're a UX designer in the making!

PR #931, opened 2017-01-03
Add acceptance test for account setup

begedin:
# What's in this PR?

Added an acceptance test for stripe account setup

## References
Fixes #926

PR #932, opened 2017-01-03
Add integration tests for account-setup

### LINKED ISSUE #865: Write component tests for payments/account-setup
# Problem

We need to write component tests for the `payments/account-setup` container.

We need to test that it should:

- render the organization name
- render the `payments/funds-recipient` component
- render the `payments/bank-account` component

There's no need to test much deeper than that these components are rendered with at least some subcomponent, since they will rely on passing down the `stripeConnectAccount` model in order to show their statuses.

We should also remove the `payments-contact` info from being rendered in this component for the time being.

begedin:
# What's in this PR?

Adds integration tests for the `account-setup` component. 

There are likely to be some overlaps with #931 when merging this.

## References
Fixes #865

### DIFF
--- a/tests/pages/components/payments/funds-recipient/verification-document.js
+++ b/tests/pages/components/payments/funds-recipient/verification-document.js
@@ -1,9 +1,23 @@
+import Ember from 'ember';
 import { clickable, isVisible, text } from 'ember-cli-page-object';
+
+const { $, run } = Ember;
 
 export default {
   scope: '.verification-document',
   errorText: text('.error'),
   rendersFileUpload: isVisible('input[type=file]'),
-  clickSubmit: clickable('button')
+  clickSubmit: clickable('button'),
+
+  // since it's near impossible to simulate selecting a file with a filepicker
+  // this does the "next best" thing and triggers the "upload done" action on the component
+  triggerDocumentSubmitted(context, stripeFileId = 'fil_test') {
+    let id = $(this.scope).attr('id');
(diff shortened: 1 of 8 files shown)


tests/pages/components/payments/funds-recipient/verification-document.js
```diff
@@ -1,9 +1,23 @@
...
+  // since it's near impossible to simulate selecting a file with a filepicker
+  // this does the "next best" thing and triggers the "upload done" action on the component
+  triggerDocumentSubmitted(context, stripeFileId = 'fil_test') {
+    let id = $(this.scope).attr('id');
+    let component = context.container.lookup('-view-registry:main')[id];
+
+    run(component, () => {
+      component.get('onUploadDone').call(component, stripeFileId);
+    });
+  }
```
JoshSmith: Should we remove the simulation where we use `sinon` (in the photo drop zone test) and replace with this instead? I know that is one of the things breaking in our tests in Ember 2.10.

JoshSmith:
Question before you merge, but otherwise 👍

tests/pages/components/payments/funds-recipient/verification-document.js
begedin: We might want to. Can extract into helper.

begedin: It didn't work for me now either, btw

PR #946, opened 2017-01-06
Add inbox to task lists

JoshSmith:
# What's in this PR?

Adds the concept of an `inbox` to `TaskList` which is the default container for receiving new tasks in a project.


JoshSmith:
The behavior here is still very hit or miss on whether the task appears to flicker when saving it.

Issue #947, opened 2017-01-07
Disallow activating donations until transfers are available


rileytaylor:
# Problem

Can donation goals be set if you can't take transfers?

JoshSmith:
Yes, this appears to be the case. `{{donations/donation-progress}}` renders right now if `project.donationsActive` is `true`. On the API this means:

```
Enum.any?(project.donation_goals) && project.stripe_connect_plan != nil
```

You can create a `StripeConnectPlan` regardless of whether the account can accept donations.

This should probably lead us to consider some behavioral changes:

- The `project.donationsActive` is fairly misleading right now (on the API side mostly).
- `activateDonations` in Ember is also fairly misleading and should probably be more explicit that this action is creating the `StripeConnectPlan` for the project.
- Could show the button to activate only when transfers can be accepted.

Anything else I'm not thinking of @begedin?

begedin:
I'm not sure about being more explicit with creating the `StripeConnectPlan`. It sort of feels like we're familiarising the user about our internals more than needed. It's definitely less confusing for us, but I'm not sure that would be the case for the user. I think we may want to consider better wording, something like

"Allow this project to start managing donation goals and/or receiving donations"

However,  I don't think this wording mentioning stripe plans would make it any clearer from the user's point of view. 

To make it clearer for us, developers, I'd rather go with inline documentation.

begedin:
As for transfers and when to show and not to show the button, this was initially implemented when we didn't have the payments setup form, so it might make sense to change the conditions now. It might even make sense not to allow activating or editing donations at all until the account setup process has been completed.

JoshSmith:
I think we should allow setting up donations, but disallow activating them. Setting them up can still be done even if verification takes awhile.

JoshSmith:
This way a project can parallelize some work and not feel "stuck".

JoshSmith:
I believe this needs a task on the API side, too.

begedin:
Created code-corps/code-corps-api#630 as a related issue.

On the ember side, the first step in making the process clearer is being done in #969 by simplifying and flattening the menus.

The second step is to add messages to the UI of donation goal and payment pages.

The donation goals page should allow activating donations only if the payments are fully set up. There should be a message linking to the payments page if this requirement has not been met.

The payments page should link to the donation goals page if the payments have been fully setup.

PR #969, opened 2017-01-13
Remove duplicate settings menus

JoshSmith:
# What's in this PR?

First move to remove duplicate settings menus.

Progress on #947


app/styles/_settings.scss
```diff
@@ -1,9 +1,6 @@
-.settings-menu {
-  @include span-columns(2.5);
-}
-
 .settings {
-  @include span-columns(9.5);
+  // @include shift(2.5);
```
begedin: Need to remove this line

PR #973, opened 2017-01-15
Fix donation goal activation

JoshSmith:
# What's in this PR?

Fixes some issues with activating donation goals and splits off a separate component for it.


JoshSmith:
Opened #974 to do this.

PR #978, opened 2017-01-16
Fix account setup

JoshSmith:
# What's in this PR?

Minor changes to account setup.


begedin:
The two commits look separate from each other, so not sure if you wanna squash before merging.

PR #990, opened 2017-01-25
Fix details form behavior on account setup

begedin:
# What's in this PR?

The upgrade of emberx-select contained a bug which this PR should fix. Upon initial load, even though the UI made it seem to be the case, the selects actually weren't setting their bound properties (as they really shouldn't implicitly). This PR explicitly sets these so payment details can now be properly submitted.

While I was at it, I modified the birth date component so that the month and day dropdowns are populated dynamically, based on the yer and the relationship to the current date.

This means

- the day dropdown will contains 30, 31, 28 or 29 days depending on the selected month and/or year
- if the current year is selected, the day and month will contain values at most up to the current date, so the user will not be able to select a date in the future.

### DIFF
--- a/app/components/payments/funds-recipient/details-form.js
+++ b/app/components/payments/funds-recipient/details-form.js
@@ -4,6 +4,7 @@ const {
   Component,
   computed: { equal },
   get,
+  isEmpty,
   set,
   setProperties
 } = Ember;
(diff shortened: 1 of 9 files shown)


JoshSmith:
We want to restrict the year to no earlier than 13 years ago from the current date, and remove any restrictions on month and day.

JoshSmith:
Also it appears that "Individual" is not selected by default on the account creation step now.

begedin:
@JoshSmith I have made the chenges to the birth date component.

However, I was unable to reproduce the issue with the business type selection.

PR #992, opened 2017-01-26
Simplified subscription creation

### LINKED ISSUE #989: Simplify subscription creation
# Problem

This is a "sister" issue to code-corps/code-corps-api#590

What we need to do when creating a subscription:

- fetch the user, quantity and stripeConnectPlan
- create and save a stripeConnectSubscription record using those properties
- get rid of the whole adapterOptions logic
- get rid of the projectId virtual attribute

## References
Dependant on code-corps/code-corps-api#590

begedin:
# What's in this PR?

This PR simplifies the subscription process by relating a subscription to a project directly instead of through a plan. This change makes it so that the plan is not exposed to the regular user through a subscription. Instead, it is only exposed to the organisation owner through the project.

## References
Fixes #989 
Dependent on code-corps/code-corps-api#664


JoshSmith:
I can re-review this once #990 is done.

PR #995, opened 2017-01-27
Fix issue with comment posting

### LINKED ISSUE #993: User comment doesn't appear after submitting it on a task
After submitting a comment on the test task, a box appears below the previous comment stating that I've made a comment, but doesn't show the actual comment. 

Tried making a comment on two separate test accounts, both have boxes that say the user has comment, but the actual comments don't appear on the task.

When I clicked "edit" then saved the comment, the actual comment showed up, but after hitting refresh, the comment disappeared. 

screen shot 2017-01-26 at 2 07 39 pm

begedin:
# What's in this PR?

In progress. The issue seems to be fixed, but tests are broken, so will need to rewrite them.

## References
Fixes #993

### DIFF
--- a/config/coverage.js
+++ b/config/coverage.js
@@ -1,3 +1,4 @@
 module.exports = {
-  parallel: true
+  parallel: true,
+  useBabelInstrumenter: true
 }
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -1,7 +1,7 @@
 module.exports = {
   root: true,
   parserOptions: {
-    ecmaVersion: 6,
+    ecmaVersion: 8,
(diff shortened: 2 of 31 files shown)


begedin:
Needs review, but blocked until we merge #990

config/coverage.js
```diff
@@ -1,3 +1,4 @@
 module.exports = {
-  parallel: true
+  parallel: true,
+  useBabelInstrumenter: true
```
begedin: Needed due to using ESNext features.

begedin:
@JoshSmith Should be good for another review now.

PR #998, opened 2017-01-30
Added prompt on navigating away from task creation

### LINKED ISSUE #991: Clicking out of new task form without submitting leaves new task
In the task manager, I clicked on the "new task" button. It brought me to the new task creation form. I clicked on the tasks tab in the secondary navigation to go back to the task UI. When I got back, there was an empty task (marked by a #) in the inbox. When I refreshed the task disappeared. I repeated the process  to make test the issue and got the same results.

screen shot 2017-01-26 at 1 35 38 pm

begedin:
# What's in this PR?

This PR adds a basic confirmation prompt when navigating away from the task creation page.
- destroys new record if confirmed
- aborts transition if not confirmed

## References
Fixes #991

### DIFF
--- a/app/routes/project/tasks/new.js
+++ b/app/routes/project/tasks/new.js
@@ -52,6 +52,20 @@ export default Route.extend(AuthenticatedRouteMixin, {
           this.controllerFor('project.tasks.new').set('error', error);
         }
       });
+    },
+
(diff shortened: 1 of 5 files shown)


JoshSmith:
@begedin just rebase and then we're good to merge.

PR #1002, opened 2017-02-01
Add cloudinary uploads

JoshSmith:
# What's in this PR?

This makes some significant changes to the way images are uploaded by allowing us to upload to Cloudinary, and from there to our application.

Work in progress. Needs updated tests. Also needs the image-drop in the onboarding test changed. May be room for simplification.

### DIFF
--- a/app/components/organization-settings-form.js
+++ b/app/components/organization-settings-form.js
@@ -3,13 +3,33 @@ import Ember from 'ember';
 const {
   Component,
   get,
-  inject: { service }
+  inject: { service },
+  set
 } = Ember;
 
 export default Component.extend({
   classNames: ['organization-settings-form'],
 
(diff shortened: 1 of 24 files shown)


app/components/organization-settings-form.js
```diff
@@ -3,19 +3,48 @@ import Ember from 'ember';
...
 
   flashMessages: service(),
+  loadingBar: service(),
+
+  uploadDone(cloudinaryPublicId) {
+    let organization = get(this, 'organization');
+    set(organization, 'cloudinaryPublicId', cloudinaryPublicId);
+    organization.save().then(() => {
+      this._stopLoadingBar();
+      console.log(organization);
```
begedin: Left a `console.log`

PR #1007, opened 2017-02-01
Improve forms, add biography, fix some CSS issues

### LINKED ISSUE #1001: Add user bio to user profile
Though you can input text for a bio in profile settings, it doesn't appear on the profile page after submission.

We should place the bio below the user's name and username and above the other elements (Twitter, website, etc).

JoshSmith:
# What's in this PR?

- Improve layout of forms some
- Add bio to user profile
- Minor CSS fixes

Fixes #1001

### DIFF
--- a/app/templates/components/user-sidebar.hbs
+++ b/app/templates/components/user-sidebar.hbs
@@ -1,4 +1,5 @@
 <img src={{user.photoLargeUrl}} />
+
 <h2>
   {{#if user.name}}
     <div class="name">{{user.name}}</div>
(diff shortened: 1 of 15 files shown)


app/templates/components/user-sidebar.hbs
```diff
@@ -7,6 +7,11 @@
   {{/if}}
   <div class="username">{{user.username}}</div>
 </h2>
+{{#if user.biography}}
+<p class="biography">
+  {{user.biography}}
+</p>
+{{/if}}
```
begedin: Requested change was here. Dunno how the comment failed (probably my mistake). I was saying we usually indent something like this.

PR #1011, opened 2017-02-03
Refactor user skills service

JoshSmith:
Attempt to refactor user skills service to make it generalizable to any type of skill relationship.


app/utils/skills-list.js
```diff
@@ -0,0 +1,49 @@
...
+    }
+  }
+}
+
+function _matchForModel(model, item, skill, relationship) {
+  switch(model) {
+    case 'user-skill':
+      return _matchWithRelationship(item, skill, relationship, 'user');
+    default:
+      return undefined;
```
JoshSmith: Not sure what to return for `default` here.

app/utils/skills-list.js
```diff
@@ -0,0 +1,49 @@
...
+    } else {
+      false;
+    }
+  }
+}
+
+function _matchForModel(model, item, skill, relationship) {
+  switch(model) {
+    case 'user-skill':
+      return _matchWithRelationship(item, skill, relationship, 'user');
```
JoshSmith: This `switch` was best way I could think of. Would appreciate other suggestions.

app/utils/skills-list.js
```diff
@@ -0,0 +1,49 @@
+import Ember from 'ember';
+
+const {
+  get
+} = Ember;
+
+let skillsList = {
```
JoshSmith: Not sure of a better convention here for this.

app/components/skill-list-item.js
```diff
@@ -12,15 +13,13 @@ export default Component.extend({
   tagName: 'li',
 
   matched: notEmpty('userSkill'),
-  userSkills: service(),
+  userSkillsList: service(),
```
npendery: should we change this service to just skills list?

npendery: Did you look into injecting the type of skills list here?

npendery: Did you look into seeing if we can inject the type of skills list this is?

JoshSmith: As you said earlier, I think we'll loop back to this once we have to expand out to other cases.

app/services/user-skills-list.js
```diff
@@ -0,0 +1,55 @@
...
+  userSkills: alias('user.userSkills'),
+
+  add(skill) {
+    let { store, user } = getProperties(this, 'store', 'user');
+    return store.createRecord('user-skill', { user, skill }).save();
+  },
+
+  contains(skill) {
+    let userSkills = get(this, 'userSkills');
+    return skillsList.contains(userSkills, skill);
```
WenInCode: Has [contains been deprecated](http://emberjs.com/deprecations/v2.x/#toc_enumerable-contains) in favour for `Enumerable#includes`?

app/utils/skills-list.js
WenInCode: I love that this is split out into a regular module, rather than a mixin/service.

As for the conventions, did you mean you are not sure if you should be exporting just an object literal? For this use case, where you are exporting 2 functions, I don't really see a problem with it. 

you *could* export them both as named exports and let things that are importing it choose what they import. Or even make the default export be an object literal of the two functions still.

app/utils/skills-list.js
```diff
@@ -0,0 +1,48 @@
...
+    } else {
+      false;
+    }
+  }
+};
+
+function _matchForModel(model, item, skill, relationship) {
+  switch (model) {
+    case 'user-skill':
+      return _matchWithRelationship(item, skill, relationship, 'user');
```
WenInCode: @JoshSmith I am not sure what the plan is down the line, or if the other model types would leverage the same `_matchWithRelationship` function. If they did you could do a mapping object for relationship name. Otherwise, this seems like a valid use-case for a `switch` statement.

app/utils/skills-list.js
```diff
@@ -0,0 +1,48 @@
...
+
+function _matchWithRelationship(item, skill, relationship, relationshipName) {
+  let itemRelationshipId = item.belongsTo(relationshipName).id();
+  let itemSkillId = item.belongsTo('skill').id();
+  let relationshipId = get(relationship, 'id');
+  let skillId = get(skill, 'id');
+  return (itemRelationshipId === relationshipId) && (itemSkillId === skillId);
+}
+
+export default skillsList;
```
WenInCode: Is there any chance we can add some documentation around this module? I'm a big fan of documentation in util modules.

app/services/user-skills-list.js
JoshSmith: Oh, it has. This `contains` is kind of a custom thing. It's really doing array-like operations on something that's not an array. Maybe that's confusing?

app/utils/skills-list.js
JoshSmith: Sure. May want to add some unit tests for this, too.

app/services/user-skills-list.js
WenInCode: ah yes - I definitely read this as if it was `Enumerable#contains` at first. But that makes sense now.

PR #1014, opened 2017-02-04
Add skills to projects

JoshSmith:
WIP PR to:

- Add skills to project from project settings
- Add UI for adding/removing skills from anywhere in app that skills are user-facing
- Add skill lists to project sidebar


app/services/project-skills-list.js
```diff
@@ -0,0 +1,58 @@
...
+    alias
+  },
+  get,
+  getProperties,
+  inject: { service },
+  Service,
+  set
+} = Ember;
+
+export default Service.extend({
```
WenInCode: This seems to be almost the same as the `user-skills-list` service. It might be worth it to either create a mixin for these services or perhaps a base service that they extend.

tests/pages/project/settings/profile.js
```diff
@@ -1,17 +1,30 @@
...
   visitable
 } from 'ember-cli-page-object';
 import projectSettingsForm from '../../components/project-settings-form';
+import skillsInput from '../../components/skills-input';
 
 export default create({
   visit: visitable(':organization/:project/settings/profile'),
 
+  projectSkillsList: collection({
+    scope: '.project-skills-list',
```
WenInCode: This same comment would apply to the other page objects that were created, but I don't want to pollute the review with several of the same things.

app/components/skill-list-item-link.js
```diff
@@ -0,0 +1,44 @@
...
+
+export default Component.extend({
+  classNames: 'skill-list-item-link',
+  classNameBindings: ['justClicked', 'justRemoved', 'matched'],
+  tagName: 'a',
+
+  session: service(),
+
+  click() {
+    if (get(this, 'session.isAuthenticated')) {
```
WenInCode: Would this ever happen? I think you are only rendering this component if the session is authenticated as it is?

app/services/project-skills-list.js
JoshSmith: I agree. If you'd like to create an issue, this PR is large enough as it is. There are a good number of cases here to use a mixin, and this will also be the case for the skill input when it comes time to use the keyboard events for things going forward.

app/components/skill-list-item-link.js
JoshSmith: No, we render this when unauthed, too.

JoshSmith: Eventually this won't be necessary, when I make the changes we talked about earlier. But for now it sadly is.

WenInCode:
@JoshSmith I created an issue to tackle the mixin. Could you please add documentation to any _new_ components added here (skills-list-item-link). Also see [this comment](https://github.com/code-corps/code-corps-ember/pull/1014#discussion_r99479427).

PR #1025, opened 2017-02-07
Refactor skills-list

JoshSmith:
# What's in this PR?

- Refactor `skills-list` into a `records-list` instead.
- Renames `skills-input` to `skills-typeahead`
- Renames `skills-input-item` to `skills-typeahead-result`

Not happy about looking into `content` here. If anyone has suggestions, I am all ears.


app/utils/records-list.js
```diff
@@ -0,0 +1,70 @@
...
+    @param {DS.Model} target The target model
+    @return {Boolean} `true` if found, otherwise `false`
+    @public
+  */
+  contains(records, target) {
+    if (records) {
+      return records.any((found) => {
+        let targetId = get(target, 'id');
+        let targetModelName = get(target, 'constructor.modelName')
+                              || get(target, 'content.constructor.modelName');
```
begedin: Wouldn't we usually align the `||` with the `=`?

app/utils/records-list.js
```diff
@@ -0,0 +1,70 @@
...
+    @return {DS.Model} Found item or `undefined`
+    @public
+  */
+  find(records, target, relationship) {
+    if (records) {
+      return records.find((found) => {
+        // Get relationship id and model name
+        let relationshipId = get(relationship, 'id');
+        let relationshipModelName = get(relationship, 'constructor.modelName')
+                                    || get(relationship, 'content.constructor.modelName');
```
begedin: Same alignment question here.

app/utils/records-list.js
```diff
@@ -0,0 +1,70 @@
...
+          return false;
+        }
+
+        // Get the id of the found model instance and
+        // get the id of the found model's relationship model instance
+        let foundId = found.belongsTo(targetModelName).id();
+        let foundRelationshipId = found.belongsTo(relationshipModelName).id();
+
+        return (foundRelationshipId === relationshipId) // relationships match
+            && (foundId             === targetId);      // found matches target
```
begedin: Similar question. We do not usually align things this way. I do not personally have a problem with it, but it is consistent with how we'd usually do it.

Issue #957, opened 2017-01-07
Last 4 input allows too many characters


rileytaylor:
# Problem

It should only allow 4 characters

M0nica:
@rileytaylor this issue should be resolved. I just merged a PR. https://github.com/code-corps/code-corps-ember/pull/1026

PR #1026, opened 2017-02-09
#957 - Last 4 input allows too many characters

M0nica:
# Data Validation for ssn-last-4
Set max length for input for ssn-last-4 to 4 chars. In order to prevent users from entering more than 4 characters. 

## References
Progress on: #957 

cc: @rileytaylor, I wasn't able to load the screen to create an organization based on how the local site is setup. Were you able to see this validation error locally or only in production? I would love any guidance as to how to view the user-facing page associated with creating accounts. I tried logging in as `admin` and `owner` with the mirage data but still was not able to access that page.

### DIFF
--- a/app/templates/components/payments/funds-recipient/details-form.hbs
+++ b/app/templates/components/payments/funds-recipient/details-form.hbs
@@ -99,7 +99,7 @@
   <div class="input-section__label">SSN Last 4</div>
   <div class="input-section__content">
     <div class="input-group">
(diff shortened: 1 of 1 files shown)


M0nica:
Update: I was able to navigate to the form and confirm that it only allows me to enter a max of 4 chars. http://localhost:4200/code-corps/code-corps/settings/donations/payments

PR #1035, opened 2017-02-10
Redo CSS for headers and add breakpoints

JoshSmith:
# What's in this PR?

Changes the header, adds breakpoints, makes things generally more responsive.


JoshSmith:
Does things like:



and:

PR #1037, opened 2017-02-11
Fix issues with Safari

JoshSmith:
# What's in this PR?

Safari did not work well with flex box. I think this may fix some of the issues here, but the spacer no longer works as expected.

Could use a second set of eyes on that.


JoshSmith:
Closing in favor of #1038.

PR #1039, opened 2017-02-11
Fix safari project card responsive issue

appleJax:
What's in this PR?

Fixes flexbox issue with project cards on Safari

Changes:

Add flexbox container around nav-bar
Make "header" display: block to prevent project cards from overlapping
References #1037


JoshSmith:
🙌

Issue #1043, opened 2017-02-13
Rewrite skill button so only authenticated users can trigger the remove action


begedin:
# Problem

Currently, with the skill button, there is no check to see if the user hovering/clicking it can remove the skill from whichever related parent.

We could technically deal with this at a higher level, route for example, but due to the hover effect also being applied, It would be very complicated to do it that way. We need to either use the session service directly, or pass in a `canRemove` flag into the component. The later is preferable, since it provides more flexibility in what that flag is actually determined from

We do not need this for project skills, since the place where that component is rendered for project skills is a route only accessible by owners.

For task skills, however, any contributor should be able to toggle skills.

## Subtasks

- [ ] Pass in `canRemove` flag into `skill-button`
- [ ] Hover effect should work depending on the `canRemove` flag
- [ ] `remove` action should work depending on the `canRemove` flag

JoshSmith:
I think this actually needs rewritten to *not* show the skill buttons unless in an editing state, and otherwise show a `skills-list`.

begedin:
@JoshSmith I missed your comment. I absolutely agree. It does sound like a better approach.

JoshSmith:
I think this deserves some more discussion and clarification still, since it seems like we've developed several competing UIs, and I'd like to do some work to clear some of those up and simplify them.

This really needs design-level clarification right now.

Issue #1042, opened 2017-02-13
Add skills to tasks


begedin:
# Problem

We want users to be able to add and remove skills to/from tasks

## Subtasks

- [x] Assigned skills should be visible on task card - #1045
- [x] Add/remove UI should be present on the task page - #1045
- [x] Add/remove UI should be present on the task new page? - #1046
- [ ] Add/remove UI should only be usable by project contributors or higher - #1043 

I think the last one, considering how much harder it is to deal with, might be OK to punt for a later time. I'm not really sure, but I think it's worth discussing.

PR #1045, opened 2017-02-13
Add skills to tasks

begedin:
# What's in this PR?

Adds a task skill UI to the task sidebar on the task page. Allows assigning of task skills via a search input as well as removing assinged task skills.

Also adds a list of skills on the task card. For this purpose, I rewrote `project-card-skills` into `related-skills`, since behavior is near-identical and only styling is slightly different.

## TODO: 

- Styles. Can be done here or separately. Currently, the design issue (code-corps/design#5) has a broken PDF link, so I could not use it as reference. Once I have the pdf, I can add the styles here, or someone else can do it here, or it can be done as a separate PR - #1047
- Currently, anyone can assign/unassign skills. We should change it so only contributors or higher can do it - #1043 
- Do we want to add a skill UI to the new task page? This one is going to be trickier, but it's still doable. - #1046

## References
Progress on #1042
Requires code-corps/code-corps-api#696


begedin:
@JoshSmith I went ahead and wrote a bunch of issues to split the TODO's into that. I think this is in a mergeable state as is, so we can focus on smaller issues/PRs individually (and in parallel)

PR #1049, opened 2017-02-13
Update designs of project page to be responsive

JoshSmith:
# What's in this PR?

Brings in some changes to the project and the sidebar.

### DIFF
--- a/app/styles/components/organization-members.scss
+++ b/app/styles/components/organization-members.scss
@@ -1,3 +1,16 @@
+@mixin member-list-item($number, $margin-right) {
+  $percentage: 100 / $number;
+  $total-margins: $margin-right * $number;
+  $width: calc(#{$percentage}% - #{$total-margins / $number});
+
+  margin-right: $margin-right;
+  width: $width;
+
+  &nth-child(#{$number}n) {
+    margin-right: 0;
+  }
+}
+
 .organization-members {
   h3 {
     display: flex;
@@ -16,16 +29,21 @@
   }
 
   li {
-    width: calc(20% - 16px/5);
-    margin-right: 4px;
-
-    &:nth-child(5n) {
-      margin-right: 0;
-    }
+    $number: 6;
+    $margin-right: 3px;
+    @include member-list-item(6, 3px);
 
     .icon {
       height: auto;
       width: 100%;
+    }
+
+    @include media($md-screen) {
+      @include member-list-item(9, 3px);
+    }
+
+    @include media($sm-screen) {
+      @include member-list-item(6, 3px);
     }
   }
 }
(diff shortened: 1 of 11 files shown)


app/styles/components/organization-members.scss
```diff
@@ -1,3 +1,16 @@
+@mixin member-list-item($number, $margin-right) {
+  $percentage: 100 / $number;
+  $total-margins: $margin-right * $number;
+  $width: calc(#{$percentage}% - #{$total-margins / $number});
+
+  margin-right: $margin-right;
+  width: $width;
+
+  &nth-child(#{$number}n) {
```
npendery: Is the last `n` here the placeholder that was supposed to be taken out?

app/styles/templates/project/index.scss
```diff
@@ -5,10 +5,23 @@
 
 .project-main {
   @include span-columns(8);
+  @include media($lg-screen) {
```
npendery: Just for my knowledge, where does the `media($lg-screen)` come into play? I see from the `_breakpoints.scss` its defining sizes but how does it recognize/render this?

app/styles/components/project-long-description.scss
```diff
@@ -1,10 +1,4 @@
-.project-long-description {
```
npendery: Also remove the class in the component js file?

app/templates/project/index.hbs
```diff
@@ -9,13 +9,14 @@
 <div class="container">
   <div class="project-wrapper">
 
-    {{project-long-description project=project}}
+    <div class="project-main">
+      {{project-long-description project=project}}
```
npendery: Include this tag and class in the component js?

app/styles/components/organization-members.scss
JoshSmith: No this is the `_x_n`-th item we want that to happen on. That means it repeats.

app/styles/templates/project/index.scss
JoshSmith: Just normal media breakpoints. Browser behavior comes into play here.

app/styles/components/project-long-description.scss
JoshSmith: I didn't remove that because I like to keep the component 1-1 with classes, even if it doesn't have styling.

app/templates/project/index.hbs
JoshSmith: There may be more things that go into the main area.

JoshSmith: See the sidebar below to see what I mean.

app/styles/components/organization-members.scss
npendery: Ahh gotcha, makes sense

Issue #1052, opened 2017-02-14
Add project skills to the skill assignment UI on the new task page


begedin:
# Problem

When creating a task, in the skill assigment UI, there should be a section listing all of the project's skills, to allow the user to quickly add them without using the search field.

![image](https://cloud.githubusercontent.com/assets/1925540/22924226/9eb2d37e-f2a4-11e6-8892-c8730f9b3ba9.png)

On the screenshot, it's the part separated by "or click to add skills below"

Issue #1046, opened 2017-02-13
Add task skill UI to the new task page


begedin:
# Problem

Ideally, we want to be able to assign skills to a task while creating the task, on the new task page or within the new task form.

This is trickier then assigning/unassigning to an existing task, but should be doable

## Proposed solution

- [ ] new task form component should also hold a skills-typeahead, as well as a skill list
- there is also a list of project skills, which we can use to add skills listed on the project quickly with a single click - this can be done separately in #1052 
- [ ] instead of the classical approach of task-skills being created and destroyed when using this UI, we instead store/remove them to/from an array held by the component
- [ ] on hitting the form submit button, the save action also passed out a list of skills, which are then created as task skills alongside the task itself, after the task was created

## Optional improvement

- Once the implementation above is done, we can look into optimizing it (probably initially not necessary) by modifying our task create request to also embed task skills. This would require API changes as well.

## UI

Sidebar of the new task page:

![image](https://cloud.githubusercontent.com/assets/1925540/22921399/c292ecd0-f299-11e6-9d21-8b60c2a88cba.png)

## References

Progress on: #1042

JoshSmith:
@begedin is this good to close now?

begedin:
This part can be closed. The part in #1052 is not done yet.

PR #1054, opened 2017-02-14
Add skill ui to new task page

begedin:
# What's in this PR?

This PR adds basic skill assignment UI to the new task page. It works basically like the UI on the edit page, except the `task-skill` records are actually created after the task is saved.

Styling is not final, but can be dealt with in another PR.

The section listing the project skills for quick adding is also not there. It can be added with #1052 

This should be reviewed and merged **after** reviewing and merging #1045 

I also made a code change in the x-skill service layer. I renamed `.contains` to `.includes`, reason being `.contains` is being deprecated on ember arrays. Since for the task new page, we use a plain ember array instead of a service, it would make sense for the interface to match.

## References
Progress on: #1046

### DIFF
--- a/app/controllers/project/tasks/new.js
+++ b/app/controllers/project/tasks/new.js
@@ -4,21 +4,23 @@ import { isNonValidationError } from 'code-corps-ember/utils/error-utils';
 const {
   Controller,
   get,
+  RSVP,
   set
 } = Ember;
 
 export default Controller.extend({
+  unsavedTaskSkills: [],
   actions: {
     /**
-     * saveTask - action
-     *
-     * Triggered when user clicks 'save' on the new task form.
-     * Adds task to the correct task list, which is the inbox task list.
-     * Saves task and, on success, transitions to the task route.
-     * On failure, calls the save error handler.
-     *
-     * @param  {DS.Model} task The task to be saved.
-     */
+     saveTask - action
+
+     Triggered when user clicks 'save' on the new task form.
+     Adds task to the correct task list, which is the inbox task list.
+     Saves task and, on success, transitions to the task route.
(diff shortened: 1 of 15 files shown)


app/controllers/project/tasks/new.js
```diff
@@ -26,8 +28,22 @@ export default Controller.extend({
       set(task, 'taskList', inboxTaskList);
 
       return task.save()
+                 .then((task) => this._saveSkills(task))
                  .then((task) => this.transitionToRoute('project.tasks.task', get(task, 'number')))
                  .catch((payload) => this._handleTaskSaveError(payload));
+    },
+    deselectSkill(skill) {
```
JoshSmith: Spacing here.

app/controllers/project/tasks/new.js
```diff
@@ -58,5 +74,16 @@ export default Controller.extend({
     if (isNonValidationError(payload)) {
       set(this, 'error', payload);
     }
+  },
+
+  async _saveSkills(task) {
+    let unsavedTaskSkills = get(this, 'unsavedTaskSkills');
+    let promises = unsavedTaskSkills.map((skill) => this._createTaskSkill(skill, task));
+    return RSVP.all(promises).then(() => task);
+  },
```
JoshSmith: Maybe this below `_create...` to keep alpha?

app/services/user-skills-list.js
```diff
@@ -24,9 +24,9 @@ export default Service.extend({
     return store.createRecord('user-skill', { user, skill }).save();
   },
 
-  contains(skill) {
+  includes(skill) {
```
JoshSmith: Like renaming to `includes` here.

PR #1058, opened 2017-02-15
Make signup page responsive

appleJax:
# What's in this PR?
Makes signup page responsive.

Changes for smaller screens:
- widen inputs
- error messages appear below inputs

signup

### DIFF
--- a/app/styles/components/signup-form.scss
+++ b/app/styles/components/signup-form.scss
@@ -5,4 +5,20 @@
   input[type="submit"] {
     width: 100%;
   }
+  @include media($lg-screen) {
+    @include shift(2);
+    @include span-columns(8);
+  }
+  @include media($sm-screen) {
+    @include shift(0);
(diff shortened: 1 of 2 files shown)


app/styles/components/signup-form.scss
```diff
@@ -5,4 +5,25 @@
...
+  }
+  @include media($sm-screen) {
+    @include shift(1);
+    @include span-columns(10);
+  }
+  @include media($sm-screen) {
+    @include shift(0);
+    @include span-columns(12);
+    padding: 0 8px;
+  }
```
JoshSmith: Looks like there is some repetition on the `sm-screen` here.

Also, I see there being some padding on the last `sm-screen`. If we wrap the login in a `.container`, then it should be fine.

appleJax:
@JoshSmith good catch. I removed the duplicate code and wrapped the form in a .container

PR #1061, opened 2017-02-17
Responsive hello page

appleJax:
# What's in this PR?

Makes start/hello page responsive

Changes:
This one took an embarrassingly long time to do. I finally hacked together a couple different solutions. 

Solution (aka commit) # 1:
I tried to make it work entirely with css. The right-hand image made this particularly difficult because it is the background-image to the same container that holds the rest of the page content. The end result is pretty hideous code-wise, with negative margins and a bunch of pixel-specific juggling of margin and padding. I abandoned this solution before it was 100% complete, so if we decide to go this route, I will have some more work to do.

Solution (aka commit) # 2:
I gave in and rearranged the html markup to make the responsive transitions a little less hacky. The XL-width version of the page is not exactly the same as before, but it's ~98% of the way there. I could not keep the right-hand image from getting cut off at certain screen sizes without making a slight compromise that pushes the image a little to the left  of its original position on XL-width screens. Overall, I'm pretty happy with how the page flows across the range of different screen sizes, but I am open to suggestions of ways to improve the code.

screen shot 2017-02-16 at 6 32 42 pm
screen shot 2017-02-16 at 6 34 00 pm

### DIFF
--- a/app/styles/layout/_header.scss
+++ b/app/styles/layout/_header.scss
@@ -90,9 +90,13 @@
 
 .onboarding__progress {
   padding: 23px 0 22px 0;
-  width: 180px;
+  width: 100px;
+
+  .progress-bar-container {
+    height: 7px;
+  }
 }
 
 .onboarding__steps {
-  padding: 0 30px;
+  padding: 16px 30px;
 }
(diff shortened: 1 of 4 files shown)


JoshSmith:
This looks fantastic and I definitely get why it took so long. Thanks for all the hard work on it.

PR #1066, opened 2017-02-18
Remove task icons and types and fix user dropdown

JoshSmith:
# What's in this PR?

Removes task types and icons.

Also does some fixes for the task user select dropdown.


app/components/skills-textfield.js
```diff
@@ -0,0 +1,28 @@
...
+      case 'Comma':
+      case 'Enter':
+        e.preventDefault();
+        break;
+      default:
+        this._super(...arguments);
+    }
+    get(this, 'getKeyDown')(key);
+  }
+});
```
begedin: Would this one benefit from extending `select-textfield`? Not something I would force into this PR, but might at least be worth a "discussion" task.

app/components/task-card.js
```diff
@@ -31,21 +43,134 @@ export default Component.extend({
...
+    get(this, 'selectRemoteController').actions.open();
+    this._unbindKeydown();
+  },
+
+  willDestroyElement() {
+    this._unbindKeydown();
+  },
+
+  actions: {
+    async searchUsers(query) {
```
begedin: This is marked as `async`, but we do not call await anywhere in there.

app/components/task-card.js
```diff
@@ -31,21 +43,134 @@ export default Component.extend({
...
+      let store = get(this, 'store');
+      store.query('user', { query }).then((users) => {
+        set(this, 'results', users);
+      });
```
begedin: To elaborate on the previous comment, if the function is`async`, then we could rewrite this as 

```Javascript
let users= await store.query('user', { query });
set(this, 'results', users);
```

If we don't want to write it that way, then marking the function `async` would be misleading as well, so I would remove it. @

app/templates/components/task-new-form.hbs
```diff
@@ -1,21 +1,5 @@
...
-        <option value="issue" selected={{eq task.taskType 'issue'}}>Issue</option>
-      {{/if}}
-      {{#if (can "create idea task in organization" organization)}}
-        <option value="idea" selected={{eq task.taskType 'idea'}}>Idea</option>
-      {{/if}}
-    </select>
-    {{#each task.errors.taskType as |error|}}
-      <p class="error">{{error.message}}</p>
-    {{/each}}
-  </div>
```
begedin: I love that this part is going away.

app/templates/components/task-card.hbs
```diff
@@ -7,15 +7,47 @@
...
+    <div class="select-inline-dropdown__list-item__content">
+      <span class="select-inline-dropdown__list-item__username">{{user.username}}</span>
+      {{#if user.name}}
+        <span class="select-inline-dropdown__list-item__name">{{user.name}}</span>
+      {{else}}
+        <span class="select-inline-dropdown__list-item__name">{{user.username}}</span>
+      {{/if}}
+    </div>
+  {{/power-select}}
+  {{!-- {{task/user/user-select
```
begedin: Any particular reason why this is commented out but not removed? I see a few related lines of component code that could get removed to in that case.

app/components/task-card.js
```diff
@@ -1,51 +1,108 @@
...
+      return get(user, 'id') === get(item, 'id');
+    });
   }),
+
   taskSkills: mapBy('task.taskSkills', 'skill'),
+  taskUserId: alias('taskUser.id'),
+  users: alias('members'),
+
+  userOptions: computed('currentUser.user', 'taskUserId', 'users', function() {
+    let { currentUserId, taskUserId, users } = getProperties(this, 'currentUserId', 'taskUserId', 'users');
```
begedin: Any specific reason why the computed depends on `currentUser.user`. but then uses `currentUserId`? Does it not work if it depends on `currentUserId`?

app/components/task-card.js
```diff
@@ -1,51 +1,108 @@
...
+  click(e) {
+    // TODO: Find a better way to do this
+    // Currently necessary due to the way that power select handles trigger
+    let iconClassName = 'ember-power-select-status-icon';
+    let triggerClassName = 'ember-power-select-trigger';
+    if (e.target.className.includes(iconClassName) || e.target.className.includes(triggerClassName)) {
+      return;
+    }
     let isLoading = get(this, 'isLoading');
     let task = get(this, 'task');
```
begedin: `getProperties` would probably be nicer here.

app/components/task-new-form.js
```diff
@@ -2,14 +2,12 @@ import Ember from 'ember';
...
-  computed,
-  get,
   inject: { service }
 } = Ember;
 
 /**
-  The task-new-form component is used for creating new tasks. It includes the
-  task type, title, editor and preview
+  The task-new-form component is used for creating new tasks. It includes
+  the task type, title, editor and preview
```
begedin: task type is no longer here

app/components/task-card.js
JoshSmith: I believe it had not, and not sure why.

PR #1068, opened 2017-02-20
Remove junk code related to select-option

begedin:
# What's in this PR?

Best I an tell, the code removed here is completely unused. The tests pass, and I'm not seing anything behaving differently. I guess at some point, we had a custom select component of our own, which we removed.

Furthermore, the component is defined in `app/components/select-option.js`, while the template
is in `app/templates/select-option.hbs`, when it should be in `app/templates/components/select-option.hbs`, so it was broken to begin with.


begedin:
I'm not sure why coveralls is no longer reporting the change in coverage, but in this specific case, it changed from 83.74 to 84.01

PR #1069, opened 2017-02-20
Add tests, switch tests to page objects #1

begedin:
# What's in this PR?

Just a PR that adds some test coverage and moves a couple of existing test suites towards using page objects.

## Changes

* Added a unit test for `power-select/before-options`. There is really no user interaction to be tested here, so a unit test is preferable.

### DIFF
--- a/tests/unit/components/power-select/before-options-test.js
+++ b/tests/unit/components/power-select/before-options-test.js
@@ -1,0 +1,28 @@
+import { moduleForComponent, test } from 'ember-qunit';
+import Ember from 'ember';
+
+const { Controller, set } = Ember;
+
+moduleForComponent(
+  'power-select/before-options',
+  'Unit | Component | power select before options',
+  {
+    unit: true
+  }
+);
+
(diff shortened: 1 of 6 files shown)


tests/unit/components/power-select/before-options-test.js
```diff
@@ -0,0 +1,28 @@
...
+const { Controller, set } = Ember;
+
+moduleForComponent(
+  'power-select/before-options',
+  'Unit | Component | power select before options',
+  {
+    unit: true
+  }
+);
+
```
JoshSmith: What is a unit component test vs an integration? Not sure I've seen this distinction.

begedin: An integration test will partially boot the app (albeit not fully, like an acceptance test). It is also expected to render the component, which will have access to (and render) all subcomponents/services/etc.

The intention of an integration test it so check how the component is working with connected parts (while not fully starting the app) and how it interacts with the user/other layers.

The purpose of a component unit test would be to simply execute the component code to test, for example, some complex function within the component.

PR #1071, opened 2017-02-20
Fix task-card unassignment

JoshSmith:
# What's in this PR?

Uses the as-yet undocumented `buildSelection` in `ember-power-select` to deselect a user by passing `null` in order to unassign.


venkatd:
👍 looks good to me

PR #1077, opened 2017-02-21
Task assignment tests, page objects, bug fixes

### LINKED ISSUE #1076: Sentry: "TypeError: Cannot read property 'id' of null" on task list
# Problem

https://sentry.io/code-corps/code-corps-ember/issues/222295709/

begedin:
# What's in this PR?

This deals with the sentry issue we've been having and adds test to cover user task assignment in integration.

I'm not 100% positive this will fix the issue, but it should be the case, since I've been able to reproduce it on ocasion in development, but I'm no longer able to after this change.

The best I can figure out was, to populate the assignment dropdown, we were doing a `users.mapBy('content')`. However, in some cases, the `content` property of the `DS.PromiseObject` was `null` (not loaded yet). This happens rarely in development, more often in staging, but almost always on production.

I was implementing integration tests on the task card component and, while trying to implement the power select page object, managed to write a failing test for this without realizing. 

The fix for it was relatively simple - we do not map by content. @JoshSmith do you remember why you were mapping by content in the first place? If a user is a `DS.PromiseObject` then binding `user.username` will render nothing while the user is being loaded, but should switch to rendering `user.content.username` as soon as the loading promise resolves. This is the behaviour I'm observing locally, so there should be no need to map by content at all. Is it not the case on your side, or am I missing something.

Once the fix was applied and the failing case was passing, I wrote more tests to cover cases where we have promises and already loaded objects.

I guess, if we want nicer UX, we could have the list item component also react to `user.isLoading` and show a different state while that is happening. I believe power-select actually supports something like that in some way.

## References
Fixes #1076

### DIFF
--- a/app/components/task-card.js
+++ b/app/components/task-card.js
@@ -70,8 +70,7 @@ export default Component.extend({
     let { currentUserId, taskUserId, users }
       = getProperties(this, 'currentUserId', 'taskUserId', 'users');
     if (users) {
-      // TODO: Replace content here!
-      return createTaskUserOptions(users.mapBy('content'), currentUserId, taskUserId);
+      return createTaskUserOptions(users, currentUserId, taskUserId);
     } else {
       return [];
     }
@@ -100,7 +99,7 @@ export default Component.extend({
       return option;
     },
 
-    async changeUser(user) {
+    changeUser(user) {
       let { task, taskAssignment } = getProperties(this, 'task', 'taskAssignment');
 
       if (user) {
--- a/tests/pages/components/power-select.js
+++ b/tests/pages/components/power-select.js
@@ -1,0 +1,36 @@
+import { collection } from 'ember-cli-page-object';
+import { clickTrigger, nativeMouseUp } from 'code-corps-ember/tests/helpers/ember-power-select';
+
+// NOTE: The default ember-cli-page-objects do not work properly here
+// Instead, ember-power-select exposes the `clickTrigger` and `nativeMouseUp`
+// helpers which we used to implement this page object.
+//
+// Especially of note is the usage of `nativeMouseUp`.
+// - `clickable`, `click`, or even `nativeMouseDown` do not work.
+// - there is no documentation for this either
+// - the `selectChoose` test helper can only be used in acceptance tests
+// - using `nativeMouseUp` up for this is undocumented, but I started using it after looking at the code at:
+//   https://github.com/cibernox/ember-power-select/blob/master/tests/integration/components/power-select/general-behaviour-test.js#L236
+
+// NOTE: Since the power select is appended to the application root by default,
+// We're also using the `testContainer` feature provided by ember-cli-page-object
+
+export default {
+  trigger: {
+    open: clickTrigger
+  },
+
+  dropdown: {
+    resetScope: true,
+    testContainer: '.ember-power-select-dropdown',
+    options: collection({
+      itemScope: '.ember-power-select-option',
+      item: {
+        select() {
+          nativeMouseUp(this.scope);
+          return this;
+        }
+      }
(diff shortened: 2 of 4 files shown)


begedin:
@JoshSmith After rebasing, I had some items to fix. However, the tests that were failing on my machine also included some `submittable-textarea` related tests, so I added another commit to resolve those. Correct me if I'm wrong, but it looks like there was a regression introduced in 50c7c34c25a6d6d157a841ca117bc0ba52ad0e9f. With that commit, the submittable textarea no longer supported the `ctrl+enter` combination; only `cmd+enter`. I'm not sure why the tests were still passing on CI, but locally for me, the `ctrl+enter` tests were failing, so I re-added and rewrote the page objects a bit so they are more-selfcontained.

app/components/task-card.js
```diff
@@ -70,8 +70,7 @@ export default Component.extend({
     let { currentUserId, taskUserId, users }
       = getProperties(this, 'currentUserId', 'taskUserId', 'users');
     if (users) {
-      // TODO: Replace content here!
-      return createTaskUserOptions(users.mapBy('content'), currentUserId, taskUserId);
+      return createTaskUserOptions(users, currentUserId, taskUserId);
```
begedin: This is the fix.

app/components/task-card.js
```diff
@@ -100,7 +99,7 @@ export default Component.extend({
       return option;
     },
 
-    async changeUser(user) {
+    changeUser(user) {
```
begedin: No reason for it to be declared `async`

tests/integration/components/editor-with-preview-test.js
```diff
@@ -182,7 +181,6 @@ test('it sends the modifiedSubmit action with ctrl+enter', function(assert) {
   });
   page.render(hbs`{{editor-with-preview modifiedSubmit="modifiedSubmit"}}`);
 
-  page.focus();
-  triggerKeyDown('Enter+cmd');
+  page.textarea.ctrlEnterSubmit();
```
begedin: If they go in pair, might as well make them a single page method.

tests/integration/components/submittable-textarea-test.js
```diff
@@ -1,43 +1,58 @@
 import { moduleForComponent, test } from 'ember-qunit';
 import hbs from 'htmlbars-inline-precompile';
-import { triggerKeyDown, initialize } from 'ember-keyboard';
+import pageComponent from 'code-corps-ember/tests/pages/components/submittable-textarea';
+import PageObject from 'ember-cli-page-object';
```
begedin: Introduced page component.

tests/pages/components/editor-with-preview.js
```diff
@@ -28,20 +26,12 @@ export default {
     isActive: hasClass('active')
   },
 
-  focus: triggerable('focus', 'textarea'),
+  textarea: submittableTextarea,
```
begedin: Extracted the textarea into a page object of its own, since it already is a component of its own

tests/pages/components/power-select.js
```diff
@@ -0,0 +1,36 @@
...
+    options: collection({
+      itemScope: '.ember-power-select-option',
+      item: {
+        select() {
+          nativeMouseUp(this.scope);
+          return this;
+        }
+      }
+    })
+  }
```
begedin: The explanation is huge, but the implementation itself is tiny in comparison.

begedin:
Test still failing on CI for some reason. Trying to figure it out.

begedin:
@JoshSmith I've moved the commit for `submittable-textarea` into #1078 , to avoid blocking this from merge.

PR #1078, opened 2017-02-21
Fix issues with submittable-textarea, refactor/add page objects

begedin:
# What's in this PR?

Cherry-picked from #1077 

This fixes tests for editor-with-preview and submittable-textarea on my machine, but, apparently, breaks them on CI.

It also makes the page objects a bit nicer and introduces a submittable-textarea page object.

I think it's worth figuring out and fixing, but I did not want to block the merge of #1077 so I cherry picked into a separate PR.


JoshSmith:
Ah I spent a ton of time on this and you figured it out! 🎩 s off to you!

PR #1079, opened 2017-02-21
Introduce category-item page object

begedin:
# What's in this PR?

This PR introduces the `category-item` page object.

While implementing it, I also simplified the component slightly.

Before this, we were binding the `selected` and `category.slug` properties as css classes on multiple different elements within the component.

With a minor `.scss` rewrite, I was able to make it so we just need a `classsNameBindings` for both at the root level of the component.

I decided to leave the `getFlashMessageX` helpers as is, since they deal with testing the associated service, not the actual UI, so they don't really fit within the page object. Integration tests will not have the flash messages rendered, since the service does that at the application level, meaning we can't ditch them at integration test level.

### DIFF
--- a/app/components/category-item.js
+++ b/app/components/category-item.js
@@ -24,7 +24,6 @@ const {
  */
 export default Component.extend({
   classNames: ['category-item'],
-  classNameBindings: ['selected'],
   isLoading: false,
 
   /**
@@ -38,6 +37,21 @@ export default Component.extend({
    * @type Ember.Service
    */
   userCategories: service(),
+
+  /**
+   * Returns the class name for the icon.
+   *
+   * @property iconClass
+   * @type String
+   */
+  iconClass: computed('category.slug', 'selected', function() {
+    let slug = get(this, 'category.slug');
+    if (get(this, 'selected')) {
+      return `category-item__icon--${slug}--selected`;
+    } else {
+      return `category-item__icon--${slug}`;
+    }
+  }),
 
   /**
    * Returns if the category has been selected by the user.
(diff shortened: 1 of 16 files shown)


app/components/category-item.js
```diff
@@ -24,7 +24,7 @@ const {
  */
 export default Component.extend({
   classNames: ['category-item'],
-  classNameBindings: ['selected'],
+  classNameBindings: ['selected', 'category.slug'],
```
JoshSmith: I'm going to rework this some myself, but with BEM we would rather have it be something like `--selected` with an existing class rather than adding the `.selected` class onto this. The namespacing is really critical here to ensuring that our styles don't overlap. I know you're just using the existing class, but this should probably be in the refactor.

app/styles/components/category-item.scss
```diff
@@ -13,7 +13,7 @@
     font-size: $body-font-size-small;
     margin: 10px 0;
 
-    &.selected {
+    .selected & {
       color: $default-color;
```
begedin: @JoshSmith With the change I made, this css would apply to the description. Once you added your change, though, this is no longer the case, so this style block does nothing, as far as I can tell. 

This means that the `description` text is always using the `$text--light` (`#666`) color and does not switch to `$default-color` (`#08A8FC` when selected.

I'm not against your approach and we probably want BEM to apply. However, if we want the description color to change, then we want the class to bind to the component, not the component icon. Otherwise, we need to bind to the description as well, so we're back to square one.

If you're ok with this behavior as is (where the description stays `#666`, regardless of status, then we should at the very least remove this style block, since it does nothing.

JoshSmith: Whoops can you add an issue to fix that? I also realized that my overall BEM change is not really the direction I wanted to head in but was unwilling to spend more time on something not super critical.

PR #1091, opened 2017-02-22
1083 fix unassign issue

### LINKED ISSUE #1083: Look into splitting up ember-exam partitions across CI containers
# Problem

Our Ember tests (just the `ember exam` portion) take about 3 minutes to run on Circle, and we could hypothetically speed this up to be 2-4x faster by making use of splitting containers.

If someone can look into whether this is feasible or ideal, that would be fantastic.

### LINKED ISSUE #1090: Bug: Unassignment of tasks no longer works
# Problem

I'm not sure when it happened exactly, but as far as I can tell, tasks can no longer be unassigned.

There is no difference between rendering an assigned or unassigned user in the dropdown and clicking on an assigned user calls the same method to re-assign as it would clicking some other user entirely.

begedin:
# What's in this PR?

This is initial work on fixing #1090

I discovered the issue while adding acceptance tests to task assignment.

Turns out, I caused the bug by removing the `mapBy('content')` within the `userOptions` method. This caused the `ember-power-select-is-equal` helper to internally fail, since it was comparing a `DS.Model` with a `DS.PromiseObject`.

This likely breaks a couple of tests, but with the computed change introduced in #1082, the sentry bug where the task list page failed to render due to fetching on id on a null object should no longer be a problem.

That being said, with the feature enabled again, it does seem like it's a bit buggy and sluggish, so maybe we should consider using another approach altogether. For example, a separate option to clear assignment might work, or we could map the options into a simpler object.

## References
Fixes #1090

### DIFF
--- a/app/routes/project/tasks/index.js
+++ b/app/routes/project/tasks/index.js
@@ -3,14 +3,21 @@ import Ember from 'ember';
 const {
   get,
   inject: { service },
-  Route
+  Route,
+  RSVP
  } = Ember;
 
 export default Route.extend({
   projectTaskBoard: service(),
 
   model() {
-    return this.modelFor('project');
+    let project = this.modelFor('project');
+    let members = RSVP.all(get(project, 'organization.organizationMemberships').mapBy('member'));
+    return RSVP.hash({ project, members });
+  },
+
+  setupController(controller, models) {
+    controller.setProperties(models);
   },
 
   actions: {
--- a/app/services/task-assignment.js
+++ b/app/services/task-assignment.js
@@ -10,35 +10,70 @@ const {
 export default Service.extend({
   store: service(),
 
+  /**
+    Assign a task to a user.
+
+    If there's already a user task for the task passed in, it updates th
+    existing record. Otherwise, it createes a new record.
+
+    @method assign
+    @param {DS.Model} task
+    @param {DS.Model} user
+    @return {DS.Model} userTask
+  */
   async assign(task, user) {
     let userTask = await get(task, 'userTask');
+    // If there's already a userTask, update, otherwise create a new one
     return userTask ? this._update(userTask, user) : this._create(user, task);
   },
 
+  /**
+    Check if a task is assigned to a user.
+
+    @method isAssignedTo
+    @param {DS.Model} task
+    @param {DS.Model} user
+    @return {Boolean}
+  */
   async isAssignedTo(task, user) {
     return await get(task, 'userTask.user.id') === get(user, 'id');
   },
 
+  /**
+    Unassign a task.
+
+    For the given task, unassign it from whomever it's assigned to. This simply
+    destroys the userTask record for that task.
+
+    @method assign
+    @param {DS.Model} task
+    @return {DS.Model} userTask
+  */
   async unassign(task) {
     let userTask = await get(task, 'userTask');
 
     if (userTask) {
-      return this._destroy(userTask);
+      return userTask.destroyRecord();
     }
   },
 
   _create(user, task) {
     let store = get(this, 'store');
-    return store.createRecord('user-task', { user, task })
-                .save();
-  },
-
-  _destroy(userTask) {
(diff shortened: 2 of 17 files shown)


JoshSmith:
@begedin is this supposed to say 1083 or 1082 or something else?

app/services/task-assignment.js
```diff
@@ -29,16 +29,25 @@ export default Service.extend({
 
   _create(user, task) {
     let store = get(this, 'store');
-    return store.createRecord('user-task', { user, task })
-                .save();
+    let userTask = store.createRecord('user-task', { user, task });
+    return userTask.save().catch(() => {
+      userTask.deleteRecord();
```
runspired: if an error occurs while creating a record, why make an API call to delete it vs just unloading it?

JoshSmith:
@begedin I'm resolving the promise at the level of the route because I was told this evening that relying on async relationships in computeds seems like a code smell.

> Computeds based on async relationships are (in my mind) a smell. It’s really hard to control the data flow and timing of changes. At the route level or a similar higher level, it’s much easier to do this in a functional, controlled manner: either already resolved or whatever handles the promise needs to chain it and be responsible for the work.

JoshSmith:
This appears to not have fixed tests, too.

app/services/task-assignment.js
begedin: Nitpicking, `deleteRecord` is not an api call. `destroyRecord` is. 

However, you are right. `unloadRecord` is the correct choice here.

app/routes/project/tasks/index.js
```diff
@@ -3,14 +3,22 @@ import Ember from 'ember';
...
   projectTaskBoard: service(),
 
-  model() {
-    return this.modelFor('project');
+  async model() {
+    let project = this.modelFor('project');
+    let memberships = get(project, 'organization.organizationMemberships');
+    let members = await RSVP.resolve(memberships.mapBy('member.content'));
+    return RSVP.hash({ project, members });
+  },
```
begedin: In my opinion, this is worse than dealing with it in the previous manner. I'm actually not sure it works correctly.

`RSVP.resolve` is a helper that is intended to be used to simulate a promise (which resolves with the argument provided in the call immediately.

What might work here would be something like

```Javascript
let memberPromises = get(project. 'organization.organizationMemberships').then((memberships) => {
  return memberships.map((membership) => get(membership, 'member'));
});

return RSVP.hash({ project, members: RSVP.all(memberPromises) });
```

This would return a fully resolved model. I don't think the above code would.

Possibly, this would work also, but I'm not sure:

```javascript
let memberships = RSVP.all(get(project, 'organization.organizationMemberships').mapBy('member'));
```

But I'm pretty sure in that case, `.mapBy('member')` would just return null for each item.

Generally speaking, though, hashes in models are considered a smell to, and in the case of the previous approach, the whole purpose of `DS.PromiseObject`, `DS.PromiseArray` and `DS.PromiseManyArray`, which is what requests that `get` async relationships do, is to deal with the smell of promises in a computed.

These objects proxy all `get` and element requests to the `content` property, which is null until the associated promise resolves. The whole idea of having these classes in `ember-data` is so that promises "just work".

It's my opinion that this should be ok. Resolving all promises in the model hook, however, is probably the next best thing. The above code has a bug, though.

Just to reiterate:

```javascript
RSVP.resolve(x)
```

returns a promise which instantly resolves with x. Doesn't matter what that x is.


```javascript
RSVP.all(arrayOfPromises)
```

returns a promise which resolves when all promises in the array resolve.

JoshSmith:
@begedin can you confirm that this is not addressing 1083 and update accordingly?

JoshSmith:
@begedin removing ref to #1083 and replacing with what I believe is #1090.

app/services/task-assignment.js
JoshSmith: I've updated now to `unloadRecord`.

PR #1093, opened 2017-02-23
[SKILL-BUTTON]: rewrite to put logic into component js

### LINKED ISSUE #1044: Rewrite skill-button so there's no unnecessarily deep HTML
# Problem

Right now, the `skill-button` is a `div` which, internally, renders a single, albeit complicated `button` tag.

We should rewrite it so that component itself is a `button` and all the associated bindings are determined at the component root level

## Subtasks
- [x] Remove `button` "wrapper" tag from the `skill-button.hbs
- [x] Add `tagName: button` to `skill-button.js`
- [x] Reimplement class and action binding at the `.js` instead of the `.hbs` level

This task probably requires a decent amount of ember component knowledge, but with guidance, should not be too complicated.

snewcomer:
# What's in this PR?
Move hbs logic into js file

## References
Fixes #1044

Progress on: #

### DIFF
--- a/app/components/skill-button.js
+++ b/app/components/skill-button.js
@@ -10,13 +10,32 @@ const {
 
 export default Component.extend({
   alwaysShowX: false,
+  attributeBindings: ['isLoading:disabled'],
+  classNames: ['has-spinner', 'skill'],
+  classNameBindings: ['isHovering:can-delete', 'size', 'type'],
   hasCheck: false,
+  iconAfter: not('iconBefore'),
   iconBefore: true,
   isHovering: false,
   size: 'large',
+  tagName: 'button',
 
-  iconAfter: not('iconBefore'),
+  click() {
+    get(this, 'remove')();
+  },
 
+  mouseEnter() {
+    set(this, 'isHovering', true);
+  },
+
+  mouseLeave() {
+    set(this, 'isHovering', false);
+  },
+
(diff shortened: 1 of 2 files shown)


app/components/skill-button.js
```diff
@@ -15,8 +15,26 @@ export default Component.extend({
...
+  mouseEnter() {
+    set(this, 'isHovering', true);
+  },
+  mouseLeave() {
+    set(this, 'isHovering', false);
+  },
+
+  click() {
+    get(this, 'remove')();
+  },
```
JoshSmith: Can you reorder the fns to be alpha-ordered and with a space between each?

app/components/skill-button.js
```diff
@@ -15,8 +15,26 @@ export default Component.extend({
   isHovering: false,
   size: 'large',
 
+  classNames: ['has-spinner', 'skill'],
+  classNameBindings: ['isHovering:can-delete', 'size', 'type'],
+
+  attributeBindings: ['isLoading:disabled'],
+
+  tagName: 'button',
```
JoshSmith: Can you interleave these with the top set of attributes in alpha order?

snewcomer: fixed

JoshSmith:
@snewcomer can you squash and merge?

PR #1101, opened 2017-02-24
Fix membership async loading for tasks

JoshSmith:
# What's in this PR?

Fixes the error on production right now where the `members` are an array of `null` values, as predicted by @begedin.

### DIFF
--- a/app/routes/project/tasks/index.js
+++ b/app/routes/project/tasks/index.js
@@ -10,10 +10,12 @@ const {
 export default Route.extend({
   projectTaskBoard: service(),
 
-  model() {
+  async model() {
     let project = this.modelFor('project');
-    let members = RSVP.all(get(project, 'organization.organizationMemberships').mapBy('member'));
-    return RSVP.hash({ project, members });
+    let memberPromises = await get(project, 'organization.organizationMemberships').then((memberships) => {
+      return memberships.map((membership) => get(membership, 'member'));
+    });
+    return RSVP.hash({ project, members: RSVP.all(memberPromises) });
   },
 
   setupController(controller, models) {


JoshSmith:
I could not make a test that actually fails for this. What I was able to do in dev though, was do `Process.sleep(5000)` in the API which makes the error happen every time. With this code, the error does not happen.

JoshSmith:
@begedin I merged just to fix the issues in production right now, because they are not good.

app/routes/project/tasks/index.js
```diff
@@ -10,10 +10,12 @@ const {
 export default Route.extend({
   projectTaskBoard: service(),
 
-  model() {
+  async model() {
     let project = this.modelFor('project');
-    let members = RSVP.all(get(project, 'organization.organizationMemberships').mapBy('member'));
```
begedin: Let me try and explain why this failed

```Javascript
get(project, 'organization.organizationMemberships');
```

The result of this is not a promise. It's a `DS.ProxyArray`. The way that this object behaves is the following.
* It has a `content` property. It delegates all array/property gets to that `content` property. 
* It has a `promise` property, where the get request to the API is stored. We can chain with a `.then` from that, because that's how this class works. Either way, once that promise resolves, the result will be set to the `content`
* Due to that, `.mapBy('member')` will simply delegate to content, and return a bunch of nulls.


The point of the `DS.PromiseObject` and `DS.PromiseArray` classes is to deal with something like this. Basically, they are data objects and promises at the same time. We can chain from them like a promise, or we can bind to the data, but the values will be null until the promise resolves internally.

That is why I would still prefer if we tried the route I originally suggested. I really don't see that as a smell. We return just the project as the model, and then on a controller level, we define alias/mapBy to get to the value we need. At that point, we can just test the component with fully resolved values. The idea of these two classes is to enable us to do that. They're supposed to "just work".

PR #1102, opened 2017-02-24
Push fix for task list to production

JoshSmith:


JoshSmith:
@begedin pushed to prod. Tested on staging and appears to work.

PR #1119, opened 2017-03-02
Fix alignment of ember-power-select-option icon in Firefox

### LINKED ISSUE #1116: Icon in assign task list out of alignment in firefox
# Problem
Firefox issue:
In dropdown assign task list, the icon meant to be next to assigned contributors is on top of the name.

## References
[https://cl.ly/2V082B090I1P](https://cl.ly/2V082B090I1P)
Progress on: #

filipecrosk:
# What's in this PR?

Fix alignment of ember-power-select-option icon in Firefox

## References
Fixes #1116

### DIFF
--- a/app/styles/layout/_select-inline.scss
+++ b/app/styles/layout/_select-inline.scss
@@ -201,6 +201,7 @@ $list-height: 220px;
     &:before {
       content: "";
       height: 10px;
+      left: 10px; // fix for firefox inline
       margin: 5px 7px 0 0;
       position: absolute;
       width: $icon-width;


app/styles/layout/_select-inline.scss
```diff
@@ -204,6 +204,7 @@ $list-height: 220px;
       margin: 5px 7px 0 0;
       position: absolute;
       width: $icon-width;
+      left: 10px;
```
JoshSmith: @filipecrosk I would maybe move this above `margin` for alphabetical ordering and then just comment that this is a fix for Firefox inline, so we know. Can you show a screenshot of the fix, btw?

filipecrosk: Do we have some style guides for CSS? Just alphabetical ordering or maybe like this:

## Declaration Order

Declarations should first be consistent with the rest of the
application. After that it should be in order of what affects height.

`margin`
`padding`
`display`
`position`
`top`
`right`
`bottom`
`left`
`float`
`overflow`
`border`
`border-radius`
`background`
`font`
`line-height`
`color`
`letter-spacing`
`transitions`

If there is a shorthand version for that set of styles, follow in the
order of the shorthand. For `margin` that would be `margin-top`,
`margin-right`, `margin-bottom` then `margin-left`. For `border` that
would be `border-width`, `border-style` then `border-color`.

from: https://github.com/dockyard/styleguides/blob/master/ux-dev/css.md#declaration-order

chjpugh: This is definitely the right fix.  I did a codepen because I could not figure out why: http://codepen.io/chjpugh/pen/yMJLZw

Basically, because the left is not set, the spec is to make it auto.  For Firefox, that means treating this basically as a relative div (so the float affects it).  For Chrome, it for some reason adds the left = 10px.

I took a screenshot of both browser's compiling-
Firefox: https://cl.ly/2h3z2Y0i3i3b
Chrome: https://cl.ly/2t2n1P2r1a1N

JoshSmith:
@filipecrosk can you rebase on develop and squash into one commit, then merge? Thanks!

app/styles/layout/_select-inline.scss
JoshSmith: @filipecrosk we just do alphabetically for the sake of newcomers to the project. It's the only foolproof ordering method I've found here.

PR #1129, opened 2017-03-08
Hide task assignment dropdown if user has no ability to assign

### LINKED ISSUE #1081: Task assignment should not be accessible when you're unauthenticated
The task assignment should not be accessible when you're unauthenticated, but it apparently is accessible when you're not logged in.

https://sentry.io/code-corps/code-corps-ember/issues/222617837/

```
Error: Ember Data Request POST https://api.codecorps.org/user-tasks returned a 401
Payload (application/json; charset=utf-8)
[object Object]
  at comparableMoment (moment/index.js:58)
  at comparableMoment (moment/index.js:58)
  at comparableMoment (moment/index.js:58)
  at update (bower_components/ember/ember.prod.js:41934)
  at comparableMoment (moment/index.js:58)
...
(6 additional frame(s) were not displayed)
```

begedin:
# What's in this PR?

Closes #1081 

We already had the ability as a component property, so all it took was to update the template and add a test.

From now on, 

- if user has the ability to assign (author, or at least project contributor), the dropdown will render
- if user has no ability, the assigned user will render if there is one, otherwise, nothing will render

## References
Fixes #1081

### DIFF
--- a/app/components/task-card.js
+++ b/app/components/task-card.js
@@ -29,6 +29,7 @@ export default Component.extend({
   bound: false,
   shouldShowUsers: false,
 
+  // auto-assigns 'task' property from component as ability 'model'
   ability: EmberCan.computed.ability('task'),
   canAssign: alias('ability.canAssign'),
   canEdit: alias('ability.canEdit'),
--- a/app/styles/layout/_select-inline.scss
+++ b/app/styles/layout/_select-inline.scss
@@ -45,8 +45,9 @@ $list-height: 220px;
 }
 
 .select-inline {
-  display: flex;
   align-items: center;
+  display: flex;
+  font-size: $body-font-size-small;
 
   .ember-power-select-trigger {
     align-items: left;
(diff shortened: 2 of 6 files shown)


begedin:
@JoshSmith Made the changes. This made the styling I added previously obsolete, so I removed it. Should be good to go.

JoshSmith:
I'm still not sure this is what we want.

When signed out, I see the following:

screen shot 2017-03-14 at 9 33 37 pm

I would not expect to be able to click the dropdown nor for the link to be hoverable.

Should we talk about this further synchronously to clarify what's being miscommunicated?

begedin:
image

This is what it looks like for me when I look at the task list when not signed in (and images not loading, due to env issues). As far as I can tell, images not loading should not cause the lack of arrow, since that's pure css.

It loads fine and is interractive when I sign in:

image

In mirage development, it looks fine to:

image

Not sure what the issue is, but definitely something that would be good to debug paired.

JoshSmith:
Hm, weird. I don't have this issue any longer.

That said, `font-size` was different here and needed to be fixed. That's fixed now, so I think this can be merged.

PR #1130, opened 2017-03-08
Add ability to quick-add project skills to new task

### LINKED ISSUE #1052: Add project skills to the skill assignment UI on the new task page

begedin:
# What's in this PR?

A new set of components which lists project skills in the new tasks sidebar. Clicking these assigns them as task skills to the new task.

## References
Fixes #1052

### DIFF
--- a/app/styles/_buttons.scss
+++ b/app/styles/_buttons.scss
@@ -53,16 +53,6 @@
     @include focus($default-color);
   }
 
-  &.small {
-    font-size: $body-font-size-small;
-    padding: 6px 14px;
-
-    &.skill {
-      padding: 6px 9px;
-      font-size: $body-font-size-normal;
-    }
-  }
-
   &.large {
     font-size: $body-font-size-large;
     padding: 15px 40px;
@@ -88,6 +78,16 @@
 
     &.skill {
       padding: 11px 14px;
+    }
+  }
+
+  &.small {
+    font-size: $body-font-size-small;
+    padding: 6px 14px;
+
+    &.skill {
+      padding: 6px 9px;
+      font-size: $body-font-size-normal;
     }
   }
 
--- a/app/components/project-skill-item.js
+++ b/app/components/project-skill-item.js
@@ -1,0 +1,16 @@
+import Ember from 'ember';
+
+const {
+  Component,
+  getProperties
+} = Ember;
+
+export default Component.extend({
+  classNames: ['skill', 'has-spinner', 'default', 'small'],
+  tagName: 'button',
+
+  click() {
+    let { skill, onClicked } = getProperties(this, 'skill', 'onClicked');
+    onClicked(skill);
+  }
+});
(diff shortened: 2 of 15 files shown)


app/styles/_buttons.scss
```diff
@@ -53,16 +53,6 @@
...
-  &.small {
-    font-size: $body-font-size-small;
-    padding: 6px 14px;
-
-    &.skill {
-      padding: 6px 9px;
-      font-size: $body-font-size-normal;
-    }
-  }
-
```
begedin: Style for `.default` was overriding the style for `.small` so I had to reorder. That being said, this set of classes is really difficult to follow, so we should consider rewriting it.

begedin:
@JoshSmith Fixed the issue. I'm afraid it means a slightly more complex computed. The objects in `projectSkills` are proxies, while `excludedSkills` are plain models, so we can't just use `setDiff`. Instead, it has to be computed that figures our exclusion based on id.

JoshSmith:
This has a bug still.

begedin:
@JoshSmith While part of the behavior was fixed with the latest commit, the UI itself still has a bug, which will take a longer time to fix. 

If I add a skill using the project skill list, then try to search for the same skill in the dropdown, it will appear "uncrossed", as if not yet added, in the results:

image

Even though it looks incorrectly, clicking it will still remove it, not add a duplicate, so at least the background logic works.

Basically,
- `skill-typeahead-result` usually uses `service:x-skills.includes(skill)` to determine if skill has already been added. this method is NOT easily rewritten into something else
- In the case of `task.new`, the `skill-typeahead-result` instead uses just an `array.includes(skill)`, but this no longer works when we add project skills into the mix, because those are `DS.PromiseObject` records, not `DS.Model` records. This happened to work because the method was named the same, but the proper method to use here is `array.isAny('id', skillId)` so it works with both `DS.Model` and `DS.PromiseObject` records

- monkeypatch the `includes` method in this specific case - quick and dirty, definitely against it
- wrap the unsaved skills collection into another service, which makes it work - clean, but makes thing worse in the long run, I think. Another complex service, difficult to follow.
- rethink the `projectSkillsList` component added here. make it an extremely simple `skills-item-list` component which just receives and renders a list of skills. let the controller handle the exclusion of already added skills, etc. - clean and doesn't make things worse; a good compromise.
- rewrite the services, rethink them, really. now that I got back into them, I found them extremely difficult to follow and, especially with "sometimes" accessing the `content` property of proxy objects, ex: 

```Javascript
return records.any((found) => {
  let targetId = get(target, 'id');
  let targetModelName = 
     get(target, 'constructor.modelName') || get(target, 'content.constructor.modelName');
  let foundId = found.belongsTo(targetModelName).id();

  return (foundId === targetId);
});
```

I think, with proper controller usage, the whole xSkillsList collection of services could be almost if not completely eliminated.

begedin:
My vote here is, since the UI works and there's just a graphical glitch, we merge this, create an issue, copy my previous comment there and discuss the solution.

PR #1135, opened 2017-03-08
[HELLO]: disable btn while uploading.

### LINKED ISSUE #1009: Transitioning during signup while photo is uploading fails
# Problem

- When on `start/hello`, started uploading a photo.
- Hit submit button to transition to `start/interests`.
- Successful transition to `start/interests`.
- Photo finishes uploading and Ember tries to `PATCH /users/:id`.
- User has the following attributes:

```
{
  "data": {
    "attributes": {
      "cloudinary-public-id": "gogsd44pqghib04nk0fz",
      "state": "edited_profile",
      "state-transition": "edit_profile"
    },
  }
}
```

This results in a `422` error because we're attempting to transition from `edited_profile` to `edited_profile` via the `edit_profile` transition, which is not allowed.

snewcomer:
# What's in this PR?

- [x] disabled btn while user is uploading image during signup.

## References
Fixes #1009

### DIFF
--- a/app/controllers/start/hello.js
+++ b/app/controllers/start/hello.js
@@ -10,6 +10,13 @@ const {
 } = Ember;
 
 export default Controller.extend(OnboardingControllerMixin, {
+  /**
+   * disables continue button while uploading
+   * @prop uploadingImage
+   * @default false
+   */
+  uploadingImage: false,
+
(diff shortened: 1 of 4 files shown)


JoshSmith:
:raised_hands:

Issue #1133, opened 2017-03-08
Add a website field to Project model and its settings form


JoshSmith:
# Problem

We want to be able to link out to projects. We're going to add a `website` field to `Project` so that we can allow for this linking later.

We should also add a `shouldLinkExternally` property to `Project`.

Both will need model tests.

JoshSmith:
Closed by #1136.

PR #1136, opened 2017-03-09
Addresses #1133 Adds website field for projects also adds boolean to

jderr-mx:
model for shouldLinkExternally

# What's in this PR?

List the changes you made and your reasons for them.

Make sure any changes to code include changes to documentation.

## References

Addresses #1133 Adds `website` field to `project` model and form. Also adds `shouldLinkExternally` boolean to `project` model

### DIFF
--- a/mirage/scenarios/default.js
+++ b/mirage/scenarios/default.js
@@ -185,7 +185,8 @@ export default function(server) {
     organization,
     owner,
     slug: 'code-corps',
-    title: 'Code Corps'
+    title: 'Code Corps',
+    website: 'https://www.codecorps.org'
   });
 
   server.create('project-user', { user: owner, project, role: 'owner' });
--- a/app/models/project.js
+++ b/app/models/project.js
@@ -16,9 +16,11 @@ export default Model.extend({
   longDescriptionBody: attr(),
   longDescriptionMarkdown: attr(),
   openTasksCount: attr('number'),
+  shouldLinkExternally: attr('boolean'),
   slug: attr(),
   title: attr(),
   totalMonthlyDonated: attr('dollar-cents'),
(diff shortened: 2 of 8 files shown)


package.json
```diff
@@ -11,7 +11,8 @@
   "scripts": {
     "build": "ember build",
     "start": "ember server",
-    "test": "ember try:one default"
+    "test": "ember try:one default",
+    "ember-exam": "ember exam --split=3 --parallel --weighted"
```
JoshSmith: Why was this added?

mirage/scenarios/default.js
```diff
@@ -185,7 +185,8 @@ export default function(server) {
     organization,
     owner,
     slug: 'code-corps',
-    title: 'Code Corps'
+    title: 'Code Corps',
+    website: 'www.codecorps.org'
```
JoshSmith: I think we want this to be prefixed `https://` here, since that's what the server would do.

package.json
```diff
@@ -11,7 +11,7 @@
   "scripts": {
     "build": "ember build",
     "start": "ember server",
-    "test": "ember try:one default"
+    "test": "ember try:one default",
```
JoshSmith: Would remove the trailing comma here.

JoshSmith:
@jderr-mx it looks like the build is failing to start – not sure if it's because of that trailing comma issue or what.

mirage/scenarios/default.js
jderr-mx: Yes! I did this with the other tests, so it makes sense

jderr-mx:
@JoshSmith I think this one is set now

JoshSmith:
🙌

PR #1145, opened 2017-03-11
Fixes #1143 sort projects by id

### LINKED ISSUE #1143: Order projects by id on projects list
# Problem

The API returns projects ordered by id but it appears that the ordering is variable depending on behaviors taken in the client. We should explicitly order by id and write a test for this specifically.

jderr-mx:
# What's in this PR?

Sorts projects by id

## References
Fixes #1143 

Progress on: #

### DIFF
--- a/tests/integration/components/project-list-test.js
+++ b/tests/integration/components/project-list-test.js
@@ -1,8 +1,23 @@
 import { moduleForComponent, test } from 'ember-qunit';
 import hbs from 'htmlbars-inline-precompile';
+import Ember from 'ember';
+import PageObject from 'ember-cli-page-object';
+import component from 'code-corps-ember/tests/pages/components/project-list';
+
+const {
+  set
+} = Ember;
+
+let page = PageObject.create(component);
 
 moduleForComponent('project-list', 'Integration | Component | project list', {
-  integration: true
+  integration: true,
+  beforeEach() {
+    page.setContext(this);
+  },
+  afterEach() {
+    page.removeContext();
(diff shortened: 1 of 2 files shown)


jderr-mx:
@JoshSmith This one is ready for review. The build failed on the step `yarn global add phantomjs-prebuilt` in circle-ci. It builds and tests pass on my end. Not sure if I have permissions to trigger a build without pushing.

tests/integration/components/project-list-test.js
```diff
@@ -21,8 +36,27 @@ test('it renders an item for each project in the list', function(assert) {
...
+
+  mockProjectIds.forEach(function(id) {
+    mockProjects.push({ id, title: `Project ${id}`, slug: `project_${id}` });
+  });
+
+  set(this, 'projects', mockProjects);
+  page.render(hbs`{{project-list projects=projects}}`);
+
+  for (let i = 0; i < 3; i++) {
+    let id = i + 1;
```
filipecrosk: I think you could do a little small improvement here, just remove the id line and change the for to `for (let id = 1; id <= 3; id++) {`. 

Nice work! 👍

jderr-mx: Thanks!

I needed the 0 - 2 of `i` for the index of `page.items` but I need 1 - 3 for `id` to assert project id's if that makes sense.

filipecrosk: you're right, my bad, sorry 🙈

PR #1147, opened 2017-03-13
Switch project owners to projectUser records

### LINKED ISSUE #1142: Replace project.owner with ProjectUser where role is set to "owner"
# Problem

Separating owners from other project members by having an explicit `project.owner` association adds needless complexity. This should instead be replaced with an explicitly created `projectUser` record where the role is set to `"owner"`. A similar thing should happen on the API.

- remove references from models and mirage models
- update abilities, authorization, etc.
- update mirage seeds

begedin:
# What's in this PR?

Removes `project.owner` concept and replaces it with a `projectUser` record with `role: 'owner'`

## References
Fixes #1142
Related to: https://github.com/code-corps/code-corps-api/pull/751

### DIFF
--- a/app/abilities/project.js
+++ b/app/abilities/project.js
@@ -3,9 +3,10 @@ import { Ability } from 'ember-can';
 
 const {
   computed,
-  computed: { alias },
+  computed: { alias, equal },
   get,
-  inject: { service }
+  inject: { service },
+  isEmpty
 } = Ember;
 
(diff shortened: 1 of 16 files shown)


JoshSmith:
@begedin looks good, do you want to add follow-up issues for this?

begedin:
Created #1150, and #1151

PR #1156, opened 2017-03-16
Add ember-stripe-elements

JoshSmith:
# What's in this PR?

Adds `ember-stripe-elements` and removes a lot of our custom CC code.

Cannot test filling in the form due to cross-origin issues with the `iframe`. Wasted a lot of time trying to do so. Am going to see if I can't get Stripe to add some sort of ability to at least be able to fill the form, though I understand the implications of allowing this. At the very least, a testable component would be a nice alternative.


JoshSmith:
A lot of the lines of code added here are from `svg`s.

app/controllers/project/donate.js
```diff
@@ -54,15 +56,19 @@ export default Controller.extend({
...
+    let stripe = get(this, 'stripev3');
 
-    return stripe.card.createToken(stripeCard)
-                      .then((stripeResponse) => RSVP.resolve(stripeResponse))
-                      .catch((reason) => this._handleCreditCardTokenError(reason));
+    return await stripe.createToken(stripeElement).then(({ error, token }) => {
+      if (error) {
+        return this._handleCreditCardTokenError(error);
+      }
+      if (token) {
```
begedin: Realistically, the case should probably not happen ever, but I still seems more correct to have an `else if` here.

PR #1170, opened 2017-03-20
Add project joining modal

JoshSmith:
# What's in this PR?

This adds the project joining modal dialog.

The implementation is basically fully working. I might add more polish but this is essentially shippable in its current state.

Before merging, though, we should (at least):

- [x] Add page objects
- [x] Write component tests
- [x] Write acceptance tests
- [x] Clean up the template logic of the project card's buttons
- [x] Optionally remove the usage of `store` in the component
- [x] Remove tests for tooltips in `skill-list-item-link`
- [x] Maybe remove styling for these tooltips? If it exists


app/styles/components/project-join-modal.scss
```diff
@@ -0,0 +1,59 @@
...
+.project-join-modal__skills {
+  margin-bottom: 20px;
+  text-align: center;
+
+  p {
+    font-size: $body-font-size-normal;
+    margin-bottom: 10px;
+  }
+
+  .related-skills .skills {
```
JoshSmith: I don't like that this styling is happening in here. Seems like there should be a more universal way to fix this.

app/templates/components/project-card.hbs
```diff
@@ -13,6 +13,29 @@
...
+          {{/if}}
+        {{/if}}
+
+        {{#unless currentProjectMembership}}
+          <button class="button default small" {{action "toggleJoinProjectModal"}}>Join project</button>
+        {{/unless}}
+      {{else}}
+        {{link-to "Join project" "signup" class="button default small"}}
+      {{/if}}
+    {{/if}}
```
JoshSmith: All of this logic is way too complex.

app/templates/components/project-join-modal.hbs
```diff
@@ -0,0 +1,27 @@
...
+          <img class="icon medium" src={{user.photoLargeUrl}} />
+        </div>
+        <div class="project-join-modal__icons__icon project-join-modal__icons__icon--join"></div>
+        <div class="project-join-modal__icons__avatar">
+          <img class="icon medium" src={{project.iconLargeUrl}} />
+        </div>
+      </div>
+      <div class="project-join-modal__skills">
+        <p>What can you help with?</p>
+        {{related-skills isClickable=true skills=skills}}
```
JoshSmith: This `isClickable` is not at all tested on a couple of the higher-level components.

app/templates/components/project-card.hbs
begedin: I've simplified this some by moving the modal and the button that opens it into the same component. My suggestion for further simplification would be to do some extraction, but probably in a separate PR.

The two base cases are:

- There is a membership
- There is no membership

For the first one, show a `membership-status` component with 3 states

- loading
- request sent (role eq pending)
- member (role eq something else)

For the second one, we just use the current component.

```hbs
{{#if session.isAuthenticated}}
  {{#if currentMembership}}
    {{membership-status membership=currentMembership}}
  {{else}}
    {{project-join-modal onJoin=(action onJoin project) project=project skills=projectSkills}}
  {{/if}}
{{else}}
  {{link-to "Join project" "signup" class="button default small"}}
{{/if}}
```

JoshSmith:
@begedin had mentioned this morning that there page objects for working with modals. Surfacing that here.

Do let me know if you'll be working on this further, and whenever you're done with it, assign over to me so I'm tracking.

begedin:
@JoshSmith Ready for review. The button logic in project card was simplified some, but there's more room for improvement. Probably better done as part of another PR.

Also, while doing this, I noticed the project-card component is severely under-tested.

Before merge:

- [ ] Create issue: Simplify project card button logic
- [ ] Create issue: Write integration tests for project card

app/templates/components/project-card.hbs
JoshSmith: What do you think about this being called `project-user-status` Instead?

JoshSmith:
I also added the modal to the project page itself. We need to fix the duplication here.

@begedin I'm wondering if we should open an issue to move the actions defined as `joinProject` up into either a service, or even into the application controller. I think maybe the latter makes the most sense, actually?

JoshSmith:
I refactored the `joinProject` action to instead call a `projectUser` service with `joinProject`.

JoshSmith:
We need to create an issue to write project user service tests @begedin.

PR #1183, opened 2017-03-22
Allow users to signup from donate route

### LINKED ISSUE #1099: Add signup/login form to the donation page when unauthenticated
# Problem

When trying to donate to a project, we should have a form that should be switchable to/from sign up and login, with signup as the default. Once you sign up, the application should redirect you to the donation page you were at to move onto the next step.

begedin:
# What's in this PR?

I tried to make this change as lightweight as possible, but also squeezed in some unplanned changes.

I was initially going for changing the donate route template. While trying to do that, I noticed difficulties with the way the routes and outlets are structured, so the first two commits were attempts to simplify that. Then I moved away from that solution, but I would still consider them improvements, so I left them in.

## Refactorings

- Moved `project-header` to `donation/project-header`, to make it clearer what this component is, and what it's used for. It's a special form of the regular `project-header` component specific to the donation pages.
- Renamed `project-details` to `project-header` to make it clear that this is the standard `project-header` component used in most project routes.
- Moved common parts of the project subroutes into the `project.hbs` template, since this is the intended way to structure templates. 
- Explicitly specified that the `project.tasks`, `project.donate` and `project.thank-you` routes, render into the parent, `application` template instead of the default `project` template. We should look into further unifying this. **I feel there is no reason for the task route to have the different, wide mode for the header. The UX for that feels weird.**

The actual change that happens here is the following:

## Solution

* The donate route no longer requires authentication in a standard way
* If not authenticated, the user will get redirected to the signup route, with a `?donate=true`. The current transition will be saved so the session will go back to it, upon sign in, after signup.
* The signup route still initializes a new, unsaved user, but if `donate` is set to true, it initializes it with the `state` set to `signed_up_donating`.
* When the user is created, once the API changes are in, the signup route will redirect to the donate route same as if the user simply signed in.

I feel this solution makes for the least amount of changes to the existing system. The first two commits should also help with other, UI tasks in this project.

## References
Fixes #1099 
Related tohttps://github.com/code-corps/code-corps-api/pull/775

**This PR should not go into production until the API PR is ready to go.**


JoshSmith:
I think we should be doing something like the following:

- Move the donate component on projects to the donation route
- Clicking "Become a donor" on the project's page takes you to `/donate/new` where you select/enter the amount
- Clicking "Donate" from here will attempt a transition to `/donate/checkout?amount=10`, eg.
  - If authenticated, you will see the credit card form
  - If unauthenticated, you will see a signup/signin form
    - The router keeps a `next` query parameter indicating the route should we should transition to (`/donate/checkout?amount=10`)
    - You can switch back and forth between "Sign Up" and "Sign In" and the `next` should remain the same
    - This means we can extend this to other actions in the future, and keep complex (non-sensitive) data in these transitions

app/routes/signup.js
```diff
@@ -3,21 +3,35 @@ import UnauthenticatedRouteMixin from 'ember-simple-auth/mixins/unauthenticated-
...
+   *
+   * There is a specific case when an unauthenticated user will navigate to the
+   * project.donate route. They will get redirected to signup here, with a
+   * `donate=true` query parameter.
+   *
+   * This changes initialization of the user so they start with a 'donating'
+   * state, instead of the default 'signing_up'.
+   */
+  model({ donate }) {
+    let user = donate ? { state: 'donating' } : {};
```
JoshSmith: This does not match the API, which has `signed_up_donating`.

app/templates/components/signup-form.hbs
```diff
@@ -1,7 +1,11 @@
 <div class="container">
   <form>
 
-    <h2>Join Code Corps today.</h2>
+    {{#if donate}}
+      <h2>Join Code Corps today.</h2>
+    {{else}}
+      <h2>Join Code Corps to continue.</h2>
+    {{/if}}
```
JoshSmith: Didn't get this working yet.

app/templates/project/donate.hbs
```diff
@@ -1,15 +1,5 @@
...
-    isProcessing=isProcessing
-    projectTitle=project.title
-    saveAndDonate=(action 'saveAndDonate' amount)
-  }}
-    {{#if error}}
-      {{error-formatter error=error}}
-    {{/if}}
-  {{/donation/donation-container}}
+
+  {{donations/create-donation amount=amount continue=(action "continueToCheckout")}}
```
JoshSmith: This likely is not tested.

tests/acceptance/project-checkout-test.js
```diff
@@ -5,7 +5,7 @@ import Mirage from 'ember-cli-mirage';
 
 import { authenticateSession } from 'code-corps-ember/tests/helpers/ember-simple-auth';
 import { getFlashMessageCount } from 'code-corps-ember/tests/helpers/flash-message';
-import projectDonatePage from '../pages/project/donate';
```
JoshSmith: We'll need a page for the new `projectDonatePage`, borrowing probably from the component (or using it directly, I think).

app/styles/components/donations/create-donation.scss
```diff
@@ -0,0 +1,59 @@
...
+
+  @include media($md-screen) {
+    @include shift(1);
+    @include span-columns(10);
+  }
+
+  @include media($sm-screen) {
+    @include shift(0);
+    @include span-columns(12);
+  }
```
JoshSmith: There's a lot of shared code here between the two styles of donation pages, this and the checkout flow.

app/components/donations/create-donation.js
```diff
@@ -1,11 +1,24 @@
...
+
+  init() {
+    this._super(...arguments);
+
+    let amount = get(this, 'amount');
+    let presetAmounts = get(this, 'presetAmounts');
+
+    if (!presetAmounts.includes(amount)) {
+      set(this, 'customAmount', amount);
+    }
```
JoshSmith: Amount handling here is a little funky for floats, etc. I would also expect the URL to actually change when the `customAmount` changes, too.

JoshSmith:
Actually, I realized that this is still really missing tests, etc. It probably should get another review from me as a separate series of commits. Sorry for confusion above.

JoshSmith:
This could use a really thorough once-over review, too, to see what I missed.

app/templates/components/donations/create-donation.hbs
```diff
@@ -1,11 +1,10 @@
...
 <div class="amounts">
 {{#each presetAmounts as |presetAmount|}}
   {{donations/donation-amount-button
-    customAmount=customAmount
     presetAmount=presetAmount
-    selectedAmount=amount
-    setCustomAmount=(action (mut customAmount))
-    setAmount=(action (mut amount))
+    click=(action (mut amount) presetAmount)
+    selected=(and (eq presetAmount (float amount)) (not customAmount))
```
begedin: A lot of data was being set down for the sake of computing the flag. I figured, makes more sense to compute the flag where the data is and send that flag down instead.

The part that is weird to me is the `(not customAmount)`. This statement here basically says, "if amount is equal to one of the 4 button amounts, render that button as selected, unless that amount is written in the custom amount box.

It would make sense to highlight the appropriate button in that case as well, but there is a test that explicitly ensures that behavior is as defined here, so I did not touch it.

app/templates/components/donations/create-donation.hbs
```diff
@@ -1,11 +1,10 @@
+<h3>Support this project each month</h3>
 <div class="amounts">
 {{#each presetAmounts as |presetAmount|}}
   {{donations/donation-amount-button
-    customAmount=customAmount
     presetAmount=presetAmount
-    selectedAmount=amount
-    setCustomAmount=(action (mut customAmount))
-    setAmount=(action (mut amount))
+    click=(action (mut amount) presetAmount)
```
begedin: There was a click handler inside the child component, which triggers 2 actions on the parent. This removes that behavior and simply puts the parent in charge of the click. On click, `amount` will be set to the value of `presetAmount`

app/helpers/float.js
```diff
@@ -0,0 +1,22 @@
...
+ * a string indicating a float
+ *
+ * @param  {String}  value   The value which could be a float or a string
+ * @return {Number}          The value, in float
+ */
+export function float([value]) {
+  return convertToFloat(value);
+}
+
+export default helper(float);
```
begedin: A helper which we can use to parse a string/float value into a float. Basic now, but we can expand on it as required.

begedin:
@JoshSmith Took me a bit to figure everything out, but I think I got it to a reviewable/mergeable state no. I'm not really sure, with the review comments. It's not completely clear to me which are still relevant.

Some parts, I've made simpler, but I think the behavior is the way you wanted it.

app/components/donation/project-header.js
```diff
@@ -0,0 +1,9 @@
+import Ember from 'ember';
+
+const {
+  Component
+} = Ember;
+
+export default Component.extend({
+  classNames: ['project-header']
```
JoshSmith: This is confusing compared to project__header.

begedin: This is from the old code and styles, unchanged, just copied to a new location. I agree, but it should be part of a "fix" issue. I think the styles are from before the switch BEM, even.

begedin:
>It seems a little odd to me to be doing this like this. Why not use the component lifecycle hooks here to update values in the component?

I'm not sure to which part you're referring to.

**If it's a question of why is this not still in the `donation-amount-button` component**

It's a matter of concerns. All possible options for a collection of `donation-amount-button` components should not be a concern of a single `donation-amount-button` component. The parent's currently selected amount should not be its concern either. 

We were sending all that data down to compute a flag, which depends on the parent and the siblings. The parent should be in charge of itself and the siblings.

**If it's a matter of it being in the template instead of `create-button.js`**

My argument for this is, it's a simple condition, we have the helpers for it installed as an addon and this way, it's immediately clear from the template, instead of us having to go into the `.js` to figure it out.

> Does this convert 50.5 to 50.50? Looks like probably not.

I'm assuming you're talking about the `float` helper. It does not and it should not. It convers to a float. `50.5 === 50.50` and the javascript output for it will be `50.5`

image

For formatting, we have the `format-currency` helper and it's already used for the `donation-amount-button` component to format the amount on the button.

I'm thinking you might want to format the query param, but again, from the comments, it's not immediately clear. We can do that, I think, but I also think it's better fitting as another, smaller issue/PR.

> Given the button component itself is not computing any of this, should we even have a component for it?

It still makes the parent template clearer and it handles the class binding and styling for the selected state. It's a simple component, but it has behavior so I don't have a problem with it remaining here.

begedin:
@JoshSmith The comments got weird here. Not sure why. I tried to answer best I could.

JoshSmith:
Before this can be merged, I'll need to rewrite to have a `fallbackAmount` on the controller which is used when the input is filled and then cleared.

PR #1189, opened 2017-03-24
Start making changes to the onboarding flow

### LINKED ISSUE #1094: Do not redirect the user when loading terms of service or privacy policy pages
# Problem

Our onboarding flow right now redirects you from any route to the onboarding routes if you're in an onboarding state. This means that a user cannot access the terms of service or privacy policy when in this flow.

We should find a way to prevent these pages from ever redirecting on transition.

### LINKED ISSUE #1097: Donation thank you page shows "finish signing up" button if you just signed up
# Problem

When donating and just signing up for the first time, the thank you page should have a button that navigates you to the rest of the onboarding process. You should not be able to navigate elsewhere except to finish onboarding.

### LINKED ISSUE #1098: Change onboarding service to allow for a donating state
# Problem

The onboarding service needs refactored to allow for an additional state of `donating` and the project you were donating to in order to handle redirects and make logical sense of the signup flow.

JoshSmith:
# What's in this PR?

This is a work in progress to make changes to the onboarding flow when donating.

## References
Fixes #1094 
Fixes #1097 
Fixes #1098


JoshSmith:
@begedin I have scattered thoughts around the issues above here, but combined with the progress made here and #1098 especially, you should be able to get the outline of direction I'd ideally like to see.

Hopefully you can provide some thoughts/notes at least on how to proceed, so I can maybe wrap this up tomorrow morning.

app/mixins/donation-route.js
```diff
@@ -0,0 +1,22 @@
...
+
+  afterModel({ project, subscription }) {
+    if (subscription) {
+      get(this, 'flashMessages').success(ALREADY_A_SUBSCRIBER);
+      this.transitionTo('project', project);
+    } else {
+      this._super.call(...arguments);
+    }
+  }
+});
```
begedin: This mixin works, and is probably the next best thing, but combined with the fact that these routes have an explicitly separate layout to other project routes, it's really further telling us that our onboarding routes ought to be grouped under a parent route of some sort. That would allow us to have this `afterModel` as well as our `model` within that parent route. 

It's not something suitable for this route, but it is something to consider and probably discuss.

app/controllers/application.js
```diff
@@ -20,6 +23,10 @@ export default Controller.extend({
   shouldShowFooter: computed.and('isNotOnboarding', 'isNotViewingProjectTaskBoard'),
   shouldShowSpacer: computed.alias('isNotViewingProjectTaskBoard'),
 
+  _updateRoute: observer('currentRouteName', function() {
+    get(this, 'currentRoute').didTransition();
+  }),
```
begedin: I can't say it's clear to me what this part does. I mean, I get the what, not clear on the why

app/mixins/donation-route.js
```diff
@@ -0,0 +1,26 @@
...
+      get(this, 'flashMessages').success(ALREADY_A_SUBSCRIBER);
+      this.transitionTo('project', project);
+    } else {
+      this._super(...arguments);
+    }
+  },
+
+  setupController(controller, models) {
+    controller.setProperties(models);
+  }
```
begedin: I would destructive the hash here, so it's clearer what the models are

app/routes/project/donate.js
```diff
@@ -1,12 +1,42 @@
...
+  userSubscriptions: service(),
+
   model() {
-    return this.modelFor('project').reload();
+    return this.modelFor('project').reload().then((project) => {
+      if (get(this, 'session.isAuthenticated')) {
+        let subscription = get(this, 'userSubscriptions').fetchForProject(project);
+        return RSVP.hash({ project, subscription });
+      } else {
+        return RSVP.hash({ project, subscription: null });
```
begedin: Do we use this behavior anywhere else? If so, maybe fetchForProject should just  return null if no current user.

app/templates/project/donate.hbs
```diff
@@ -1,5 +1,6 @@
 <div class="container">
   {{donation/project-header project=project}}
+  {{log project}}
```
begedin: This one is easily left behind, so just a note here to remove it.

app/routes/project/donate.js
JoshSmith: It looks like it already does...hm.

JoshSmith:
This very badly needs tests. I think we need to test that:

- You can advance forward and backward through the various onboarding states regardless of whether you're currently onboarding
- The onboarding header types show up when you're onboarding
- Clicking the logo in the header when signing up to donate will take you into the onboarding
- The user state is not set is not transitioned in weird ways
- Any various unit testing we can do

JoshSmith:
I've uploaded a demo that shows all the behavior as I would expect it to work, since this is simpler than writing it out: https://cl.ly/2c0M3Q3g0z27

JoshSmith:
I also included the URI in the browser so that you could see where exactly we're navigating.

app/services/current-route.js
```diff
@@ -0,0 +1,16 @@
...
+  set
+} = Ember;
+
+export default Service.extend({
+  currentRouteName: null,
+
+  updateCurrentRouteName() {
+    let { currentPath } = getOwner(this).lookup('controller:application');
+    set(this, 'currentRouteName', currentPath);
+  }
```
begedin: This works fine, but it's a bit of a confusing way to just set a property on an object, which this service really is. It's an object, with a single property, which is a singleton in the context of our application. I'm thinking, why not just have the routes that call `get(this, 'currentRoute').updateCurrentRouteName()`  instead just call something like 

```Javascript
set(this, 'currentRoute.currentRouteName', 'some.new.name');
```

While we're at it, since the name of the service is already `currentRoute`, why not call the property `currentRoute.name`, instead of the redundant `currentRoute.currentRouteName`.

Lastly, and this is a big one, there is a router service in ember, though it's still private right now:

It's deinfed [here](https://github.com/emberjs/ember.js/blob/master/packages/ember-routing/lib/services/routing.js#L26) and is used by the `LinkComponent` [here](https://github.com/emberjs/ember.js/blob/v2.12.0/packages/ember-glimmer/lib/components/link-to.js#L528).

Basically, we can replace the usage of our own currentRoute service by simply replacing injections of it with 

```Javascript
routing: service('-routing'),
```

We can name it `currentRoute` to, if we want.

We can also then alias the `currentRouteName` easily:

```Javascriot
currentRouteName: alias('router.currentRouteName')
```

begedin: Also note that while I do understand any concerns with this service being marked as private right now, there have been RFCs about opening it up, the documentation itself states it, and it seems to be a big part in making components routable, so I would not be too concerned about it.

JoshSmith: Yes, and I think it's relatively small exposure of its API surface area.

PR #1200, opened 2017-03-27
chore(package): update ember-cli to version 2.12.1

JoshSmith:
Closes #1158

https://greenkeeper.io/


begedin:
Will be pushing up some stuff soon, though there's a good amount left to do:

- any service dependencies in unit tests are now explicitly required to be stated in `needs: []`. Before, we could get away with not listing them if they aren't being used for the test, but no longer. This means the tests have a bunch of `attempting to register an uknown factory` errors. Adding stuff to the `needs` array resolves those.
- the exception is `service:store`. It's being injected by ember-data differently, so we can't just do it in a needs. Instead, in the unit test, we should do a `this.subject({ store: mockStore })`. For most tests, the mock will just be an empty object `{}`.
- I did not run `ember init` yet. We should, and perform any package updates, then fix tests again.
- There's a bunch of deprecations, mostly related to component lifecycle hooks, that need handling
- There's also some addon deprecations. We should try updating those addons.

JoshSmith:
I ran ember init already.

begedin:
One more, `stubService` was giving me trouble in some unit tests. If you run into the same, it's simpler to just do a `set(this, 'serviceName', serviceStub)` in those cases. It's a unit test, it should be as unintegrated as possible anyway.

begedin:
This should be ready to go now. There are a couple of deprecations left, but most of them are due to other addons.

# Component lifecycle hooks

There is a bunch of deprecations for using component lifecycle hooks with arguments. 

Some of them come from either ember-cli-dragula. Since the addon author is relatively inactive, but they do accept and welcome PRs, we should consider writing one. #1204

The remainder comes from emberx-select and there as a PR merged to handle it, but there was no new release yet. We should update when it's released. #1205

# Other deprecations

There is a deprecation for ember-cli-deploy-sentry, which we are warned about at build time. I've submitted a PR to that package, which ought to resolve it. It has been merged, but no official release yet.  #1206

There is a deprecation using due to using native dom helpers (`click`, `find`, etc.) in our tests as globals. Apparently, there is now an addon we should add to our app to use these and import from. #1207

begedin:
@JoshSmith Tests are failing due to failing slack notification, but as far as I can tell, nothing has changed in that package to cause it.

PR #1210, opened 2017-03-28
chore(package): update ember-concurrency to version 0.8.1

JoshSmith:
Closes #1198

https://greenkeeper.io/

User menu dropdown test was breaking in `0.8.0`. Not sure yet if it's breaking here, too.


JoshSmith:
That test is breaking here. Needs work still to pass.

PR #1226, opened 2017-03-30
Increase test coverage

JoshSmith:
# What's in this PR?

Increasing the test coverage by adding some tests.

### DIFF
--- a/tests/unit/controllers/start/hello-test.js
+++ b/tests/unit/controllers/start/hello-test.js
@@ -1,4 +1,7 @@
 import { moduleFor, test } from 'ember-qunit';
+import Ember from 'ember';
+
+const { set } = Ember;
 
 moduleFor('controller:start/hello', 'Unit | Controller | start/hello', {
   needs: [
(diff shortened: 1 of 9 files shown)


tests/unit/controllers/start/hello-test.js
```diff
@@ -10,8 +10,28 @@ moduleFor('controller:start/hello', 'Unit | Controller | start/hello', {
   ]
 });
 
-// Replace this with your real tests.
 test('it exists', function(assert) {
+  assert.expect(3);
   let controller = this.subject();
-  assert.ok(controller);
+
+  controller.uploadingImage = true;
```
begedin: Shouldn't we use `set(controller, 'uploadingImage', true);` here?

Same for the rest of the code, really.

tests/unit/controllers/start/hello-test.js
```diff
@@ -10,8 +10,28 @@ moduleFor('controller:start/hello', 'Unit | Controller | start/hello', {
   ]
 });
 
-// Replace this with your real tests.
 test('it exists', function(assert) {
```
begedin: We should rename this test to something more descriptive.

Issue #1126, opened 2017-03-07
Add a /forgot-password route and component


begedin:
# Problem

We need UI to request a password reset email in case the user forgot their password

## Subtasks
- [ ] Add `/forgot-password` route
- [x] Add `components/password/forgot-password`
- [x] Component should have a form with an email field
- [x] On submitting the form, an action should trigger
- [ ] Route should handle action by sending an API `POST /forgot-password` request with the following data: 

```Javascript
{
  email
}
```

- [x] on success, route should redirect to index with a flash message indicating an email has been sent. The message should be vague. We don't want phishing to happen. Something like

"If your email is associated with an account registered here, a password reset link has been sent to it."

# Notes

We can use mirage to do this until https://github.com/code-corps/code-corps-api/issues/498 is done. Otherwise, it's partially blocked by that set of API issues.

I would not consider it blocked by #5, though, since there is nothing preventing us from displaying an unstyled form for now. The UI won't be hooked up with the rest of the app immediately.

JoshSmith:
Thinking we may want to call this `/password/forgot`

Issue #1125, opened 2017-03-07
Add a /reset-password route and component


begedin:
# Problem

We need a UI to perform a password reset action

## Subtasks
- [ ] Add `/reset-password/:token` route
- [x] Add `components/password/reset-password`
- [x] Component should have a form two password fields, for password and password confirmation
- [x] On submitting the form, an action should trigger
- [ ] Route should handle action by sending an API `POST /reset-password` request with the following data: 

```Javascript
{
  value: token // from query param,
  password,
  password_confirmation
}
```

- [x] on success, route should log user in with new credentials and show a flash with a success message
- [ ] on a 422 failure, it should indicate that the passwords do not match

# Notes

We can use mirage to do this until https://github.com/code-corps/code-corps-api/issues/498 is done. Otherwise, it's partially blocked by that set of API issues.

I would not consider it blocked by #5, though, since there is nothing preventing us from displaying an unstyled form for now. The UI won't be hooked up with the rest of the app immediately.

JoshSmith:
Thinking we may want to call this `/password/reset`

snewcomer:
@JoshSmith What were you referring to?  The route name being called `password/reset`?

JoshSmith:
@snewcomer yeah the route name being called `password/reset`. Not sure about API. What do you think?

snewcomer:
@begedin #1231 is pretty close now.  One remaining item is to log in user with new credentials.  How do you think we should go about doing this?

begedin:
During password reset, you have access to the email and the typed in password (and confirmation). At that point, it should be a matter of calling session to authenticate.

You can see how it's done in the login form and do something similar:


```javascript
authenticate() {
  set(this, 'isLoading', true);

  let credentials = this.getProperties('identification', 'password');

  get(this, 'session').authenticate('authenticator:jwt', credentials).catch((reason) => {
    set(this, 'isLoading', false);
    set(this, 'errors', reason.error || reason);
  });
}
```

You can ignore the `isLoading` part for now, as well as error handling. Just get the email and the password the user typed in and call 

```javascript
get(this, 'session').authenticate('authenticator:jwt', { identification: email, password: password });
```

once the reset task is successfully completed.

snewcomer:
@begedin for your review! Almost ready to go.  https://github.com/code-corps/code-corps-ember/pull/1231

PR #1231, opened 2017-04-05
[PASSWORD]: reset and forgot WIP

snewcomer:
## References
Progress on: #1126 #1125

### DIFF
--- a/app/components/password/reset-password.js
+++ b/app/components/password/reset-password.js
@@ -1,0 +1,55 @@
+import Ember from 'ember';
+const { Component, set, get, inject: { service } } = Ember;
+import { task } from 'ember-concurrency';
+
+export default Component.extend({
+  classNames: ['form--centered', 'reset-password-form'],
+
+  /**
+   * @property flashMessages
+   * @type Ember.Service
+   */
+  flashMessages: service(),
+  /**
+   * @property password
+   * @default String
+   */
+  password: '',
+  /**
+   * @property passwordConfirmation
+   * @default String
+   */
+  passwordConfirmation: '',
+  /**
+   * @property error
+   */
+  error: null,
+
+  /**
+   * @property resetPasswordTask
+   * @param password
+   * @param passwordConfirmation
+   */
+  resetPasswordTask: task(function* (password, passwordConfirmation) {
+    try {
+      yield get(this, 'resetPassword')(password, passwordConfirmation);
+      get(this, 'flashMessages').clearMessages().success("Your password has been reset and you're now signed in.");
+      set(this, 'error', null);
+    } catch(e) {
+      set(this, 'error', e);
+    }
+  }),
+
+  actions: {
+
+  /**
+   * @method resetPassword
+   * @param password
+   * @param passwordConfirmation
+   */
+    resetPassword(password, passwordConfirmation) {
+      return get(this, 'resetPasswordTask').perform(password, passwordConfirmation);
+    }
+
+  }
+});
--- a/app/routes/password/reset.js
+++ b/app/routes/password/reset.js
@@ -1,0 +1,34 @@
+import Ember from 'ember';
+const { Route, get, set, inject: { service } } = Ember;
+
+export default Route.extend({
+  queryParams: {
+    token: { refreshModel: true }
+  },
+
+  ajax: service(),
+  session: service(),
+
+  model(params) {
+    return params.token;
+  },
+
+  setupController(controller, model) {
+    set(controller, 'token', model);
+  },
+
+  actions: {
+    resetPassword(password, passwordConfirmation) {
+      return get(this, 'ajax').request('/password/reset', {
+        method: 'POST',
+        data: {
+          token: get(this, 'controller.token'),
+          password,
+          'password-confirmation': passwordConfirmation
+        }
+      }).then((response) => {
+        return get(this, 'session').authenticate('authenticator:jwt', { identification: response.email, password });
+      });
+    }
+  }
+});
--- a/app/templates/components/login-form.hbs
+++ b/app/templates/components/login-form.hbs
@@ -20,6 +20,10 @@
     </button>
   </div>
 
+  {{#link-to 'password.forgot' classNames="t-forgot-password"}}
+    Forgot your password?
+  {{/link-to}}
+
   {{#if errors}}
     {{#each errors.errors as |error|}}
       <p class="error">{{error.detail}}</p>
--- a/app/templates/components/password/forgot-password.hbs
+++ b/app/templates/components/password/forgot-password.hbs
@@ -1,0 +1,23 @@
+<div class="container">
+  <form>
+    <h2>Reset your password</h2>
+
+    <p data-test-id="forgot-password-header">Enter your email and we'll send you a link to reset your password.</p>
+
+    <div class="input-group">
+      {{#auto-focus}}
+        {{input name="email" autocapitalize="off" type="text" value=email}}
+      {{/auto-focus}}
+    </div>
+
+    <div class="input-group">
+      <button id="reset-password" class="default" type="submit" disabled={{isLoading}} {{action "forgotPassword" email}}>
+       Send password reset email
+      </button>
+    </div>
+
+    {{#if error}}
+      {{error-formatter error=error}}
+    {{/if}}
+  </form>
+</div>
--- a/app/templates/components/password/reset-password.hbs
+++ b/app/templates/components/password/reset-password.hbs
@@ -1,0 +1,27 @@
+<div class="container">
+  <form>
+    <h2>Change your password</h2>
+    
+    <div class="input-group">
+      <label data-test-id="password-label">Password</label>
+      {{#auto-focus}}
+        {{input id="password" autofocus="true" class="has-progress" name="password" type="password" autocomplete="off" value=password}}
+      {{/auto-focus}}
+    </div>
+
+    <div class="input-group">
+      <label data-test-id="password-confirmation-label">Confirm password</label>
+      {{input id="password-confirmation" class="has-progress" name="password" type="password" autocomplete="off" value=passwordConfirmation}}
+    </div>
+
+    <div class="input-group">
+      <button id="reset-password" class="default" type="submit" disabled={{resetPasswordTask.isRunning}} {{action "resetPassword" password passwordConfirmation}}>
+       Change password
+      </button>
+    </div>
+
+    {{#if error}}
+      {{error-formatter error=error}}
+    {{/if}}
+  </form>
+</div>
--- a/app/templates/password/reset.hbs
+++ b/app/templates/password/reset.hbs
@@ -1,0 +1,3 @@
+{{password/reset-password 
+  resetPassword=(route-action "resetPassword")
+}}
--- a/mirage/config.js
+++ b/mirage/config.js
@@ -171,6 +171,45 @@ export default function() {
   this.patch('/organizations/:id');
 
   /**
+  * Password
+  */
+
+  // POST /password/forgot
+  this.post('/password/forgot', () => {
+
+    // just return something?
+    return new Mirage.Response(201, {}, {
+      email: 'test@test.com'
+    });
+  });
+
+  this.post('/password/reset', function(schema) {
+    let { password, 'password-confirmation': passwordConfirmation, token } = this.normalizedRequestAttrs();
+
+    let [matchedUser] = schema.users.where({ token }).models;
+
+    if (password === passwordConfirmation) {
+      return new Mirage.Response(201, {}, {
+        email: matchedUser.email,
+        token,
+        user_id: matchedUser.id
+      });
+    } else {
+      let errorDetail = 'Your passwords do not match';
+      return new Mirage.Response(422, {}, {
+        errors: [
+          {
+            id: 'VALIDATION_ERROR',
+            title: '422',
+            detail: errorDetail,
+            status: 422
+          }
+        ]
+      });
+    }
+  });
+
+  /**
   * Previews
   */
 
--- a/tests/acceptance/password-test.js
+++ b/tests/acceptance/password-test.js
@@ -1,0 +1,70 @@
+import { test } from 'qunit';
+import moduleForAcceptance from 'code-corps-ember/tests/helpers/module-for-acceptance';
+import passwordPage from '../pages/password';
+import { getFlashMessageCount } from 'code-corps-ember/tests/helpers/flash-message';
+
+moduleForAcceptance('Acceptance | password test');
+
+test('visiting /password/reset and logging in successfully', function(assert) {
+  let email = 'test@test.com';
+  let password = 'password';
+  let token = 'abc123';
+
+  server.create('user', { email, password, token });
+
+  passwordPage.visitReset({ token });
+
+  andThen(() => {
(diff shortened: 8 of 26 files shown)


app/routes/password/reset.js
```diff
@@ -0,0 +1,20 @@
...
+
+  ajax: service(),
+
+  actions: {
+    resetPassword(password, passwordConfirmation) {
+      return this.get('ajax').request('/password/reset', {
+        method: 'POST',
+        data: {
+          password,
+          passwordConfirmation
```
JoshSmith: Do we need to send the confirmation, as well, or can we just use that on the client-side? I don't believe we have a concept on the server of a confirmation.

app/templates/components/password/forgot-password.hbs
```diff
@@ -0,0 +1,14 @@
+<div class="container">
+  <form>
+
+    {{input name="email" autocapitalize="off" type="text" placeholder='Your email' value=email key-down="keyDown"}}
```
JoshSmith: Above this we can have a `p` that says:

> Enter your email and we'll send you a link to reset your password

If the user's authenticated, this could even be auto-populated with the user's email!

Don't think `keyDown` is needed here.

app/templates/components/password/forgot-password.hbs
```diff
@@ -0,0 +1,14 @@
...
+
+    <div class="input-group">
+      <input class="button default {{if hasError "has-error"}}" name="forgot-password" type="submit" value='Forgot Password' {{action forgotPassword email}} />
```
JoshSmith: Would this or the text `input` have the error?

app/templates/components/password/reset-password.hbs
```diff
@@ -0,0 +1,15 @@
+<div class="container">
+  <form>
+    {{input autofocus="true" class="has-progress" name="password" type="password" placeholder='Enter new password' autocomplete="off" value=password}}
+
+    {{input class="has-progress" name="password" type="password" placeholder='Confirm password' autocomplete="off" value=passwordConfirmation}}
```
JoshSmith: I wouldn't do `placeholder` on either of these.

app/templates/components/password/reset-password.hbs
```diff
@@ -0,0 +1,15 @@
+<div class="container">
+  <form>
+    {{input autofocus="true" class="has-progress" name="password" type="password" placeholder='Enter new password' autocomplete="off" value=password}}
```
JoshSmith: This should probably have a `label` of "Password".

app/templates/components/password/reset-password.hbs
JoshSmith: This should probably have a `label` of "Confirm password".

app/templates/components/password/reset-password.hbs
```diff
@@ -0,0 +1,15 @@
...
+
+    {{input class="has-progress" name="password" type="password" placeholder='Confirm password' autocomplete="off" value=passwordConfirmation}}
+
+    <div class="input-group">
+      <input class="button default {{if hasError "has-error"}}" name="reset-password" type="submit" value='Reset Password' {{action resetPassword password passwordConfirmation}} />
```
JoshSmith: Would change `value` to "Change password".

app/templates/password/reset.hbs
```diff
@@ -0,0 +1,3 @@
+{{password/reset-password 
+  resetPassword=(route-action "resetPassword")
```
JoshSmith: I think `route-action`'s are less testable in component tests, though you can correct me if I'm wrong on this. Also fairly certain that you don't need it here if the action is on the route, and similarly wonder if it shouldn't be in a controller.

app/routes/password/reset.js
snewcomer: @JoshSmith so on the backend we do have a check `validate_confirmation` check 

https://github.com/code-corps/code-corps-api/blob/develop/web/models/user.ex#L101

app/templates/components/password/forgot-password.hbs
snewcomer: I think so.  If the ajax request hit a 500 or something...

app/templates/components/password/reset-password.hbs
snewcomer: I'm guessing this applies to forgot pwd as well?

app/routes/password/reset.js
JoshSmith: Oh, I missed that this was added. My mistake. Is it added to the `user` model in Ember here, too?

app/templates/components/password/forgot-password.hbs
JoshSmith: I'm asking which one would have it. I think the button having it is a little weird, but the text having it makes more sense.

app/templates/components/password/reset-password.hbs
JoshSmith: Yeah, I don't think passwords should typically have a placeholder since it's a little confusing UX.

snewcomer:
@JoshSmith I'm guessing a `Forgot Password` link below the sign-in button?  Text underline, smaller font, etc?

JoshSmith:
@snewcomer I think actually if we can do it just below the password field, that might be ideal. I'd like to have a "Don't have an account? Sign up." below the button on Sign In, and a "Have an account? Log in." below the button on Sign Up.

app/components/password/reset-password.js
```diff
@@ -11,5 +12,35 @@ export default Component.extend({
...
    */
-  passwordConfirmation: ''
+  passwordConfirmation: '',
+  /**
+   * @property error
+   */
+  error: null,
+
+  /**
+   * @property resetPasswordTask
```
snewcomer: @begedin Could you review this file?  Was wondering what you thought of it.  Also, how should I handle errors here?  This displays the error as you mentioned in the slack a while back.  But have no way of getting rid of it.

tests/integration/components/password/reset-password-test.js
```diff
@@ -1,3 +1,5 @@
+import Ember from 'ember';
+const { RSVP } = Ember;
 import { moduleForComponent, test } from 'ember-qunit';
 import hbs from 'htmlbars-inline-precompile';
```
snewcomer: Here is the associated component test.

app/templates/components/login-form.hbs
```diff
@@ -9,6 +9,9 @@
   <div class="input-group">
     {{input name="password" id="password" placeholder="Password" type="password" value=password}}
   </div>
+  {{#link-to 'password.forgot' classNames="t-forgot-password"}}
```
snewcomer: @JoshSmith this forgot-password link could use some styling when you get a chance.

app/templates/components/password/forgot-password.hbs
```diff
@@ -1,13 +1,13 @@
 <div class="container">
   <form>
 
-    <p>Enter your email and we'll send you a link to reset your password</p>
+    <p data-test-id="forgot-password-header">Enter your email and we'll send you a link to reset your password</p>
```
snewcomer: data test ids if you want to use them to target for tests.  Thoughts?

app/components/password/reset-password.js
begedin: For clearing errors, I would just clear it in the `yield`. As in, when you start the task again, clear the errors from the previous task.

app/components/password/reset-password.js
```diff
@@ -0,0 +1,46 @@
...
+   * @param password
+   * @param passwordConfirmation
+   */
+  resetPasswordTask: task(function* (password, passwordConfirmation) {
+    try {
+      yield get(this, 'resetPassword')(password, passwordConfirmation);
+    } catch(e) {
+      set(this, 'error', e);
+    }
+  }),
```
begedin: Will `get(this, 'resetPassword')` actually return the function here? Isn't it defined under `this.actions.resetPassword`?

app/components/password/reset-password.js
```diff
@@ -0,0 +1,46 @@
...
+  error: null,
+
+  /**
+   * @property resetPasswordTask
+   * @param password
+   * @param passwordConfirmation
+   */
+  resetPasswordTask: task(function* (password, passwordConfirmation) {
+    try {
+      yield get(this, 'resetPassword')(password, passwordConfirmation);
```
begedin: Will `get(this, 'resetPassword')` actually return the function here? Isn't it defined under `this.actions.resetPassword`?

begedin: Also, this is where I would clear the error. Simply do a `set(this, 'error', null)` before the `yield`.

app/templates/password/reset.hbs
snewcomer: So I believe route-action is for discoverability.  If I change to action, it errors.  I can create a controller for this action if you would like.

app/components/password/reset-password.js
snewcomer: Yeah to it is passed in as closure action, so just need to get it off the props and call that function.  This way also allows me to patch that action in the integration test.

app/routes/password/reset.js
```diff
@@ -0,0 +1,19 @@
...
+
+  actions: {
+    resetPassword(password, passwordConfirmation) {
+      return get(this, 'ajax').request('/password/reset', {
+        method: 'POST',
+        data: {
+          password,
+          passwordConfirmation
+        }
+      });
```
snewcomer: Need to handle success/error cases still.

app/templates/password/reset.hbs
begedin: This is at a route level, so no need to generate a controller just for the sake of not using `route-action`, I think. The component is just as easily testable this way as it would be with a controller, but the boilerplate is reduced.

Also, thank you for that input, @snewcomer. I was in a mental place where I wasn't sure how route actions are useful exactly, since, in my experience, using them in nested components makes everything harder to test.

This example is where using a route action is useful for reducing boilerplate with no (to me) apparent negatives.

tests/pages/password.js
```diff
@@ -0,0 +1,16 @@
...
+import errorFormatter from './components/error-formatter';
+
+export default create({
+  visitReset: visitable('/password/reset'),
+  visitForgot: visitable('/password/forgot'),
+
+  form,
+  resetPasswordForm,
+  errorFormatter
+});
```
begedin: I believe a page object should match with a route 1 to 1. Otherwise, we risk confusing people. Is there any particular reason why you chose to merge the two password pages?

Split up, we would not have `visitReset` and `visitForgot`. The two pages would simply have a `visit`, a `form` and an `errorFormatter`.

snewcomer: Yeah I see your point.  Not really.  Only because I have an `acceptance/password-test`.  So sounds like I should split them up?

snewcomer: @begedin ^^

snewcomer:
@JoshSmith all checks have passed.  Probably a good idea to style some pieces here.  I can tackle, but didn't know if you wanted to do that.

mirage/config.js
```diff
@@ -185,8 +185,9 @@ export default function() {
 
   this.post('/password/reset', (_schema, request) => {
     let params = request.requestBody;
-    let password = params.split('&')[0].split('=')[1];
-    let passwordConfirmation = params.split('&')[1].split('=')[1];
+    let [first, second] = params.split('&');
+    let password = first.split('=')[1];
+    let passwordConfirmation = second.split('=')[1];
```
begedin: There should be a `this._getAttrsForFormRequest()` which returns a POJO containing the form data, doing this job for you. It's undocumented, but it is supported by mirage.

However, you are using an arrow function, so you do not have access to `this`.

I believe something like
```Javascript
this.post('/password/reset', function() {
  let { password, passwordConfirmation } = this._getAttrsForFormRequest();

  // continue with the rest of your code here
));
```
ought to do the trick.

snewcomer: Awesome trick!

snewcomer:
@JoshSmith fyi this is failing only on phantom.  That is why I put in PR #1253

app/routes/password/reset.js
```diff
@@ -0,0 +1,22 @@
+import Ember from 'ember';
+const { Route, get, inject: { service } } = Ember;
+
+export default Route.extend({
+
+  ajax: service(),
+  session: service(),
+
```
snewcomer: One thing I need here is a model hook that extracts the token from `params` and used to send up with the POST request.

begedin: To get the same functionality you proposed by changing the route in API, you could do something like:

```Javascript
queryParams: {
  token: { refreshModel: true }
}
```

If you add that code into the route, the `params` hash in the model hook will contain a `token` key.

The standard way to specify these query parameters which are not part of the dynamic route URL would be to specify them in the controller:

```
  queryParams: ['token']
});
```

but the way I described above allows us to skip unnecessarily defining a controller file.

snewcomer:
@JoshSmith @begedin So I tried this exact same code out on a simple app I just spun up and it works.  Just need the email created part.

app/templates/components/password/forgot-password.hbs
begedin: We've been discussing the usage of https://github.com/simplabs/ember-test-selectors in the past, but I'm nut sure where we're with it now. The addon makes it so these test-only attributes are not present in the final build.

I think, let's not introduce it now, but what you could do is create an issue where we can discuss it and, if we opt into it, create a bunch of other issues. It's a tasks that would be very suitable for newbie volunteer.

app/routes/password/reset.js
```diff
@@ -24,10 +24,10 @@ export default Route.extend({
         data: {
           token: get(this, 'controller.token'),
           password,
-          passwordConfirmation
+          "password-confirmation": passwordConfirmation
```
snewcomer: Curious about this change.  Is this a stylistic thing?  Just asking for future ref.

tests/acceptance/password-test.js
```diff
@@ -28,19 +31,22 @@ test('visiting /password/reset and logging in successfully', function(assert) {
...
 
-  server.create('user', { email, password });
+  server.create('user', { email, password, token });
 
-  passwordPage.visitReset();
+  passwordPage.visitReset({ token });
 
   andThen(() => {
-    assert.equal(currentURL(), '/password/reset');
+    assert.equal(currentURL(), '/password/reset?token=abc123');
```
snewcomer: ah yep, I forgot about those changes.

tests/acceptance/password-test.js
```diff
@@ -53,7 +59,9 @@ test('visiting /password/forgot', function(assert) {
     assert.equal(currentURL(), '/password/forgot');
   });
 
-  passwordPage.forgotPasswordForm.sendForgotPasswordSuccessfully('admin@gmail.com');
+  andThen(() => {
```
snewcomer: surprised this has to be wrapped in an `andThen`.

JoshSmith: It's because the Mirage response is a promise and we don't know when it will return.

app/routes/password/reset.js
JoshSmith: It's because most of the keys in JSON API are dasherized by default. There's some lack of consistency in some places at present that are not standard CRUD (like here), but would like to bring at least that consistency to bear here, even if the resource itself isn't a JSON API resource.

PR #1250, opened 2017-04-12
Add tests for identity document file upload

begedin:
# What's in this PR?

We've never gotten around to testing file uploads properly, so I dedicated some time to finally do this in the case of the identity document for account activation.

All my research tells me it's not possible to mock the file select user action or the event. It might be possible to mock a file drop, but this does not apply here.

Instead, I did the next best thing and found a way to call the bottom-most EmberUploader trigger function, which is the `change` event handler on the `EmberUploader.FileField` component. This one is well tested as a unit test in the addon itself and is a basic Ember event handling hook, so it's best I could think off.

I've wrapped this into a page object and it should be relatively easy to make it reusable for any other integration test, as well as acceptance tests, if necessary.


JoshSmith:
Looks good! Was going to review this today, but you beat me to it.

PR #1268, opened 2017-04-25
Create GitHub connect button component #1263

### LINKED ISSUE #1263: Create GitHub connect button component
# Problem

Create a component which is simply an anchor tag, styled as a button, with the `href` attribute being a computed property in the format:

```
`https://github.com/login/oauth/authorize?scope=${get(this, 'scope')}&client_id=${get(this, 'clientId'}&state=${get(this, 'state')}&redirect_uri=${get(this, 'redirectUri')}`
```

The properties composing the URL can initially be defined as 

```
clientId: null,
scope: null,
state: null,
redirectUrl: null,
```

## Subtasks
- [x] Create component, ensure proper tag, and proper attribute binding to href
- [x] Write integration tests for component, using ember cli page objects, which ensure the following
  - component is an `a` tag
  - for various combinations of `clientId, scope, state, redirectUri` bound to the templates, the `href` attribute is properly bound to the component

terrancebryant:
# What's in this PR?

- Github connect button

- button url takes in `scope`, `state`, `cliendId`, `redirectUri`

- component is a anchor tag styled like a button

## References
Fixes #1263 

Progress on: #

### DIFF
--- a/app/components/github-connect.js
+++ b/app/components/github-connect.js
@@ -1,0 +1,40 @@
+import Ember from 'ember';
+
+const {
+  Component,
+  set
+} = Ember;
+const baseUrl = 'https://github.com/login/oauth/authorize';
+
+/**
+  The github-connect component is used to to connect to github
+
+  ## default usage
+
+  ```handlebars
+    {{github-connect clientId=clientId scope=scope state=state redirectUri=redirectUri}}
+  ```
+
+  @module Component
+  @class github-connect
+  @extends Ember.Component
+  @public
+ */
+
+export default Component.extend({
+  tagName: 'a',
+  classNames: ['button', 'default'],
+  attributeBindings: ['url:href'],
+
+  clientId: null,
+  scope: null,
+  state: null,
+  redirectUri: null,
+
+  init() {
+    this._super(...arguments);
+    let { clientId, scope, state, redirectUri } = this.getProperties('clientId', 'scope', 'state', 'redirectUri');
+    set(this, 'url', `${baseUrl}?scope=${scope}&client_id=${clientId}&state=${state}&redirect_uri=${redirectUri}`);
+  }
+
+});
--- a/tests/integration/components/github-connect-test.js
+++ b/tests/integration/components/github-connect-test.js
@@ -1,0 +1,29 @@
+import Ember from 'ember';
+import { moduleForComponent, test } from 'ember-qunit';
+import hbs from 'htmlbars-inline-precompile';
+import PageObject from 'ember-cli-page-object';
+import component from 'code-corps-ember/tests/pages/components/github-connect';
+
+const { set } = Ember;
+const baseUrl = 'https://github.com/login/oauth/authorize';
+let page = PageObject.create(component);
+
+moduleForComponent('github-connect', 'Integration | Component | github connect', {
+  integration: true,
+  beforeEach() {
+    page.setContext(this);
+  },
+  afterEach() {
+    page.removeContext();
+  }
+});
+
+test('check if properties are bound correctly', function(assert) {
+  set(this, 'scope', '3098379083');
+  set(this, 'clientId', 'ace');
+  set(this, 'state', 'ace123');
+  set(this, 'redirectUri', 'ace.com');
+  page.render(hbs`{{github-connect scope=scope clientId=clientId state=state redirectUri=redirectUri}}`);
+  assert.equal(page.button.href, `${baseUrl}?scope=3098379083&client_id=ace&state=ace123&redirect_uri=ace.com`);
+});
+
--- a/app/templates/components/github-connect.hbs
+++ b/app/templates/components/github-connect.hbs
@@ -1,0 +1,1 @@
+GitHub Connect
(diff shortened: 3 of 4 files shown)


tests/integration/components/github-connect-test.js
```diff
@@ -0,0 +1,42 @@
...
+    page.setContext(this);
+  },
+  afterEach() {
+    page.removeContext();
+  }
+});
+
+test('check if scope is bound correctly', function(assert) {
+  set(this, 'scope', '3098379083');
+  this.render(hbs`{{github-connect scope=scope}}`);
```
snewcomer: @begedin can any one of the four properties be null at any one time?

tests/integration/components/github-connect-test.js
```diff
@@ -0,0 +1,42 @@
+import Ember from 'ember';
+import { moduleForComponent, test } from 'ember-qunit';
+import hbs from 'htmlbars-inline-precompile';
+import PageObject from 'ember-cli-page-object';
+import component from 'code-corps-ember/tests/pages/components/github-connect';
+
+const { set } = Ember;
+let page = PageObject.create(component);
+
+moduleForComponent('github-connect', 'Integration | Component | github connect', {
```
snewcomer: @terrancebryant can also assert the button contains text.

app/components/github-connect.js
```diff
@@ -0,0 +1,37 @@
...
+  tagName: 'a',
+  classNames: ['button', 'default'],
+  attributeBindings: ['url:href'],
+
+  clientId: null,
+  scope: null,
+  state: null,
+  redirectUri: null,
+
+  url: computed('clientId', 'scope', 'state', 'redirectUrl', function() {
```
snewcomer: @terrancebryant prob can setup the url on init of the component rather than a computed.

app/components/github-connect.js
```diff
@@ -0,0 +1,37 @@
...
+    return `https://github.com/login/oauth/authorize?scope=${get(this, 'scope')}&client_id=${get(this, 'clientId')}&state=${get(this, 'state')}&redirect_uri=${get(this, 'redirectUri')}`;
```
snewcomer: @begedin should this be in a config file and insert required parts into the string?

app/components/github-connect.js
begedin: @snewcomer Good idea! A bit cleaner and we are not expecting it to change anyway.

app/components/github-connect.js
begedin: The URL itself is something that's not going to change and should not vary across different parts of the app, so config doesn't really help us. It would be cleaner and clearer to declare it as a `const`, though, at the top of the modules, outside the definition of the component.

tests/integration/components/github-connect-test.js
begedin: They should not be in a real life use case, but there is little reason to enforce that. 

I would, however, suggest not to write as many tests. A single test which checks various combinations of property values ought to be good enough.

app/components/github-connect.js
terrancebryant: @snewcomer can you give an example of setting the url on init ?

snewcomer: @terrancebryant yeah so it would be something like 

```
init() {
  this._super(...arguments);
  let { clientId, scope, state, redirectUrl } = this.getProperties('clientId', 'scope', 'state', 'redirectUrl');
  set(this, 'url', `${url}?scope=${scope}&client_id=${clientId}....`)
}
```

begedin:
```Javascript
init() {
  this._super(...arguments);
  let { baseUrl, clientId, scope, state, redirectUrl } = this.getProperties('baseUrl', 'clientId', 'scope', 'state', 'redirectUrl');
  set(this, 'url', `${baseUrl}?scope=${scope}&client_id=${clientId}....`)
}
```

where `baseUrl` is a `const` declared at the top level in this module.

terrancebryant:
@begedin @snewcomer i've made the changes you guys asked for take a look when you get time.

app/components/github-connect.js
```diff
@@ -0,0 +1,35 @@
...
+
+export default Component.extend({
+  tagName: 'a',
+  classNames: ['button', 'default'],
+  attributeBindings: ['url:href'],
+
+  init() {
+    this._super(...arguments);
+    this.getProperties('clientId', 'scope', 'state', 'redirectUri');
+    set(this, 'url', `${baseUrl}?scope=${get(this, 'scope')}&client_id=${get(this, 'clientId')}&state=${get(this, 'state')}&redirect_uri=${get(this, 'redirectUri')}`);
```
begedin: This would be written as 

```javascript
let { clientId, scope, state, redirectUri } = getProperties(this, 'clientId', 'scope', 'state', 'redirectUri');
set(this, 'url', `${baseUrl}?scope=${scope)}&client_id=${clientId}&state=${state}&redirect_uri=${redirectUri}`);
```

Also, would you mind declaring these properties as default nulls in the component:

```Javascript
...
attributeBindings: ['url:href'],

clientId: null,
redirectUri: null,
scope: null,
state: null,
...
```

We like to do that, so that it's at a glance clear which properties the component relies on.

terrancebryant: @begedin using your example i get these warnings 
screen shot 2017-04-27 at 5 28 53 pm

PR #1270, opened 2017-04-29
Add github route for final step of github connect

### LINKED ISSUE #1264: Add route to handle successful github authentication
# Problem

Component created in #1263 will be a link. Clicking this link will send the user to github, where they will be able to, with their signed in github account, authorize codecoprs to connect. Once that's done, the user will be redirected back to codecorps, to a route we specified.

This route should be `/github/`

In the model hook, the route should expect a `code` query parameter. The model hook should post this `code` parameter to our API and expect an OK response of some sort.

```
POST { code: code }
```

_In the background, the API will `POST` the code to github, along with a `CLIENT_ID` and a `CLIENT_SECRET` and receive back and `auth_token` to store and associate with the current user. The API should reply to the client with a JSON containing the updated current user_

Once that's done, and the API responds, in `afterModel`, we should push the user response into the store and set it into our session service as `currentUser`. The user should then redirect to the profile route.

## Subtasks
- [ ] Add route with `model` and `afterModel` hooks
- [ ] Add acceptance test for the successful flow. Use mirage to mock a server response. Test should ensure:
  - [ ] route posts to server with proper code
  - [ ] current user is properly updated in the schema
  - [ ] flash message is displayed

## References

NOTE: This is not blocked by #1263. It can be worked on in paralel.

begedin:
# What's in this PR?

Closes #1264 

This adds a `/github` route to the application. It accepts a `code` parameter and, as part of the model hook, posts that code to the API, where we presumably use the code to get an auth token and assign it to the current user, which is returned as a response. 

In case of a success, we redirect to `settings.profile`. In case of an error, it's `project-list`, with a flash error.

## References
Fixes #1264


app/routes/github.js
```diff
@@ -0,0 +1,42 @@
...
+  currentUser: service(),
+  flashMessages:service(),
+  store: service(),
+
+  model({ code }) {
+    return this._sendRequest(code);
+  },
+
+  afterModel(currentUserData) {
+    get(this, 'store').pushPayload(currentUserData);
```
begedin: There is no need to update the currentUser in the service as I said. This ought to find the existing user record in the store and update it with new information.

Besides, the next line transitions us to `settings.profile`, which explicitly re-fetches the current user from the API.

app/routes/github.js
```diff
@@ -0,0 +1,42 @@
...
+const {
+  get,
+  inject: { service },
+  Route
+} = Ember;
+
+const CODE_INVALID = 'The GitHub authorization code is invalid.';
+
+export default Route.extend(AuthenticatedRouteMixin, {
+  queryParams: { code: { refreshModel: true } },
```
snewcomer: @begedin where is this query param used?

npendery: yeah if I am understanding this right, down in the model hook we are extracting `code` from the `params` argument and not from the `queryParams`. is that we need the `refreshModel` for something?

begedin: The `params` argument in the model hook is built using info provided in the `queryParams` hash defined in the route. It can alternatively be defined in the controller to, but either way, without `code` specified here, the first argument in the model hook will not contain a `code` key.

The second argument, which is a transition, will somewhere, deeply nested, contain a code key, but not the first argument.

begedin: You can see a lengthy bit of code that shows most of the inner workings here https://github.com/emberjs/ember.js/blob/v2.12.0/packages/ember-routing/lib/system/route.js#L171

PR #1282, opened 2017-05-03
Add github-state service for generating and validating the connect state parameter

### LINKED ISSUE #1265: Write module which generates a random string
# Problem

- #1263 Gives us a component which navigates to github for authentication
- #1264 Gives us a route which handles a successful authentication and finalizes the flow using our API

To prevent cross site forgery, we need to add a `state` parameter to the link in #1263. This parameter is then returned by github and hits our route in #1264 as a query param. We will then be able to check if the one we sent and the one we received back are identical.

We need a service module which generates this string and somehow persists it across sessions, associated to the signed in user.

**We should discuss and decide on the persisting method.** It could be local storage, or potentially an API request (meaning the API would have an action which generates the scope and somehow associates it with the user, possibly even on a DB level). The API option would certainly be safer.

## Subtasks
- [ ] Write `github-state` service:
  - [ ] Internally dependant on the `current-user` service
  - [ ] has `generate` function crates a persisted random string tied to the current user. 
  - [ ] has a `validate(state)` function which takes the parameter and compares it to the stored state tied to the current user. Returns true if the two are identical, false otherwise.
- [ ] Write unit tests for the service, ensuring the two methods work. Use mirage/pretender to mock server requests if there are any. If we opt for a localStorage/cookie based approach, then we need to figure out ways to mock that.

## References

Dependant on #1263 and #1264, but not blocked by them. Can be worked on in parallel.

begedin:
Closes #1265

# What's in this PR?

Instead of being dependant on current user, as stated in the issue, the new service simply injects the session service internally. The session service has a `set` method which persists data in the same way the authentication data is persisted, meaning it stays across browser tabs/windows, which is exactly what we need.


app/services/github-state.js
```diff
@@ -0,0 +1,73 @@
+import Ember from 'ember';
+
+const { get, Service, inject: { service } } = Ember;
+
+function dec2hex(dec) {
```
JoshSmith: This is a bit confusing as the `2` could mean digits, although I know this is a conversion function. Maybe a rename.

app/services/github-state.js
```diff
@@ -0,0 +1,73 @@
...
+  return (`0${dec.toString(16)}`.substr(-2));
+}
+
+function generateRandomString(len) {
+  let arr = new window.Uint8Array((len || 40) / 2);
```
JoshSmith: An example of why the above is potentially confusing from JS itself! The `8` here is what I mean. Similar to `2` above.

PR #1283, opened 2017-05-03
Implement csf protection

### LINKED ISSUE #1266: Implement cross site forgery protection
# Problem

Once #1263, #1264 and #1265 are done, we should implement cross site forgery protection.

## Subtasks
- [ ] The component in #1263 should set it's `state` property from the #1265 `github-state` service `generate()` function.
- [ ] The `model` hook  in the route from #1264 should check the `state` using the `validate(state)` function, before posting the `code`. If the check fails, it should redirect to root, with an error flash saying something went wrong.
- [ ] There should be an addition to the integration test for the `github-connect-link` component. The addition should test that the `href` contains the `proper` state when mocking the `github-state` service.
- [ ] There should be an additional acceptance test for the route in #1264, which ensures proper redirection happens when `state` validation fails. Again, we can mock the `github-state` service for this.

## References

Blocked by: #1263, #1264, #1265

begedin:
# What's in this PR?

This PR builds on top of #1282 to fix #1266 

It makes use of the service provided by #1282 to first generate and store a state value to send to github connect, followed by validating that value when the user comes back from github.

Right now, the PR is submitted against #1282 instead of develop, so the changes are clearer to see, but we should switch the base to develop when #1282 is merged.

### DIFF
--- a/app/routes/github.js
+++ b/app/routes/github.js
@@ -8,17 +8,31 @@ const {
 } = Ember;
 
 const CODE_INVALID = 'The GitHub authorization code is invalid.';
+const STATE_INVALID = 'Something went wrong while connecting to GitHub. Please try again.';
 
 export default Route.extend(AuthenticatedRouteMixin, {
-  queryParams: { code: { refreshModel: true } },
+  queryParams: {
+    // changing any of these query params needs to trigger the model hook, to
+    // start the connect process
+    code: { refreshModel: true },
+    state: { refreshModel: true }
+  },
 
   ajax: service(),
   currentUser: service(),
+  githubState: service(),
   flashMessages:service(),
   store: service(),
 
-  model({ code }) {
-    return this._sendRequest(code);
+  model({ code, state }) {
+    let stateValidator = get(this, 'githubState');
+
+    if (stateValidator.validate(state)) {
+      return this._sendRequest(code);
+    } else {
+      get(this, 'flashMessages').clearMessages().danger(STATE_INVALID);
+      return this.transitionTo('settings.profile');
+    }
   },
 
   afterModel(currentUserData) {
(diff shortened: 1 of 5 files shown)


app/routes/github.js
```diff
@@ -8,17 +8,29 @@ const {
 } = Ember;
 
 const CODE_INVALID = 'The GitHub authorization code is invalid.';
+const STATE_INVALID = 'Something went wrong in the connect process. Please try again';
```
JoshSmith: > Something went wrong while connecting to GitHub. Please try again.

app/routes/github.js
```diff
@@ -8,17 +8,29 @@ const {
...
 
 export default Route.extend(AuthenticatedRouteMixin, {
-  queryParams: { code: { refreshModel: true } },
+  queryParams: {
+    code: { refreshModel: true },
+    state: { refreshModel: true }
```
JoshSmith: These could use some inline commenting explaining why we use `refreshModel` here.

app/routes/github.js
```diff
@@ -8,17 +8,29 @@ const {
...
-  model({ code }) {
-    return this._sendRequest(code);
+  model({ code, state }) {
+    let stateValidator = get(this, 'githubState');
+
+    if (stateValidator.validate(state)) {
+      return this._sendRequest(code);
+    } else {
+      get(this, 'flashMessages').clearMessages().danger(STATE_INVALID);
+      return this.transitionTo('projects-list');
```
JoshSmith: Why are we transitioning to `projects-list` here? Is there a better place to transition to?

begedin: Rethinking about it, `settings.profile` makes more sense for all of these cases. Either the `code` is invalid, or the `state` is invalid. Either way, the user should be back on their profile page, where, presumably, the github connect button is going to be.

PR #1293, opened 2017-05-08
chore(package): update ember-cli to version 2.13.1

begedin:
Closes #1259

### DIFF
--- a/package.json
+++ b/package.json
@@ -20,7 +20,7 @@
     "ember-auto-focus": "1.0.6",
     "ember-autoresize": "0.5.22",
     "ember-can": "^0.8.4",
-    "ember-cli": "2.12.1",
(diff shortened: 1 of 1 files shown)


begedin:
There seem to be some bugs with the underlying babel transpilation in ember-cli at the moment, which is under ongoing investigation.

https://github.com/ember-cli/ember-cli/issues/7001

I'm not sure we should be spending time dealing with the update here until that is resolved or we get more information. Our APP seems to be misbehaving in a similar fashion when we update.

jrr:
@begedin you can update to 2.13 and hold back Babel; just leave `ember-cli-babel` and `ember-cli-htmlbars-inline-precompile` at their previous versions. (this is what we did on my current project)

Issue #1276, opened 2017-05-02
Add github service, with getIssues function


begedin:
# Problem

We need a github service, which provides data for github integration using the current user's auth token. 2 functions are needed, `getRepositories`, and `getIssues`.

This issue handles the `getIssues` function.

The function should accept a project as a parameter.

The function should use the project's githubId, which is a repository ID to return all open issues for that repository from the github API and return them as a collection in the format `{ issueName, githubId }`.

## Subtasks
- [ ] Add `/services/github.js` if it doesn't already exist
- [ ] Add `getIssues` with described behavior to the service
- [ ] Add unit tests for the function

begedin:
This is now obsolete

PR #1294, opened 2017-05-09
Add github-service with getRepositories, getIssues(project)

### LINKED ISSUE #1275: Add github service, with getRepositories function
# Problem

We need a github service, which provides data for github integration using the current user's auth token. 2 functions are needed, `getRepositories`, and `getIssues`.

This issue handles the `getRepositories` function.

The function should either accept a current user as a parameter, or have the current user service as an injected dependency, to get it that way.

The function should return all repositories from the github API and return them as a collection in the format `{ repositoryName, githubId }`.

## Subtasks
- [ ] Add `/services/github.js` if it doesn't already exist
- [ ] Add `getRepositories` with described behavior to the service
- [ ] Add unit tests for the function

begedin:
# What's in this PR?

This PR adds a `github-service`, which is a client to be used when communicating with the github API.

The service defines two public methods.

### `getRepositories`

This method uses the `githubAuthToken` for the currently signed in user and makes a request to retrieve a list of repositories for that user. It maps the retrieved data and returns a promise.

### `getIssues(project)`

This method uses the project's `githubId`, which is github repository ID to retrieve a list of issues for that repository. 

Due to the fact that we only have access to the github repository id, we are forced to make use of an undocumented github api endpoint `GET /repositories/:id`, to request a repository first, followed by a request `GET /:owner/:repo/issues`. 

If were to be storing the `owner.login` and `repo.name` values in some association to our project, we would not require the first request and would be able to request issues directly. 

That being said, regardless of how we want to proceed, this service should be a good starting point from which we can build.

## References
Fixes #1275, #1276

### DIFF
--- a/app/models/user.js
+++ b/app/models/user.js
@@ -8,9 +8,10 @@ const { computed } = Ember;
 export default Model.extend({
   biography: attr(),
   cloudinaryPublicId: attr(),
-  insertedAt: attr('date'),
   email: attr(),
   firstName: attr(),
+  githubAuthToken: attr(),
+  insertedAt: attr('date'),
   lastName: attr(),
   name: attr(),
   password: attr(),
--- a/app/services/github-service.js
+++ b/app/services/github-service.js
@@ -1,0 +1,136 @@
+import AjaxService from 'ember-ajax/services/ajax';
+import Ember from 'ember';
+
+const {
+  computed,
+  computed: { alias },
+  get,
+  inject: { service }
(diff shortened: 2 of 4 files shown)


JoshSmith:
@begedin what's the status of this?

begedin:
This is very likely obsolete and closable, but I wanna wait until we merge #1295

begedin:
Closing due to being updated and implemented differently

PR #1295, opened 2017-05-10
Added github-repo-select component

### LINKED ISSUE #1273: Add github-repository-select component
# Problem

We need to add a component to the project, which allows selecting a github repo when creating a task.

If a github repo is selected, that task is then created alongside an issue on github and the two are then kept in sync by our backend.

## Subtasks
- [x] Add `githubRepo` relationship to task
- [x] Create component
- [x] Add integration tests for the component
- [x] Add component to `tasks.new` route
- [x] Add acceptance test to ensure component works in the route (task is saved associated to github repo)

### LINKED ISSUE #1278: Add github-issue-select component to the project.tasks.new page
# Problem

A user, when creating a task, if the project is connected to github, should be able to assign a github issue to the task.

To that purpose, we should add the `github-issue-select` component to the the `project.tasks.new` page, if the conditions are met:

  - project is connected to github (it has a github id)

The route should provide the data to the component, but this should be done through a service implemented in #1276 

When the component triggers the select action, the task should be assigned the github id.

## Subtasks
- [ ] Render `{{github-issue-select}}`
- [ ] Select data should be provided by service from #1276
- [ ] When action is sent by the component, update unsaved task with github id.
- [ ] Add acceptance test for the task creation process where a github issue was selected

begedin:
Closes #1273, closes #1278 

# What's in this PR?

This PR adds a `github-repository-select` component.

Instead of using `ember-x-select` as described in the issue, it instead uses `ember-power-select` as the underlying component.

The component accepts a `githubRepos` property, which can be a collection or a promise. It renders options in the repository property and triggers an action which passes the selected option when the selection changes.

This component is then rendered in the `tasks.new` route, so, when creating a task, we can optionally select one of the parent projects github repos to create an associated github issue in. At that point, the task and the issue will be kept in sync by our backend.

Note that the original issue was advocating putting this on a project settings tab, due to the incorrect assumption that a single project will be linked to a single github repo.

## Additional notes

When creating a task associated to a github repo, the API presumably saves that association. Additionally, the API creates an issue for that github repo on github and assigns the issue number to the created task.

That means that created task will have
- a `github_repo_id`, and through it, a github repo name and account (`account/project-name` naming scheme which github uses)
- a `github_issue_number`

These two, or these three, if we go with `account/name/number` are what we need to properly identify an issue on github, so this is what then gets used to perform syncing in the background.

However, the github issue number by itself is not necessary for the ember client to understand that the task is synced. All the information we need is the fact that there's a `githubRepo` association, so the model does not require a `githubIssueNumber` attribute.

Because of that, we assume that the api serializer does not serialize `github_issue_number` either.

## References

Fixes #1273, #1278


JoshSmith:
Is this still relevant and useful @begedin as-is, or does it need a lot of work?

begedin:
@JoshSmith This is absolutely relevant and probably the onlly one that still actually is.

It adds selection of a repo when creating a task and is the MVP for task sync.

The idea is, when creating a task, the user is offered the list of repos the project is connected to in a dropdown. If a repo is selected, then that task is synced to that repo from that point on.

It's tied to the API PR code-corps/code-corps-api#805

filipecrosk:
It looks good to me 👍

JoshSmith:
@begedin has this changed at all due to the API?

begedin:
I'll make a second pass, but I'm reasonably sure any API changes have already been done. There may be a need for an additional PR once we add a user scope for github app installations on the API side, but I'm not sure yet.

JoshSmith:
Now that I'm looking at this, I'm not sure I understand why this uses power select vs a normal select. The UI would appear to be a lot more straightforward to me if it were a standard select element, where one of the options is simply "None".

Can you explain more about your reasoning @begedin?

begedin:
My assumption was we want to consolidate our selects into one, so I went with what we were already using. There was no reasoning beyond that. 

If we want to go with a simple select, we should weigh the value of switching it out now vs merging and creating a task for it.

A regular select still needs proper ember bindings with ddau, but yeah, it would definitely be a much simpler piece of ui.

begedin:
@JoshSmith I implemented the behavior for the new component and removed the old component.

The tests gave me trouble because the selection action for x-select triggers multiple times, so I had to do some "creative writing" to get tests that assert the desired behavior is there. Basically, instead of testing the action is being triggered, i render the component in the exact way we use it in our app and test the desired property is being set upon selection.

begedin:
Note that some basic styling for the select is still needed, because it does not expand to fill width of it's container.

JoshSmith:
This looks great. I'll just do the styling and then will approve/merge.

PR #1296, opened 2017-05-10
Added github-issue-select component

### LINKED ISSUE #1274: Add github-issue-select component
# Problem

Component provides a dropdown using ember-x-select. The collection to render as select options is provided from outside, by binding to the template. Examples of using ember-x-select
are in /components/select. Options should assume the format { issueName , githubId } where the name is the rendered text of the dropdown, while the id is the value.

When a user selects an option
- they need to be asked to confirm their selection
- once confirmed, the value, which is the githubId, should be sent up as an action from the component
- if the user does not confirm, the previous option should be selected again

This is similar to #1273, so if unsure how to proceed, refer to that issue's PR, if it exists.

## Subtasks
- [ ] Create component
- [ ] Add integration tests for the component
   - renders options properly
   - when selecting and confirming, githubId is sent as action parameter
   - when selecting and canceling, nothing happens

begedin:
# What's in this PR?

This PR adds a `github-issue-select` component, to be used on a task edit page, to connect an unconnected task with a github issue.

The implementation requires some fixes in #1275 so it branches off of that branch and requires that PR to be merged first.

The implemented behavior is near-identical to that of `github-repository-select` implemented in #1275. 

We need to provide an `issues` collection, which can be a promise, as well as a `task` model to the component. 

If the task has no `githubId`, the dropdown will render placeholder text and allow selecting an issue. Upon selecting an issue, the component will send out an action with the selected issue to the parent.

If the task has a `githubId`, the issue with that github id will be rendered as selected, and the dropdown will be disabled, since we ought to not support re-connecting in the MVP.

## References
Fixes #1274
Requires merging of #1275 
Requires changing PR base to develop once #1275 is merged.


tests/integration/components/github-issue-select-test.js
```diff
@@ -0,0 +1,100 @@
...
+    set(this, 'issues', []);
+    set(this, 'onIssueSelected', () => {});
+    renderPage();
+  },
+  afterEach() {
+    page.removeContext();
+  }
+});
+
+function renderPage() {
```
snewcomer: Would it make sense to just inline `page.render` in the beforeEach?

tests/pages/components/github-issue-select.js
```diff
@@ -0,0 +1,37 @@
+import { attribute, findElementWithAssert, collection } from 'ember-cli-page-object';
+import { clickTrigger, nativeMouseUp } from 'code-corps-ember/tests/helpers/ember-power-select';
+
+export default {
+  scope: '.github-issue-select',
+
+  openDropdown() {
```
snewcomer: Does it potentially make sense to utilize the `pages/components/power-select` file and maybe make it more dynamic (ie. pass `.github-issue-select`)?  Looks like somehow the `trigger.open` in the PS page file could take `this.scope` from this file?

app/templates/components/github-issue-select.hbs
```diff
@@ -0,0 +1,11 @@
+{{#power-select
+  disabled=task.githubId
+  loadingMessage='Retrieving issues...'
+  options=issues
+  onchange=onIssueSelected
+  placeholder='Connect your task with a GitHub issue'
+  renderInPlace=true
```
begedin: Not all of them are. The one for task assignment is causing layout problems if we render in place, due to parent scroll bars. 

There is, however, a fix listed in the power-select changelog that may have resolved it. We could look into it, but probably not as part of this issue.

tests/integration/components/github-issue-select-test.js
begedin: It actually would. I started the implementation thinking I'd be calling `renderPage()` for each test individually.

tests/pages/components/github-issue-select.js
begedin: I'll look into it, but I think there was a problem with that page object working for a power-select that isn't rendered in place.

begedin:
With the changes in our project - github repo relationships, this PR is no longer relevant.

PR #1312, opened 2017-05-31
chore(package): update ember-cli-htmlbars-inline-precompile to version 0.4.3

begedin:
Closes #1260


begedin:
Tried to check if the issues with ember-cli-babel we had are fixed. They are not, so this will have to wait for the next version.

Issue #1267, opened 2017-04-24
Enable github auth feature by adding button to profile route


begedin:
# Problem

Once all other tasks in this milestone are done, we should enable the connect feature by adding the link button in #1263 to the user profile route.

## Subtasks
- [ ] Add `github-connect-link-button` from #1263 to profile page
- [ ] Add profile page acceptance test checking that the button renders and has the proper href. 

## References

Blocked by #1266

JoshSmith:
I would think this would be done in settings, actually. What do you think?

JoshSmith:
We might want a settings submenu and put that into an integrations tab or something.

JoshSmith:
For example:

https://d3vv6lp55qjaqc.cloudfront.net/items/011L0n2i3B1c271g2O0u/Screen%20Shot%202017-05-31%20at%2010.35.56%20PM.png?X-CloudApp-Visitor-Id=1336043&v=09cbffb5

begedin:
Anything works, really, as long as it's somewhere.

PR #1317, opened 2017-06-01
Fix the GitHub connect authorization flow

JoshSmith:
# What's in this PR?

Some changes to the GitHub OAuth flow:

- Adds configuration for GitHub connect
- Changes the client and the API endpoints
- Adds a bearer token to the `ajax` service from `ember-simple-auth`'s token
- Uses ENV config in connect button
- Reduces default GitHub connect button down to just `{{github-connect state=state}}`
- Minor formatting changes


begedin:
@JoshSmith I reviewed and fixed the tests. Should be good to merge when you're ready.

app/templates/settings/profile.hbs
```diff
@@ -8,4 +8,6 @@
   <div class="settings">
     {{user-settings-form user=model}}
   </div>
+
+  {{github-connect state=state}}
```
begedin: You don't actually need to specify a state, no?

Also, this does not look good here. Do you wanna remove it before merge and deal with placing it in a separate PR for issue #1267 ?

I can add a PR for it.

JoshSmith: I'm not sure I know the reason for a `state`, actually.

PR #1318, opened 2017-06-01
Enable GitHub connect

### LINKED ISSUE #1267: Enable github auth feature by adding button to profile route

begedin:
Will close #1267 

The base of this is currently set to be #1317, so the changes are clearer. Should be changed to develop once that is merged.

I've added a new `settings.integrations` route. It's been added to the user menu. The layout is basic and needs work, but this gives a good base to work off of, with tests and infrastructure already written.


app/models/user.js
```diff
@@ -12,6 +12,7 @@ export default Model.extend({
   email: attr(),
   firstName: attr(),
   lastName: attr(),
+  githubAuthToken: attr(),
```
begedin: I'm not really sure if this is what we need here, but we need **a field** from the API.

`github_id` is an alternative and something that should be safe to render for all users. However, in github integrations, we'd still need to render the auth token for current users, if the client is talking to github directly. If the client will be doing it through our API, then we don't need it client side.

@joshsmith, I think this is worth making an issue for, to discuss. What do you think?

app/templates/components/user-dropdown.hbs
```diff
@@ -4,6 +4,7 @@
     seem to be used as hooks for testing only }}
     <li>{{#link-to 'slugged-route' user.username class="slugged-route"}}Your profile{{/link-to}}</li>
     <li>{{#link-to 'settings.profile' class="profile"}}Settings{{/link-to}}</li>
+    <li>{{#link-to 'settings.integrations' class="integrations"}}integrations{{/link-to}}</li>
```
begedin: Adding it to the menu should be good enough initially, but I'm ok with a different approach.

app/templates/settings/integrations.hbs
```diff
@@ -0,0 +1,15 @@
+{{title "Settings"}}
+
+<div class="container">
+  <div class="page-header">
+    <h3>Your account integrations</h3>
+  </div>
+
+  <div class="settings">
+    {{#if model.githubAuthToken}}
```
begedin: Depending on previous comment, this might be `{{#if model.githubId}}`.

app/templates/settings/integrations.hbs
```diff
@@ -0,0 +1,15 @@
...
+      Your account is already connected to github.
+    {{else}}
+      {{github-connect}}
+    {{/if}}
```
begedin: Generally, we may want to handle this via a component. Some sort of "integration card" type of thing.

app/models/user.js
JoshSmith: My gut instinct is that we do not want to expose this kind of data to the client. The API should ideally handle any sensitive data from GitHub and determine what's exposed to the client from there. We can render out a `github_id` to the client (and perhaps some other GitHub user information, as well, like username and photo), but the talking between GitHub and our service should happen via the API.

If you think this needs further discussion, we can do so in an issue. I am leaning heavily towards API-first here, though.

JoshSmith:
My tasks here are:

- [ ] Add a component for the GitHub connect button
- [ ] Tell the user what account they're connected to
- [ ] Remove usage of `githubAuthToken`
- [ ] Improve any layout and design issues

Some of this may require some API work.

JoshSmith:
Still kind of in progress. Didn't get far enough. You can finish if you'd like or I can pick it up myself tomorrow.

app/routes/github.js
```diff
@@ -28,30 +28,30 @@ export default Route.extend(AuthenticatedRouteMixin, {
     let stateValidator = get(this, 'githubState');
 
     if (stateValidator.validate(state)) {
-      return this._sendRequest(code);
+      return this._sendRequest(code, state);
```
begedin: The `github-state` service perform the validation locally, using the session data facilities provided by ember-simple-auth, so at the moment, there is no need to send the state to the API.

We discussed that as an option, but I don't think it's MVP and plenty of other libraries do such validation locally. I don't think it falls into the "should our client talk to github directly or through our API" story.

begedin:
@JoshSmith I've fixed and added missing tests. Only one acceptance test is failing now, which is the one related to the added `state` parameter, and should be easily fixed depending on how you wish to proceed.

app/routes/github.js
JoshSmith: I added this to the API and it works. GitHub does some state validation itself, apparently. At least that's how I read the docs. Can you correct me if I'm wrong on this?

begedin: The idea is

- We hit the github authorize URL with a `redirect_to` param and a `state` param
- On the authorization URL, if the user accepts, we are redirected back to whichever page is specified in `redirect_to`,  with a `code` param and a `state` param. 
- Our app is then obligated to check the `state` param against the original. Since this is a cross-site navigation, we need to store the original somehow and compare against it. In our case, this is done using the session data facilities provided by `ember-simple-auth` (which internally uses local storage), abstracted through our `github-state` service. In other cases, it could be some other form of local storage, a cookie, or a server-stored value.

begedin:
@JoshSmith Looked into this further and it really does seem like we're supposed to send the `state` in the second step as well. However, the documentation definitely implies we're supposed to be checking it ourselves to, so I'm unsure on what the fully correct method is. Either way, this works and I've fixed the final test.

PR #1327, opened 2017-06-12
Update ember-cli, babel and related dependencies. Fix issues.

begedin:
This has been a long time coming and will close a couple of greenkeeper PRs as well as simplify our future upgrade processes.

My assumption was, the tests were failing once we update ember-cli (and with that, babel) due to a babel issue.

However, a bunch of babel issues were fixed and we still had this problem, so I did some debugging.

Turns out, our problem was doing this at a module level:

```Javascript
const { Object } = Ember;
```

This code, when declared in a module root, would somehow mess with transpiling and overwrite `window.Object`. At that point, a bunch of things would outright break.

I resolved this issue by simply eliminating the need for even using `Object.create` in those plases, by switching from `someObject.get/set('prop')` to `get/set(someObject, 'prop')`. This is our conventional approach anyway, so any code that wasn't written this way was not convention compliant to begin with.


config/environment.js
```diff
@@ -1,9 +1,10 @@
 /* eslint-env node */
+'use strict';
 
 module.exports = function(environment) {
-  var ENV = {
+  let ENV = {
     modulePrefix: 'code-corps-ember',
-    environment: environment,
+    environment,
```
begedin: This file was changed due to running `ember init` and manually merging. Makes it more eslint compliant.

ember-cli-build.js
```diff
@@ -23,10 +23,6 @@ module.exports = function(defaults) {
   }
 
   var app = new EmberApp(defaults, {
-    'ember-cli-babel': {
-      // async, await, etc.
-      includePolyfill: true
-    },
```
begedin: Turns out we no longer need this. Probably due to ditching phantomjs. We used to need it to make use of `async/await`

package.json
```diff
@@ -20,13 +20,13 @@
     "ember-auto-focus": "1.0.9",
     "ember-autoresize": "0.6.0",
     "ember-can": "^0.8.4",
-    "ember-cli": "2.13.1",
+    "ember-cli": "~2.14.0-beta.2",
     "ember-cli-app-version": "^3.0.0",
     "ember-cli-autoprefixer": "0.7.0",
-    "ember-cli-babel": "^5.2.4",
+    "ember-cli-babel": "^6.0.0",
```
begedin: `6.0.0` is the one selected by ember-cli, but we should be safe to upgrade further.

package.json
```diff
@@ -20,10 +20,10 @@
     "ember-auto-focus": "1.0.9",
     "ember-autoresize": "0.6.0",
     "ember-can": "^0.8.4",
-    "ember-cli": "2.13.1",
+    "ember-cli": "~2.14.0-beta.2",
     "ember-cli-app-version": "^3.0.0",
     "ember-cli-autoprefixer": "0.7.0",
-    "ember-cli-babel": "^5.2.4",
+    "ember-cli-babel": "^6.4.1",
```
begedin: `ember init` sets the version to `6.0.0`, but really, the problem was the transition from `5.x` to `6.x`, so no point not to upgrade further.

begedin:
I was too quick to consider this done.

There is an issue with `ember-page-title`, to which I submitted a fix - https://github.com/tim-evans/ember-page-title/pull/76

There is also a resolved issue with `ember-code-coverage`, but a new version has not been released yet, so this is fixed by getting it directly from github. As soon as a new version is released, we should switch back.

package.json
```diff
@@ -20,12 +20,12 @@
...
     "ember-cli-bourbon": "2.0.0-beta.1",
-    "ember-cli-code-coverage": "0.3.12",
+    "ember-cli-code-coverage": "kategengler/ember-cli-code-coverage",
```
begedin: - 0.3.12 fails with the new ember-cli version
- 0.3.20 resolves that issue, but there's a new one, related to babel
- The babel issue is fixed in master, but a new version has not been released yet.

package.json
```diff
@@ -75,15 +75,15 @@
     "ember-modal-dialog": "2.2.0",
     "ember-moment": "7.3.1",
     "ember-normalize": "1.0.0",
-    "ember-page-title": "^3.1.2",
+    "ember-page-title": "^3.2.0",
```
begedin: - 3.1.2 has an issue with ember-cli
- 3.2.0 fixes it, but there's a regression to an old issue we already have. I submitted a PR to fix it.

begedin:
@joshsmith We should consider merging this. The tests pass, but two addons which were causing trouble have to point at github master branches in order for the tests to pass, reason being, new versions have not been released yet. 

We cannot keep the old versions because they fail with new ember-cli.

Greenkeeper will create PRs as those addons get updated anyway, and this finally takes us out of the woods with the recent ember-cli update issues.

PR #1335, opened 2017-06-26
Add project integrations tab

JoshSmith:
# What's in this PR?

Adds the beginnings of a project integrations tab for connecting to GitHub.


app/routes/project/settings/integrations.js
```diff
@@ -0,0 +1,30 @@
...
+  model() {
+    let project = this.modelFor('project');
+    let organization = get(project, 'organization');
+    let user = get(this, 'currentUser.user');
+
+    return RSVP.hash({ project, organization, user });
+  },
+
+  setupController(controller, { project, organization, user }) {
+    set(controller, 'project', project);
```
snewcomer: I think here you can do `controller.setProperties(hash)` if you want.  Otherwise, this is more explicit!

JoshSmith: Yeah, I think I like the explicitness here. What do you think?

begedin: `setProperties(controller, { project, organization, user });` is the best of both worlds, IMHO.

JoshSmith: Updated to the `setProperties`.

JoshSmith:
@begedin this could use a review and merge so I can just do the rest in a separate PR, I think.

PR #1350, opened 2017-07-20
Add start of github-app-installation component and Mirage setup

JoshSmith:
# What's in this PR?

Still a work in progress.

### DIFF
--- a/app/components/github-repo.js
+++ b/app/components/github-repo.js
@@ -1,11 +1,11 @@
 import Ember from 'ember';
-import recordsList from 'code-corps-ember/utils/records-list';
 
 const {
-  computed: { alias },
   computed,
+  computed: { alias, mapBy, notEmpty },
   Component,
-  get
+  get,
+  inject: { service }
 } = Ember;
 
 export default Component.extend({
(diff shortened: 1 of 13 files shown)


app/components/github-repo.js
```diff
@@ -0,0 +1,22 @@
...
+  classNames: ['github-repo'],
+  tagName: 'li',
+
+  isConnected: computed('githubRepo', 'project.projectGithubRepos.@each', function() {
+    let githubRepo = get(this, 'githubRepo');
+    let project = get(this, 'project');
+    let projectGithubRepos = get(this, 'project.projectGithubRepos');
+    if (projectGithubRepos) {
+      return recordsList.includes(projectGithubRepos, githubRepo);
+    }
```
JoshSmith: Not sure how to get this to work. Get the following:

> There is no belongsTo relationship named 'github-repo' on a model of modelClass 'project-github-repo'

...which is obviously not true.

begedin: In the component body:

```Javascript
githubRepo: null, // gets set from somewhere
projectGithubRepos: computed.alias('project.githubRepos'),
connectedProjectGithubRepos: computed.filter('projectGithubRepos', 'githubRepo',function(projectGithubRepo, index, projectGithubRepos) {
  return get(projectGithubRepo, 'githubRepo.id') === get(githubRepo, 'id');
}),
isConnected: computed.notEmpty('connectedProjectGithubRepos'),
```

Would that not work?

begedin: Generally, I'm not sure I love the `recordsList` utility class. It feels like we made a class just to reduce the line count, at a great cost to clarity. If there's some code repetition somewhere, that's fine. We don't have to generalize everything.

I'd write it here first, then, if it turns out the code is identical to some other piece of code, we generalize.

JoshSmith:
@begedin made some good progress with this today. Should be good to start wiring up the proper relationships for the organization GitHub app installations and making those components work, then writing some tests for both of these.

app/components/github-repo.js
```diff
@@ -3,20 +3,46 @@ import recordsList from 'code-corps-ember/utils/records-list';
...
-    let projectGithubRepos = get(this, 'project.projectGithubRepos');
-    if (projectGithubRepos) {
-      return recordsList.includes(projectGithubRepos, githubRepo);
+  githubRepo: null,
+  project: null,
+
+  store: service(),
+
+  isConnected: notEmpty('projectsGithubRepo'),
+  projectGithubRepo: computed('projectGithubRepos.@each', 'projectGithubRepos.isFulfilled', 'githubRepo.id', function() {
```
begedin: This is the part that always bugs me. I'm not sure we need to watch `isFulfilled` here. I'm not even sure we need to watch `@each`. Possibly even just the base is enough, but I'm not sure.

begedin: Either way, the naming here is not at all clear to me and it seems like maybe this is a case of overextending concerns, as in this component cares too much about its parent/siblings. Maybe the parent (controller?) should just map out a collection and assign an `isConnected` property to the component?

Again, not sure. Just trying to find an alternative to something that keeps being problematic for us.

JoshSmith: I think it's worth doing as much because it also isolates the component and makes it more testable, with significantly less testing setup. I was similarly worried about doing things with the `store` inside this component and would instead prefer to pass that up to the controller.

begedin:
@JoshSmith I tried rebasing this as best as I could, but since the merged PR from #1353 took most of the changes here into develop and then built from that, I think it will be easier and a better use of time to cherry pick changes we still might need here into a new PR.

begedin:
@JoshSmith This one is now completely out of date. 

The part that's still missing from the dev branch is done on the opi and involves having the installation index route only return installations accessible by the current user. Everything else has been implemented.

PR #1357, opened 2017-08-17
Update Ember addons en masse

JoshSmith:
# What's in this PR?

Updates a bunch of addons at once.


JoshSmith:
@begedin I've been stuck on this for a couple hours unfortunately. It looks like for whatever reason, `ember-page-title` is breaking tests with:

> The instance initializer 'head-browser' has already been registered

You had landed a fix in `ember-page-title` for this error in previous versions:

> TypeError: Cannot read property 'activeTransition' of undefined

...so I was hesitant to revert back really far since this appeared to be working for you. Nonetheless, it is not working for me, even with a `nombom`. I'm about to try to force a rebuild without cache on CI and hope that works, but the error did present there, too.

JoshSmith:
Yeah, no. This is still failing on CI even with triggering a rebuild.

begedin:
@JoshSmith I landed a fix for another issue. That fix was sadly overwritten in a later release, but seems to be back in now. However, this is a different issue which I encountered in #1295. The only thing I could do there was to revert to an older version of the addon and then implement my own fix locally, by extending our "mockRouter" helper.

I would skip even touching that addon in this PR.

begedin:
@joshsmith 

I've applied the fix from my other PR (#1295) and set the version of `ember-page-title` to `3.2.0` as the one that currently isn't broken for us. I did this as a separate commit so it's slightly better documented.

This should now be good to go, so I'll merge.

PR #1368, opened 2017-09-11
Upgrade circle version

JoshSmith:
# What's in this PR?

Attempt to upgrade to CircleCI 2.0.


begedin:
@JoshSmith Still not able to get around this, but for some information

In some cases, instead of getting an out of memory error, the test step simply hangs at 

```
Could not start watchman
Visit https://ember-cli.com/user-guide/#watchman for more info.
Building
'instrument' is imported from external module 'ember-data/-debug' but never used
Too long with no output (exceeded 10m0s)
```

JoshSmith:
@begedin no guarantees that this deploys correctly, but we'll see soon enough!

PR #1379, opened 2017-09-19
Add component to display username

JoshSmith:
# What's in this PR?

This adds a `{{user/display-username}}` component. This is used only in places where you could potentially have a username displayed for a user that does not have a Code Corps username (e.g. if they were created from GitHub).

### DIFF
--- a/app/templates/components/comment-item.hbs
+++ b/app/templates/components/comment-item.hbs
@@ -27,7 +27,7 @@
   </div>
 {{else}}
   <div class="comment-info">
-    {{#link-to 'slugged-route' comment.user.username class="username"}}{{comment.user.username}}{{/link-to}} commented {{moment-from-now comment.insertedAt}}
+    {{user/display-username user=comment.user}} commented {{moment-from-now comment.insertedAt}}
     {{#if canEdit}}
       <a class="edit" {{action 'edit'}}>edit</a>
     {{/if}}
--- a/app/templates/components/user/display-username.hbs
+++ b/app/templates/components/user/display-username.hbs
@@ -1,0 +1,5 @@
+{{#if username}}
+  {{#link-to data-test-username 'slugged-route' username class="username"}}{{username}}{{/link-to}}
+{{else if githubUsername}}
+  GitHub user <a data-test-github-username href={{githubUserUrl}} class="username">{{githubUsername}}</a>
+{{/if}}
--- a/tests/pages/components/comment-item.js
+++ b/tests/pages/components/comment-item.js
@@ -2,8 +2,7 @@ import {
   clickable,
   collection,
   fillable,
-  isVisible,
-  text
+  isVisible
 } from 'ember-cli-page-object';
 
 export default {
@@ -39,5 +38,7 @@ export default {
     itemScope: '.error'
   }),
 
-  username: text('.username')
+  username: {
+    scope: '[data-test-username]'
+  }
 };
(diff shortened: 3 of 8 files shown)


JoshSmith:
I wonder if this shouldn't be called `user/link-username` to potentially distinguish it from display-only, without a link.

JoshSmith:
This has some failing tests I'll need to circle back and address.

app/templates/components/comment-item.hbs
```diff
@@ -27,7 +27,10 @@
   </div>
 {{else}}
   <div class="comment-info">
-    {{#link-to 'slugged-route' comment.user.username class="username"}}{{comment.user.username}}{{/link-to}} commented {{moment-from-now comment.insertedAt}}
+    {{user/display-username
+      githubUsername=comment.user.githubUsername
+      username=comment.user.username
+    }} commented {{moment-from-now comment.insertedAt}}
```
begedin: This might in the future end up depending on other factors to, not just `username` and `githubUsername`.

Since the purpose of the component is "displaying the correct username content for the user, depending on the information the user provides", maybe we should pass in the entire user, not just these two fields.

app/templates/components/user/display-username.hbs
```diff
@@ -0,0 +1,5 @@
+{{#if username}}
+  {{#link-to data-test-username 'slugged-route' username class="username"}}{{username}}{{/link-to}}
+{{else if githubUsername}}
+  GitHub user <a data-test-github-username href={{githubUserUrl}} class="username">{{githubUsername}}</a>
```
begedin: Might not be a bad thing to add `(via GitHub)` here, but not sure how nice it would look with the "timestamp" that goes after in some cases.

JoshSmith: Right now it reads `GitHub user joshsmith commented 21 hours ago`. I think after that we could add `– from GitHub` with a link to the GitHub app. But that wouldn't be part of this component then. I also plan in the avatar to add the GitHub logo in the lower right corner, similar to how GitHub does with our app's logo, to make it even clearer.

app/templates/components/comment-item.hbs
JoshSmith: Makes sense, especially given the conversation that I'd eventually like some metadata on the user that helps the logic be more explicit about their current identity provider.

app/templates/components/user/display-username.hbs
begedin: I have no idea how I didn't see the `Github user` part here. Just ignore my comment

JoshSmith:
Will need squashed once approved @begedin.

tests/pages/components/comment-item.js
```diff
@@ -39,5 +39,8 @@ export default {
     itemScope: '.error'
   }),
 
-  username: text('.username')
+  username: {
+    scope: '[data-test-username]',
+    text: text()
```
begedin: I think it may be that you'll have a `text` property on `username` even without explicitly defining it. 

I'm reasonably sure all page elements (objects with `scope`) have a couple of these by default. Could you check and remove them if that's the case?

tests/pages/components/user/display-username.js
```diff
@@ -0,0 +1,15 @@
+import { text } from 'ember-cli-page-object';
+
+export default {
+  scope: '.user__display-username',
+
+  githubUsername: {
+    scope: '[data-test-github-username]',
+    text: text()
```
begedin: Same here

Issue #1385, opened 2017-09-24
Handle display of empty body in tasks/comments with "No description provided.


JoshSmith:
# Problem

It's possible for both `Task` and `Comment` to now have a `null` body. We should handle this in the various components by doing an `if` check in the hbs template and rendering: 

> _No description provided._

Just like that, in italics.

fxvarga:
Found an additional issue while looking into this.

Should we handle empty inputs in the submittable-textarea component, or will those get turned to null in server?

- Currently if the body textarea is not clicked into before save, ```'<p>undefined</p>'``` is sent to server.
- If the body textarea is clicked into but left empty then ```'<p></p>'``` is sent to the server.

JoshSmith:
@fxvarga Good call! I think it's perhaps worth opening a separate issue for this.

While we allow `NULL` values in the database it's really only due to GitHub allowing these values to be null.

I'd like to explore running client-side validations on this and not allowing anything to even be submitted without there being a value. The reason for this is general contributor experience. A task without a description is not very helpful. A comment without one is even less so. In both of these cases it should really have a client-side presence validation.

We've discussed adding in changeset validations in #609. This would be an ideal first case to introduce it, no?

begedin:
My general opinion is that client side validations are welcome and greatly enhance the client, but server side validations are still needed, generally speaking.

The specific example above, though, is a client-side bug and should absolutely be resolved. We should not be sending '<p>undefined</p>'. I'd even consider it a `submittable-textarea` bug, not something to do with validations specifically.

JoshSmith:
Closed by #1407

PR #1386, opened 2017-09-25
Handle null body in Task and Comment model

fxvarga:
# What's in this PR?
Handle null body in Task and Comment model by showing the message "No  description provided" in the comment item and task item templates.
## References
Fixes [#1385](https://github.com/code-corps/code-corps-ember/issues/1385)

### DIFF
--- a/app/components/comment-item.js
+++ b/app/components/comment-item.js
@@ -50,22 +50,42 @@ export default Component.extend({
   currentUser: service(),
 
   /**
-   * processedBody - Computed property
+   * parsedBody - Computed property
    *
    * Parses the provided body and "linkifies" the username mentions within it
    * using information provided in the mentions collection
    *
    * We are expected to assign mentions in the template
    *
+   * The component can load before the comment body has loaded.
+   * Computing isLoaded allows us to parse the body once the data is available.
+   *
    * NOTE: This feature is currently disabled, so the unmodified body is returned
    *
    * @property processedComment
    @ @type String
    */
-  processedBody: computed('comment.body', 'mentions', function() {
+  parsedBody: computed('comment.body', 'comment.isLoaded', 'mentions', function() {
     let body = get(this, 'comment.body');
     let mentions = get(this, 'mentions');
     return parse(body, mentions || []);
+  }),
+
+  /**
+   * nullbody - Computed property
+   *
+   * checks to see if outcome of processed content is null, indicating null
+   * comment body and returns computed property as true if body is null.
+   *
+   * It tests specifically against null because parsedBody generates other
+   * falsey outputs even when comment body is not null. For example, if the
+   * comment data is not loaded yet, parsedBody returns undefined.
+   */
+  nullBody: computed('parsedBody', function() {
+    if (get(this, 'parsedBody') === null) {
+      return true;
+    }
+    return false;
   }),
 
   /**
--- a/app/styles/_markdown.scss
+++ b/app/styles/_markdown.scss
@@ -155,3 +155,7 @@
     margin-bottom: 0 !important;
   }
 }
+
+.markdown-body--empty {
+    color: $text--light;
+}
--- a/app/styles/app.scss
+++ b/app/styles/app.scss
@@ -43,8 +43,6 @@
 @import "base/base";
 @import "base/helpers";
 
-
-
 //
 //
 //
@@ -54,7 +52,6 @@
 @import "code";
 @import "images";
 @import "markdown";
-
 //
 // LAYOUT
 //
--- a/tests/integration/components/task-details-test.js
+++ b/tests/integration/components/task-details-test.js
@@ -133,3 +133,29 @@ test('it saves', function(assert) {
   page.save.click();
   assert.equal(page.commentBody.text, 'A new body', 'The body is saved');
 });
+
+test('if the task body is null render the no description element', function(assert) {
+  stubService(this, 'mention-fetcher', {
+    prefetchBodyWithMentions() {
+      return null;
+    }
+  });
+
+  this.set('task', mockTask);
+
+  page.render(hbs`{{task-details task=task}}`);
+  assert.ok(page.nullCommentBody.isVisible, 'The message for no comment body is rendered');
+});
+
+test('if the task body is not null do not render the no description element', function(assert) {
+  stubService(this, 'mention-fetcher', {
+    prefetchBodyWithMentions() {
+      return 'A body';
+    }
+  });
+
+  this.set('task', mockTask);
+
+  page.render(hbs`{{task-details task=task}}`);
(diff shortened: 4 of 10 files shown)


tests/pages/components/task-details.js
```diff
@@ -22,7 +22,10 @@ export default {
     hasStrongText: isVisible('strong'),
     text: text()
   },
-
+  nullCommentBody:{
+    scope: '.null-comment-body',
```
JoshSmith: I would rename this class to `data-test-null-body` (in all cases here) which will get automatically removed later. Also think slightly less confusing than `comment` in context of task.

tests/integration/components/task-details-test.js
```diff
@@ -133,3 +133,15 @@ test('it saves', function(assert) {
   page.save.click();
   assert.equal(page.commentBody.text, 'A new body', 'The body is saved');
 });
+test('if the task is null render the no description element', function(assert) {
```
JoshSmith: Would add a space above this test.

tests/integration/components/task-details-test.js
```diff
@@ -133,3 +133,15 @@ test('it saves', function(assert) {
...
+  stubService(this, 'mention-fetcher', {
+    prefetchBodyWithMentions() {
+      return null;
+    }
+  });
+
+  this.set('task', mockTask);
+
+  page.render(hbs`{{task-details task=task}}`);
+  assert.ok(page.nullCommentBody.isVisible, 'The message for no comment body is rendered');
```
JoshSmith: Can we test for the opposite case where there _is_ a body (i.e. this should _not_ render)?

app/components/comment-item.js
```diff
@@ -62,13 +62,30 @@ export default Component.extend({
...
+   * 
+   * checks to see if outcome of processed content is null, indicating null comment body
+   * and returns computed property as true if body is null.
+   * 
+   * NOTE: only did it this way because there is a test that lead me to think that this
+   * component could possibly load before comment body is loaded. This will prevent the 
+   * "No description provided" message from showing up on load before data is available.
+   */
+  nullBody: computed('processedBody', function() {
+    if (get(this, 'processedBody') === null) {
```
JoshSmith: I think we may want to use `Ember.isBlank` here.

app/components/comment-item.js
```diff
@@ -62,13 +62,30 @@ export default Component.extend({
    * @property processedComment
    @ @type String
    */
-  processedBody: computed('comment.body', 'mentions', function() {
+  processedBody: computed('comment.body', 'comment.isLoaded', 'mentions', function() {
     let body = get(this, 'comment.body');
     let mentions = get(this, 'mentions');
-    return parse(body, mentions || []);
+    let parsedBody = parse(body, mentions || []);
+    return parsedBody;
```
JoshSmith: Also wanted to say I think we should just still write this as `return parse`. Is there a reason you rewrote this?

JoshSmith: Maybe we should rewrite `processedBody` to `parsedBody` generally? Any thoughts @begedin since I think you wrote this originally?

tests/pages/components/task-details.js
JoshSmith: Whoops. So this is not a CSS class I'm looking for but a `data-` attr. Just to clarify.

app/components/comment-item.js
begedin: Both seem to fit fine so I don't lean towards any of the names.

JoshSmith: I think that case then if we can change `processedBody` to `parsedBody` that'd be easier to grok.

tests/pages/components/task-details.js
begedin: @JoshSmith You sure you don't want a class to, for potential styling?

Github, for example, seems to italicize the text and make it a bit lighter than the normal body text.

fxvarga: To verify, is this what you are looking for?

```<em class="null-comment-body" data-test-null-body=true>No description provided.</em>```

Comment (JoshSmith) [reply tests/pages/components/task-details.js]: `<em data-test-null-body>No description provided.</em>`

Comment (JoshSmith) [reply tests/pages/components/task-details.js]: In `/styles/_markdown.scss` I would add

```css
.markdown-body--empty {
  em {
    color: $text--light;
  }
}
```

fxvarga:
Added another commit to this pull request, with the changes for above.

tests/integration/components/task-details-test.js
```diff
@@ -145,3 +146,16 @@ test('if the task is null render the no description element', function(assert) {
   page.render(hbs`{{task-details task=task}}`);
   assert.ok(page.nullCommentBody.isVisible, 'The message for no comment body is rendered');
 });
+
+test('if the task body is not nul do not render the no description element', function(assert) {
```
JoshSmith: `not nul` to `not null` here.

app/components/comment-item.js
```diff
@@ -72,15 +75,16 @@ export default Component.extend({
...
+   * checks to see if outcome of processed content is null, indicating null 
+   * comment body and returns computed property as true if body is null.
    * 
-   * NOTE: only did it this way because there is a test that lead me to think that this
-   * component could possibly load before comment body is loaded. This will prevent the 
-   * "No description provided" message from showing up on load before data is available.
+   * It tests specifically against null because parsedBody generates other 
+   * falsey outputs even when comment body is not null. 
+   * For example is if comment data is not loaded yet, 
+   * parsedBody trurns undefined
```
JoshSmith: Would reword as:

> It tests specifically against `null` because `parsedBody` generates other falsey outputs even when comment body is not `null`. For example, if the comment data is not loaded yet, `parsedBody` returns `undefined`.

tests/pages/components/task-details.js
```diff
@@ -22,7 +22,10 @@ export default {
     hasStrongText: isVisible('strong'),
     text: text()
   },
-
+  nullCommentBody:{
+    scope: '.markdown-body--empty',
```
JoshSmith: You'll want to change this and the other `scope` to `[data-test-markdown-body-empty]`, and you can remove the `isVisible` and call `isVisible()` directly in the test without issue.

app/styles/_markdown.scss
```diff
@@ -0,0 +1,3 @@
+.markdown-body--empty {
+    color: $text--light;
+}
```
JoshSmith: You can remove this file and rebase on `upstream/develop` which will pull this file in and add the class there. I forgot you were not rebased yet onto our `upstream/develop`.

app/styles/app.scss
```diff
@@ -54,7 +54,7 @@
 @import "code";
 @import "images";
 @import "inputs";
-
+@import "markdown";
```
JoshSmith: This line can be reverted.

JoshSmith:
Comment.

PR #1389, opened 2017-09-27
Improves existing layouts with new styles

JoshSmith:
# What's in this PR?

This moves some of the layout changes over into other settings areas for a more consistent look and feel throughout the app.


JoshSmith:
I know these tests are failing due to some things that tested loading states, but if it can be reviewed before I fix those, that would be awesome.

JoshSmith:
@begedin it's buried right now, somewhat intentionally. You get there at `organizations/:slug`

PR #1390, opened 2017-09-28
Fix bug in verification-document hbs that does not show if it's not errored

JoshSmith:
# What's in this PR?

This fixes a bug that displays an error (and the verification document itself) if the verification document process has errored somehow.

### DIFF
--- a/app/templates/components/payments/funds-recipient/verification-document.hbs
+++ b/app/templates/components/payments/funds-recipient/verification-document.hbs
@@ -1,4 +1,4 @@
-{{#if (eq status 'required')}}
+{{#if (or (eq status 'required') (eq status 'errored'))}}
   {{#if isBusy}}
     <p>Processing...</p>
   {{else if isUploading}}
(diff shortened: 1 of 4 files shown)


JoshSmith:
This requires a test before merging. I haven't gotten to it yet.

app/templates/components/payments/funds-recipient/verification-document.hbs
```diff
@@ -15,7 +15,10 @@
         uploadDone=(action onUploadDone)
         uploadError=(action onUploadError)
         validationError=(action onValidationError)}}
-    </div>
+    </p>
+    {{#if (eq status 'errored')}}
+      <p class="error">Stripe was unable to verify your identity with the document you provided. Please try again.</p>
+    {{/if}}
     {{#if error}}
       <div class="error">{{error}}</div>
```
begedin: We got a `did.error` and a `p.error` here. Can they be shown at the same time and are we ok with how the margins are in that case?

PR #1398, opened 2017-10-02
chore(package): update ember-cli-page-object to version 1.12.0

JoshSmith:
Closes #1227

### DIFF
--- a/package.json
+++ b/package.json
@@ -50,7 +50,7 @@
     "ember-cli-mirage": "^0.3.4",
     "ember-cli-moment-shim": "^3.4.0",
     "ember-cli-neat": "^0.0.6",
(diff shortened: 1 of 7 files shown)


JoshSmith:
It looks like tests are legitimately failing here so this will need some work to bring this in line with the changes from `ember-cli-page-object`.

begedin:
The tests seem to have been failing due to the skills typeahead control.

I had trouble with this one in the past sometimes, when  running tests in `server` mode, using the `-s` flag. They would not reliably pass.

I guess an internal change in `ember-cli-page-object` caused these tests to now always fail, which allowed me to implement a fix to make them pass reliably.

I'll merge this unsquashed, so we have a separate commit for this fix.

PR #1404, opened 2017-10-03
Change membership to project user

### LINKED ISSUE #1166: Change mentions of membership to project user
# Problem

We removed `OrganizationMembership` and replaced it with `ProjectUser`. The use of the word `member` in components like `project-card-members` is now potentially confusing, so we should attempt to change the language, e.g. to something like `project-card-users`, which is clearer.

daveconnis:
# What's in this PR?

WIP.

## References
Fixes #1166


tests/integration/components/project-members-test.js
```diff
@@ -28,7 +28,7 @@ test('it renders an item for each member in the list', function(assert) {
   }
 
   this.set('users', mockUsers);
-  page.render(hbs`{{project-members users=users}}`);
+  page.render(hbs`{{project-users users=users}}`);
 
   assert.equal(page.memberCount, 3, 'The correct number of members are rendered');
   assert.ok(page.members(0).imageSource.indexOf('image_1.png') > -1, 'The correct photo renders');
```
JoshSmith: I would change this and the page objects to be `user` and `users` instead of `member` and `members`, as well.

tests/pages/components/project-users.js
```diff
@@ -6,7 +6,7 @@ import {
 } from 'ember-cli-page-object';
 
 export default {
-  scope: '.project-members',
+  scope: '.project-users',
 
   memberCount: count('li'),
```
JoshSmith: Here's one.

JoshSmith:
🙌

PR #1413, opened 2017-10-08
Add GitHub issue link to sidebar

### LINKED ISSUE #1392: Add GitHub icon and link to issue to the task card
# Problem

For any task with a corresponding GitHub issue, we should provide a GitHub icon and link to the issue on GitHub to the task card.

### LINKED ISSUE #1393: Add GitHub icon and link to GitHub issue to the task sidebar
# Problem

For any task with a corresponding GitHub issue, we should provide a GitHub icon, the repo, the issue #, and link to the issue on GitHub to the task sidebar.

JoshSmith:
# What's in this PR?

Adds GitHub issue links to the sidebar and task cards.

## References
Fixes #1393 
Fixes #1392


app/components/github/issue-link.js
```diff
@@ -0,0 +1,21 @@
...
+export default Component.extend({
+  classNames: ['github__issue-link'],
+
+  repoName: alias('githubRepo.name'),
+  repoOwner: alias('githubRepo.githubAccountLogin'),
+
+  url: computed('repoOwner', 'repoName', 'number', function() {
+    let { repoOwner, repoName, number } = getProperties(this, 'repoOwner', 'repoName', 'number');
+    return `https://www.github.com/${repoOwner}/${repoName}/issues/${number}`;
+  })
```
begedin: I believe this is good enough for now, but generally, the GitHub API guidelines recommend **against** developers building links on their own and instead nudge towards using the prebuilt link properties.

I think we should create an ember issue to switch this property to ¸`task.githubIssue.link` once and if the API implements a `GithubIssue` model

JoshSmith: So the documentation you're referring to I think is this: https://developer.github.com/v3/guides/best-practices-for-integrators/#dont-manually-parse-urls

It actually more specifically recommends against parsing URLs and not building them. My best guess is that the `owner/repo/issues/number` convention is not going away any time soon, and we actually have a high likelihood of not breaking anything since the `GithubRepo` will update (and we'll get the new values for that).

That said, if we did have a `GithubIssue` model, we would definitely want to embed that in the JSON so we don't have to load all those records separately.

PR #1435, opened 2017-10-20
Split task-assignment component out of task-card 

### LINKED ISSUE #1434: Split task-assignment component out of task-card
# Problem

Right now there’s a `power-select` inside of `task-card.js`. Everything inside of (and including) `{{#if canAssign}}` should go inside a new component called `task-assignment`.

In `task-card.js` and then in the `power-select` there are a bunch of actions that match up to the actions in `task-card`. Those could be moved down a level into `task-assignment`. This would allow us to use `task-assignment` everywhere there’s a task, including on the task page.

To orientate yourself, spend some time looking through `task-card.hbs` and` task-card.js`,  `task-card-test.js` and `pages/component/task-card` and the pages it imports from (like `power-select`, specifically).

daveconnis:
# What's in this PR?

Splitting task-assignment component from task-card

## References
Fixes #1434

### DIFF
--- a/app/components/task-assignment.js
+++ b/app/components/task-assignment.js
@@ -1,0 +1,81 @@
+import Component from '@ember/component';
+import { alias } from '@ember/object/computed';
+import { inject as service } from '@ember/service';
+import { set, getProperties, computed } from '@ember/object';
+import EmberCan from 'ember-can';
+import createTaskUserOptions from 'code-corps-ember/utils/create-task-user-options';
+
+export default Component.extend({
+  classNames: ['task-assignment'],
+
+  currentUser: service(),
+  taskAssignment: service(),
+
+  task: null,
+  taskUser: null,
+  users: null,
+
+  // auto-assigns 'task' property from component as ability 'model'
+  ability: EmberCan.computed.ability('task'),
+  canAssign: alias('ability.canAssign'),
+  currentUserId: alias('currentUser.user.id'),
+  taskUserId: alias('taskUser.id'),
+
+  // TODO: this updates selection when it changes. However, it updates while
+  // the change is still processing, and rolls back if it fails.
+  // We should somehow skip the update if the change is processing
+  // TODO: It also fails to roll back when reassignment fails
+  didReceiveAttrs() {
+    let { taskUserId, users } = getProperties(this, 'taskUserId', 'users');
+    if (users) {
+      set(this, 'selectedOption', users.findBy('id', taskUserId));
+    }
+  },
+
+  /**
+    Computed property, builds and maintains the list used to render the
+    dropdown options on task assignment
+
+    @property userOptions
+  */
+  userOptions: computed('currentUserId', 'taskUserId', 'users', function() {
+    let { currentUserId, taskUserId, users }
+      = getProperties(this, 'currentUserId', 'taskUserId', 'users');
+    if (users) {
+      return createTaskUserOptions(users, currentUserId, taskUserId);
+    } else {
+      return [];
+    }
+  }),
+
+  actions: {
+    buildSelection(option, select) {
+      if (option === select.selected) {
+        return null;
+      }
+      return option;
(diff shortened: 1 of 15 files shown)


tests/integration/components/task-card-test.js
```diff
@@ -317,6 +317,6 @@ test('assignment dropdown does not render if user has no ability', function(asse
 
   assert.notOk(page.taskAssignment.select.triggerRenders, 'Dropdown trigger for assignment does not render.');
   page.taskAssignment.assignedUser.as((user) => {
-    assert.equal(page.taskAssignment.assignedUser.icon.url, 'test.png');
+    assert.equal(user.icon.url, 'test.png');
```
JoshSmith: Guessing this would be `page.something`. This is only passing because you're testing the data that you created yourself, not the component's data.

app/components/task-assignment.js
```diff
@@ -0,0 +1,87 @@
...
+  ability: EmberCan.computed.ability('task'),
+  canAssign: alias('ability.canAssign'),
+  currentUserId: alias('currentUser.user.id'),
+  taskUserId: alias('taskUser.id'),
+
+  // TODO: this updates selection when it changes. However, it updates while
+  // the change is still processing, and rolls back if it fails.
+  // We should somehow skip the update if the change is processing
+  // TODO: It also fails to roll back when reassignment fails
+  didReceiveAttrs() {
```
begedin: @daveconnis Could you make an issue for these two todos? Just describe where they are written in the code and copy them and the `didReceiveAttrs` function into the issue.

Generally, it's not a great approach to write todos in code, so if you see any in the future left by us, feel free to let us know by creating an issue.

tests/integration/components/task-card-test.js
```diff
@@ -4,22 +4,22 @@ import PageObject from 'ember-cli-page-object';
...
 function renderPage() {
   page.render(hbs`
-    {{task-card
-      clickedTask=clickedTask
-      users=users
+    {{task-assignment
       task=task
       taskUser=taskUser
+      users=users
     }}`);
```
begedin: This is the `task-card-test`, which means you're testing the `task-card` component, so the `renderPage` fuction should render that component.

The idea with integration tests is to ensure that the various bits of the application work as expected with each other. In `task-card-test`, we are testing how well the task card interacts with the outside and how well her internals work.

We want to write tests to ensure the internals that deal with task assignment work and the task card component interacts with the outside world correctly when those internals are used, so the old `renderPage()` was fine as it was.

tests/integration/components/task-card-test.js
```diff
@@ -49,7 +49,7 @@ test('it renders all the required elements', function(assert) {
   };
 
   set(this, 'task', task);
-  this.render(hbs`{{task-card task=task}}`);
+  page.render(hbs`{{task-card task=task}}`);
```
begedin: You want to call `renderPage()` in all of this instances instead.

The general flow of an integration test is

- prepare the data and the component - set all the data and action handlers, mock all the requires servicesw, etc. 
- render the page - using `renderPage`
- simulate using the component on the page in some way - using `page`

- assert that the thing that needed to happen happened

tests/integration/components/task-card-test.js
```diff
@@ -174,8 +173,8 @@ test('assignment works if user has ability', function(assert) {
 
   renderPage();
 
-  page.taskAssignment.trigger.open();
-  page.taskAssignment.dropdown.options(0).select();
+  page.taskAssignment.select.trigger.open();
+  page.taskAssignment.select.dropdown.options(0).select();
```
begedin: This one is on the right track. The other tasjs that call `page.render` directly should be replaced with `renderPage`, once `renderPage` is reverted back to the old version, where it renders the `task-card` component, not the `task-assignment` component

JoshSmith:

PR #1437, opened 2017-10-21
Update skills list

JoshSmith:
# What's in this PR?

Introduces a `user/skills-list` component.


JoshSmith:
@daveconnis you can finish this up by adding some component tests and a page object for this, and if necessary updating any tests in components that make use of it by pulling in that new page object.

PR #1438, opened 2017-10-22
Redesign task form

JoshSmith:
# What's in this PR?

Redesigns the new task form to look like this:

screen shot 2017-10-21 at 6 38 31 pm


JoshSmith:
This shows up regardless of whether there are GitHub repos. We need to change that and add tests.

PR #1444, opened 2017-10-23
Add keyboard shortcuts for assigning

JoshSmith:
# What's in this PR?

- Press <kbd>A</kbd> to open the task assignment dropdown when hovering the task card
- Press <kbd>Space</kbd> to self-assign/unassign when hovering the task card


JoshSmith:

JoshSmith:
This should only be worked on once #1492 is done.

PR #1463, opened 2017-10-27
Replace imports with @ember syntax

### LINKED ISSUE #1380: New Ember module import syntax.
# Problem

Ember is moving towards importing just the modules you need instead of the global `Ember` import.

https://github.com/emberjs/rfcs/blob/master/text/0176-javascript-module-api.md

## Subtasks
- [ ] use https://github.com/ember-cli/ember-modules-codemod to replace `import Ember from 'ember'` with `import Component from '@ember/component'`.
- [ ] some situations will not be covered by this codemod.  Need to evaluate on an individual basis.

Might need to plan on when this can be best accomplished given other PR's in the pipeline.

snewcomer:
using https://github.com/ember-cli/ember-modules-codemod

## References
Fixes #1380

### DIFF
--- a/app/abilities/organization.js
+++ b/app/abilities/organization.js
@@ -1,12 +1,7 @@
-import Ember from 'ember';
+import { alias } from '@ember/object/computed';
+import { get, computed } from '@ember/object';
+import { inject as service } from '@ember/service';
 import { Ability } from 'ember-can';
-
-const {
-  computed,
-  computed: { alias },
-  get,
-  inject: { service }
-} = Ember;
 
 export default Ability.extend({
   currentUser: service(),
--- a/app/abilities/project.js
+++ b/app/abilities/project.js
@@ -1,13 +1,8 @@
-import Ember from 'ember';
+import { equal, alias } from '@ember/object/computed';
+import { get, computed } from '@ember/object';
+import { inject as service } from '@ember/service';
+import { isEmpty } from '@ember/utils';
 import { Ability } from 'ember-can';
-
-const {
-  computed,
-  computed: { alias, equal },
-  get,
-  inject: { service },
-  isEmpty
-} = Ember;
 
 /**
  * Ability object used to determine what the current user can do with a project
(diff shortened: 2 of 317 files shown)


JoshSmith:
Why does coverage drop quite so much here?

snewcomer:
@JoshSmith That is a good question.  When looking at this [report](https://coveralls.io/builds/13930119/source?filename=app%2Fmixins%2Fonboarding-controller.js#L34), it shows red at the points where imports like `getProperties` were refactored to the new module import syntax.

JoshSmith:
Good observation. My guess is that the `getProperties` calls were being incorrectly picked up as "covered" even though the actual logic was not really covered.

JoshSmith:
I don't know whether it's realistic to expect to get this coverage back up in the context of this PR. But we definitely would want to open an issue to address it.

Do you have any sense as to whether any other imports were missed by codemod?

Also, this is a tiny stylistic point, but would you mind introducing PRs/commits with "Adds/fixes/removes [verb] whatever the fix was"? I think it brings more clarity to reading the commits (for me, at least) relative to `[FEATURE]:` `[BUG]:` etc style.

snewcomer:
@JoshSmith good idea.  Lmk if you like the new title.  Also, I looked around and it looks like one like this is not [replaced](https://github.com/code-corps/code-corps-ember/pull/1463/files#diff-008d2d2d10af03d5f0cf00901375ab44R8) and namespace imports `import as` aren't supported by the addon but don't see any of those.  

I don't see `testing` in this [list](https://github.com/emberjs/rfcs/blob/master/text/0176-javascript-module-api.md#addendum-2---table-of-module-names-and-exports-by-package).

snewcomer:
Also looks like `ember-data` and ember-cli-shims is still needed.  [RFC](https://github.com/emberjs/rfcs/pull/238) still open.

JoshSmith:
Good call then. This is a good title rework. I would just amend your commit message with the same and I’ll approve for merge. If you could open an issue for us to go back and fix some of this test coverage that would be great.

PR #1465, opened 2017-10-28
Rework Stripe layout

JoshSmith:
# What's in this PR?

Reworks the Stripe layout to be more in line with changes included in `_panel.scss`.

Will also rework the layout of other panels to make better use of `aside` and `section`.

Changes design from this:

...to this...


JoshSmith:

PR #1469, opened 2017-10-28
Add project github repo sync component

JoshSmith:
# What's in this PR?

_I have an important question below that could change the outcome of this PR slightly and needs addressed._

- adds a `format-number` helper and tests
- `format-currency` now uses `format-number`
- adds icons for comments, issues, pull requests, repos
- improves testing of icons
- adds error state and animation state to progress bar
- adds `github/project-repo-sync` component that
  - renders progress text that includes the count
  - animates the progress bar when actually syncing
  - renders the appropriate icon for the current step of the sync
  - computes error state from `projectGithubRepo.state`
  - when errored
    - renders error text
    - renders a red progress bar and icon


JoshSmith:
This still needs:

- unit tests for new `format-number` helper
- refactor `format-currency` to use `format-number` helper
- add icons for github issues and github comments
- test for presence of the proper icons depending on the current step
- add icon for error in the syncing
- remove testing component from `projects/index.hbs`

JoshSmith:
This should be good to review now.

JoshSmith:
An idle thought I just had: should the syncing progress state not be on `GithubRepo` rather than `ProjectGithubRepo`. It's an easy fix now, but I think the syncing operations in the API happen on a `GithubRepo` and would be easier (and perhaps even best) to track there.

begedin:
@JoshSmith 

>An idle thought I just had: should the syncing progress state not be on `GithubRepo` rather than `ProjectGithubRepo`?

Possibly on both

- `GithubRepo` has a progress for fetching and creating `GithubComment, GithubIssue, GithubPullRequest`
- `ProjectGithubRepo` has progress for creating `Task` records on that `Project`

Technically, however, the second one isn't strictly necessary. The first one, once done, means the second one just needs that one step. Possibly the second one just needs a `synced` field, if we end up creating `Task` records asynchronously.

begedin:
Also technically, that would mean that, once a `GithubRepo` is synced and we handle the webhooks we need to handle, we no longer need to sync it upon connecting subsequent `ProjectGithubRepo` records. All we need to do is create `Tasks`

JoshSmith:
@begedin keen observation there. We may want to circle back and tackle that later in the sync, since our MVP is probably just to hamfist it. I would consider creating an issue for that.

In the meantime, I'm going to recommend that we place all the heavy lifting onto `GithubRepo` and opt to rename this component, along with updating the related API issues to reflect the change from `ProjectGithubRepo` to `GithubRepo` for the syncing statuses.

Any objections?

app/components/github/project-repo-sync.js
```diff
@@ -0,0 +1,60 @@
+import Component from '@ember/component';
+import { get, computed } from '@ember/object';
+import { alias } from '@ember/object/computed';
+
+export default Component.extend({
+  classNames: ['github__project-repo-sync'],
+
+  projectGithubRepo: null,
+  totalSteps: 7,
```
begedin: I think a common thing would be to declare this as a const up top, outside the object. 

Either

```Javascript
const TOTAL_STEPS = 7;

  totalSteps: TOTAL_STEPS,
});
```
or even just using the `const` directly instead of re-assigning it to `this.`

JoshSmith: Another thought I had was pulling this from the API rather than hardcoding it, a la Stripe's `fields_needed`. That's way overdoing it, though.

Another minor refactor in that direction could be putting the steps in a simple array of objects as a `const` outside the component object.

begedin: that works as well, I actually prefer that

begedin:
>In the meantime, I'm going to recommend that we place all the heavy lifting onto `GithubRepo` and opt to rename this component, along with updating the related API issues to reflect the change from `ProjectGithubRepo` to `GithubRepo` for the syncing statuses.

>Any objections?

None

JoshSmith:
- Needs renamed to `github/repo-sync`
- Need to change model tests to `GithubRepo` from `ProjectGithubRepo`
- Component needs a `GithubRepo` rather than `ProjectGithubRepo`

PR #1470, opened 2017-10-29
Improve project view on mobile

JoshSmith:
# What's in this PR?

Improves the project view on mobile by reordering the sidebar (with more important information for newcomers to the project) above the main details.

We can improve this more later by adding the project's short description to the sidebar on mobile views with a link to read the rest of the description inside that short.

### DIFF
--- a/app/styles/templates/project/index.scss
+++ b/app/styles/templates/project/index.scss
@@ -1,6 +1,12 @@
 .project-wrapper {
+  display: flex;
+  flex-direction: row;
   min-height: 50vh;
   width: 100%;
+
+  @include media($sm-screen) {
+    flex-direction: column;
+  }
 }
 
 .project-main {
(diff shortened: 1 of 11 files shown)


app/styles/templates/project/index.scss
```diff
@@ -1,14 +1,25 @@
 .project-wrapper {
+  display: flex;
+  flex-direction: column;
   min-height: 50vh;
   width: 100%;
+
+  &--index {
```
snewcomer: huh I never knew this was a rule you could use (I see `project-wrapper--full`).  Cool!

app/styles/templates/project/index.scss
```diff
@@ -1,14 +1,25 @@
...
+    flex-direction: row;
+  }
+
+  @include media($sm-screen) {
+    flex-direction: column;
+  }
```
snewcomer: I guess one way you could do it is in the `&--index`, you could use 

```
  &--index {
    @include media(>$sm-screen) {
       flex-direction: row;    
    }
  }
```

don't know if the greater than sign can be used like this, but the thought is there.  Then you could remove this line I believe.

PR #1475, opened 2017-10-30
Fix keyframes and fix the animated high five component

JoshSmith:
# What's in this PR?

Fixes the `animated-high-five` component by adding unprefixed `@keyframes` and removes prefixed versions everywhere.

### DIFF
--- a/app/styles/base/_animations.scss
+++ b/app/styles/base/_animations.scss
@@ -46,55 +46,7 @@
   to   { opacity: 1; }
 }
 
-/* Firefox < 16 */
-@-moz-keyframes fadein {
-  from { opacity: 0; }
-  to   { opacity: 1; }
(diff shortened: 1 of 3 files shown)


snewcomer:
@JoshSmith could the prefixes be dropped other places like loading-bar.scss and image-drop.scss?

JoshSmith:
@snewcomer I believe this was just for `@keyframes`. I haven't looked into prefix support for other things.

snewcomer:
@JoshSmith ah gotcha...i saw autoprefixer as a pkg so I thought you were removing all prefixed css declarations.

JoshSmith:
@snewcomer it's possible we might be able to. I haven't actually checked to see if it's autoprefixing correctly.

PR #1480, opened 2017-11-01
Add project github repo syncing to the integrations page

### LINKED ISSUE #1467: Add settings area component for ProjectGithubRepo
# Problem

Each `ProjectGithubRepo` in the list on the project settings integrations page needs to have a settings area that fulfills the following requirements.

Progress bar shows:
“Fetching pull requests...”
“Syncing 1,234 pull requests...”

We'll have to change the `github-repo` component here:

- Clicking connect, instead of creating a `ProjectGithubRepo` component, simply slides down to reveal a settings area (our new component)
- Can click edit settings instead of remove
- When the settings area is revealed, show a close link in the actions
- When the project is syncing, slide down the sync status and don't show a close link in the actions

The settings area itself has two states:

- unconnected
  - Tells you what’s going to happen when you connect
  - Renders a connect button which creates the `ProjectGithubRepo`
- connected
  - Warns you not to disconnect with explainer of why
  - Renders a disconnect button which deletes the `ProjectGithubRepo`

The settings area should send an action up when it should be closed (on connect, disconnect) and indicate whether the project is now syncing.

Blocked by #1466

### LINKED ISSUE #1468: Get status of GithubRepo syncing by polling
# Problem

We need to evaluate which is faster vs better to do and then do that:

Using the component created in #1466:

- poll for sync status of ProjectGithubRepo
- or, get sync status through Phoenix channel

If the latter, we probably need to create some blocking issues here and in the API.

Blocked by #1466

JoshSmith:
# What's in this PR?

Work in progress.

## References
Fixes #1467 
Fixes #1468


JoshSmith:
For the life of me I can no longer get the settings panel to remain open when starting to sync. I would love for it to close when sync is finished, but not when it starts.

Thoughts @begedin?

app/components/github-repo.js
```diff
@@ -1,4 +1,5 @@
-import { notEmpty, alias } from '@ember/object/computed';
+import { computed, get } from '@ember/object';
+import { alias, notEmpty, or } from '@ember/object/computed';
 import Component from '@ember/component';
```
begedin: I'm loving this so much.

app/components/github/connected-installation.js
```diff
@@ -1,4 +1,4 @@
-import { alias } from '@ember/object/computed';
+import { alias, sort } from '@ember/object/computed';
 import Component from '@ember/component';
 import { getProperties, get, computed } from '@ember/object';
```
begedin: Guessing you still wanted to sort the repos here?

app/components/github/install-link.js
```diff
@@ -11,6 +11,10 @@ export default Component.extend({
   organization: null,
   project: null,
 
+  installUrl: computed('githubAppUrl', function() {
+    return `${get(this, 'githubAppUrl')}/installations/new`;
+  }),
```
begedin: #1481 deals with this directly, so maybe we wanna remove this when that is merged? I actually prefer the other approach. If we don't need the root app url for anything and only ever use this one, might as well set it directly.

app/components/github/repo-disconnect-confirm-modal.js
```diff
@@ -0,0 +1,22 @@
...
+export default Component.extend({
+  githubRepo: null,
+  typedGithubRepoName: '',
+
+  githubRepoName: alias('githubRepo.name'),
+  namesMatch: computed('githubRepoName', 'typedGithubRepoName', function() {
+    let githubRepoName = get(this, 'githubRepoName');
+    let typedGithubRepoName = get(this, 'typedGithubRepoName');
+    return githubRepoName === typedGithubRepoName;
+  }),
```
begedin: Cool approach 👍

app/components/github/repo-sync.js
```diff
@@ -1,51 +1,85 @@
...
+    let repoState = get(this, 'repoState');
+
+    let state = projectRepoState;
+
+    if (!state
+        || state === 'unsynced'
+        || state === 'syncing_github_repo'
+        || state === 'errored_syncing_github_repo') {
+      state = repoState;
+    }
```
begedin: I'm thinking rewrite this as

```Javascript
if (['syncing_tasks', 'syncing_comments', 'synced'].indexOf(state) == -1) {
  state = repoState;
}
```

begedin: You can also extract to a separate function, something like `getRenderedState(projectRepoState, repoState)`, or it could be a computed. You can also do both.

app/components/github/install-link.js
JoshSmith: We do need the root app url because we also later want to provide a link for editing the existing configuration of an app installation directly.

app/mixins/repo-sync-progress.js
```diff
@@ -0,0 +1,106 @@
...
+      case 'errored_syncing_comments':
+        return 8;
+      case 'synced':
+        return 9;
+      default:
+        return 0;
+    }
+  }),
+
+  shouldShowSync: computed('repoState', 'projectRepoState', 'triggeredSync', function() {
```
begedin: Something irks me about names like these. It feels sort of like naming a boolean variable `isTrue`.

Yes, the variable tells us we should show the sync UI, but it doesn't tell us why. How about `syncIncomplete` or `syncRequired`. I don't think any of these are perfect, but there's just something off to me with `shouldShowSync`.

app/mixins/repo-sync-progress.js
```diff
@@ -0,0 +1,106 @@
...
+    let repoState = get(this, 'repoState');
+    let projectRepoState = get(this, 'projectRepoState');
+    let triggeredSync = get(this, 'triggeredSync');
+
+    if (triggeredSync && projectRepoState !== 'synced') {
```
begedin: `!get(this, 'syncComplete'`)` for some re-usability, but I understand if you prefer this one as possibly more performant.

JoshSmith:
This needs a test for the mixin and new tests for:

- the `github-repo` component
- the `github-repo-settings` component
- the `repo-disconnect-confirm-modal` component

tests/integration/components/github-repo-test.js
```diff
@@ -28,47 +32,91 @@ moduleForComponent('github-repo', 'Integration | Component | github repo', {
...
+  set(this, 'githubRepo', githubRepo);
   renderPage();
 
-  run(() => set(this, 'model.githubRepo', { isLoaded: true }));
+  assert.ok(page.actions.connect.text, 'Connect', 'Connect link renders.');
+
+  run(() => set(this, 'githubRepo.syncState', 'fetching_comments'));
+  assert.notOk(page.actions.close.isVisible, 'Close link does not render while syncing.');
+  assert.notOk(page.actions.connect.isVisible, 'Connect link does not render while syncing.');
+  assert.notOk(page.actions.edit.isVisible, 'Close link does not render while syncing.');
```
begedin: Bad message here

JoshSmith: Fixed, thanks!

PR #1488, opened 2017-11-07
Fix hover state on task assignment (no tests)

joshsmith:
# What's in this PR?

No tests. Hides the tooltip for task assignment when the dropdown is open and shows it when re-hovering and not open.


JoshSmith:
@daveconnis this could use:

- tests for the mixin
- tests for each individual `_selected` component
- a test for the `task-card` component

app/mixins/tooltip-for-dropdown.js
```diff
@@ -0,0 +1,25 @@
...
+      set(this, 'tooltipShown', true);
+    }
+  },
+
+  mouseLeave() {
+    set(this, 'tooltipShown', false);
+  },
+
+  dropdownOpened: observer('dropdownOpen', function() {
+    let tooltipShown = get(this, 'dropdownOpen') ? false : null;
```
begedin: On this line alone, I think this mixin needs some documentation. 

Probably a description of how the tooltip ought to behave in relation to the dropdown.

tests/unit/mixins/tooltip-for-dropdown-test.js
```diff
@@ -0,0 +1,12 @@
...
+import { module, test } from 'qunit';
+
+module('Unit | Mixin | tooltip for dropdown');
+
+// Replace this with your real tests.
+test('it works', function(assert) {
+  let TooltipForDropdownObject = EmberObject.extend(TooltipForDropdownMixin);
+  let subject = TooltipForDropdownObject.create();
+  assert.ok(subject);
+});
```
begedin: If we're not writing tests as part of this PR, it could be an option to at least add a blank `skip`

https://api.qunitjs.com/QUnit/skip

Along the lines of

```Javascript
skip('description of behavior A', function(assert) {});
skip('description of behavior B', function(assert) {});
```

This will report those tests as skipped in the output and keep us aware.

Also, while it does not apply here, take note of `QUnit.todo` - https://api.qunitjs.com/QUnit/todo

JoshSmith:
I'm going to wait on #1492 to get merged to finish this.

tests/unit/mixins/tooltip-for-dropdown-test.js
```diff
@@ -10,3 +11,17 @@ test('it works', function(assert) {
   let subject = TooltipForDropdownObject.create();
   assert.ok(subject);
 });
+
+test('mouseEnter dropdownOpen is true', function(assert) {
+  let TooltipForDropdownObject = EmberObject.extend(TooltipForDropdownMixin);
```
daveconnis: So when I started to expect I was forcing these to pass, I deleted line 16 and 17 and the test still passed. I'm assuming it's because I'm setting `dropdownOpen` to true above the assert, so it's not actually running the tooltip. However, at the same time, I'm pretty sure I'm pulling dropdown from our `TooltipforDropdownObject` so a part of me isn't sure how else I'd approach this.

daveconnis:
Note to Self:

For the mixin, you basically need to look at, okay, what does it expose? it exposes properties of `dropdownOpen` and `tooltipShown`, has a `mouseEnter` `mouseLeave`, and observes the property of `dropdownOpen` 
`dropdownOpened` observer has branching logic that hinges on whether the `dropdownOpen` is true or false
`mouseEnter` has branching logic dependent upon whether `dropdownOpen` is true or false

so I would to test that:
- the values of `dropdownOpen` and `tooltipShown` are initially false by default
- when I call `mouseEnter()` and `dropdownOpen`, `tooltipShown` is still `false`
- when I call `mouseEnter()` and `dropdownOpen` is `false`, `tooltipShown` is now `true`

1) testing the default state of the properties
2) mouseEnter dropdownOpen is true
3) mouseEnter dropdownOpen is false
4) mouseLeave
5) dropdownOpened when dropdownOpen true
6) dropdownOpened when dropdownOpen false

daveconnis:
Done for the day. 
- I wrote two tests for the `unselected-item`, both testing the the default state of `unselected-item`. 
- Debugged the handlebars helper.
- Found that I'd incorrectly named the page object and need to change that when I pick this back up.
- Discovered that issues rendering the Tooltip isn't actually happening because the tooltip is lazy rendered. Need to parse this out a bit more.
- There are test assertions in ember tool tips that we're not making use of right now. Would like to switch the assertions we're using with the official ones.
- Still need to write tests for mouseEnter and mouseLeave.
- Not super sure how dropdownOpen impacts these tests. Any insight here would be nifty.

daveconnis:
Done for the day:
- Fixed incorrectly named page object.
- Researched the shit out of test assertion helpers for tooltip
- wrote mouseEnter test. IT IS PASSING. I WROTE A TEST THAT IS PASSING.
- writing a mouseLeave test, but I'm worried I'm making it more complicated than it needs to be. I originally wanted to just call the "mouseLeave" action within the test with hopes that I could just do a `this.get(toolTipShown` afterwards and assert that it equaled false. I think I'm doing that, but I might be doing it the hard way. I'll start back on this tomorrow. 
- re "Discovered that issues rendering the Tooltip isn't actually happening because the tooltip is lazy rendered. Need to parse this out a bit more." The test assertions via tooltip docs spoke straight to this and gave me an assertion for this issue specifically. I think we should use the same assertions in other places where we test tooltip such as `/Users/ImaginationStation/sites/code-corps-ember/tests/integration/components/github/issue-link-test.js`.
-  Still not sure how dropdownOpen impacts these tests. Any insight here would be nifty.

daveconnis:
Done for the day:
- Started writing page object for `selected-item`. Unsure how to proceed on making a selectedIcon object for it to use like we made a loadingIcon using`[data-test-unselected-icon] for 'Unselected-item`. There's no data-test object in the .hbs for selected-item, however there is 'src={{select.selected.photoThumbUrl}} />' so I'm assuming that I'll just make a fake"selectedIcon.
- Finished 'unselected-item' tests

daveconnis:
- Look at `unselected-item-test.js`. There are two tests there that you'll want  in `selected-item-test.js`

- 'the default state when user task is loading' test should be renamed to `'the tooltip is not rendered initially'`

-`when mousing over and out again` should be renamed to 'the tooltip renders lazily, triggered by mouseenter' 

- once you have the two tests, `selected-item` and `unselected-item`. Add a test to task-card
 ```render a task card component where the user selection dropdown has 2 users, one selected, one unselected
check that neither of the two users have a tooltip visible, nor rendered
hover first user
check that first user now has tooltip visible and rendered
hover out of first user
check that first user now has tooltip invisible, but still rendered
do the same for the second user```

daveconnis:
- renamed tests

- Deleted `the tooltip is not rendered initially' test in `selected-item` test because it tested the isLoading logic found in `unselected-item`. I'm assuming there won't be an `isLoading` state  because the item has already been selected and therefore loaded.

begedin:
> Deleted `the tooltip is not rendered initially' test in `selected-item` test because it tested the isLoading logic found in `unselected-item`. I'm assuming there won't be an `isLoading` state  because the item has already been selected and therefore loaded.

Not necessarily. There could be many different situations where we know for a fact a record is selected, but the record itself is not fully loaded yet. Considering a `task` is assigned to a `user` through a `taskUser` record.

Here's how it works under the hood:

- `task` is loaded, with that, we already havea  ist of `taskUser` record ids, so right now, each `taskUser.isLoading` is true, but we do know its already there and can render somethig.
- `taskUser` records are loaded for the `task` using those ids. As they load, we suddenly know a `userId` for each of those `taskUser` records that load. So now we know there's a user there, but the `user` still isn't loaded. We can definitely tell if something is selected by this point, even though not all of it is fully loaded yet
- `user` records now also get loaded and once that's done, everything is fully loaded

From that, depending on how it all runs under the hood, we could know if something is selected as early as step 1, or it could be step 2, but step 3 is when everything is loaded.

> - Have a failing test. Dying after first assertion, on `page.mouseenter`. Throwing error "Error: getTooltipFromBody(): returned an element that is not a tooltip." Assuming I'm not passing in the correct object for the test to work. Thought maybe passing in `taskUser` was what I needed to work on, but unsure. What I do know is, my `data-test-selected-icon` doesn't seem to be rendering. Any advice here would be awesome. 

The model for the component needs to be structured as 

```Javascript
{
  selected: someFakeUserModel
}
```

I can see why you went with some sort of fake `taskUser` model. The property assigned to the `selected-item` component on line 15 of `task-assignment.hbs` is called `taskUser`. However, if you step outside that component, into task card, then into task-list-cards, you will see that the `taskUser`is actually `task.userTask.user`.

In your defense, that actually **is** confusing, so it could probably be written in a nicer way.

Another method I could've used to figure it out is...

I searched the code for `selected-item`. Found it in `task-assignment.hbs`

```Handlebars
{{task-card/user/selected-item select=(hash selected = selectedOption)}}
```

I know `hash` is a helper which basically builds a JS object "on the fly", so now I know that `selected` needs to be assigned a value structured as `{selected: something}`.

Next, I need to find out what `something` is, so I search the code for `selectedOption`

I find it in `task-assignment.js`

```Javascript
set(this, 'selectedOption', users.findBy('id', taskUserId));
```

I see it's one of the assigned users, so it must be userlike in some way. 

Then, looking at `selected-item.hbs` and `selected-item.js` I check which properties this `userlike` object needs to have (as in, which properties the component is trying to access.

From `selected-item.hbs`, I see it's `.photoThumbUrl` for the icon and `username` for the text.

From `selected-item.js`, I don't see any additional properties.

So based on all that, I need to do a `renderPage` which looks like 

```Javascript
function renderPage() {
  page.render(hbs`
    {{task-card/user/selected-item select=(hash selected = selectedOption)}}
  `);
}
```

Then, in any of my tests, I do a

```Javascript
let mockUser = {
  photoThumbUrl: 'somefakeimagepath.jpg',
  username: 'somefakeusername'
};

set(this, 'selecetedOption', mockUser);

renderPage();

// do the test stuff
```

Of course, the structure of `mockUser` changes based on the needs of the test.

> - Also unsure why my commit isn't under my github profile. This is what happened in my terminal when I made the commit. Putting here for reference as well as for someone to look at and tell me what I did wrong.

This is sort of normal. Just follow the instructions you're given.

```
    git config --global user.name "daveconnis"
    git config --global user.email the.email.you.use@on.github

    git commit --amend --reset-author
```

Do this before any new commits. After the last one, you'll have to `git push --force-with-lease`

If you did any other unpushed commits, it's still solvable, or we can just ignore, since we'll likely squash everything into a single commit before the PR is merged.

## Potential tests

Since I had a deeper look at the code now, here's a couple of tests we might need if they aren't already in

- `selected-item` renders icon and username correctly - > single test, 2 assertions
- `task-assignment` renders in a deferred state initially

daveconnis:
Done for the day.

Finished selected-item tests and started working on task-card test which is going to follow this roadmap:

- render a task card component where the user selection dropdown has 2 users, one selected, one unselected
- check that neither of the two users have a tooltip visible, nor rendered
- hover first user
- check that first user now has tooltip visible and rendered
- hover out of first user
- check that first user now has tooltip invisible, but still rendered
- do the same for the second user

I initially started doing this by adding to the task-card page object what you'd find in page objects for `selected/unselected-item`. Stuff like `tooltip`, ` mouseenter: triggerable('mouseenter')`, etc, but then stopped and deleted this thinking that I was adding unnecessary things to the task-card page object. 

My thinking here is that's why we wrote selected and unselected item components, to be able to call them specifically in task-card and not have to deal with their inner-workings, aka tooltip and mouseEnter and whatnot. We're just testing the component and its integration to task-card. So I'm currently working with that logic as a foundation, and moving forward looks like figuring out how to call the functionality of the selected and unselected components themselves within the test, not adding their (scope specific) bits and pieces to the task-card. I could be completely wrong here, as I'm not 100% sure that this is the right way to move forward because it does seem more straightforward to share those specific `selected-item` and `unselected-item` bits and then write road map above by copy and pasting tests straight from `selected-item` and `unselected-item`, but this seems like it wouldn't be testing the actual components themselves. So, tomorrow I'm going to go from there.

Also, currently getting a `error: getTooltipFromBody(): returned an element that is not a tooltip` on my task-card test. Sure it's just because I haven't yet thought through all that's at play here. I also haven't even finished the test, but I just wanted to include what was happening at my stop so I could have more pointed direction.

daveconnis:
- Had to fix `selected-item` tests, as I was passing in the data wrong. 
- Started on writing tests for `task-card-test` that test if `unselected-item` and `selected-item` play well with `task-card`.
- Following this flow:

- [ ] when task has a user assigned (same data as unassignment test), the selected-item component is visible
- [ ] when task has no user assigned (same data as assignment test), the unselected-item component is visible
- [ ] when task has a user assigned, the selected item component renders tooltip lazily
- [ ] when task has no user assigned, the unselected item component renders tooltip lazily

- Finished `task-card` tests for `selected-item` component, now working on `unselected-item` component. Upon stopping, test is failing because I'm getting a false value on ```assert.ok(page.unselectedItem.unselectedIcon.isVisible, 'the unselected user icon renders.');```. There are two icons in the `unselected-item.hbs`, loading and unselected. I don't think we will ever test a tooltip on the loading icon, i tried in the UI and it was impossible, so I'm trying to test the tool tip on the unselectedIcon, which I added to the `unselected-item` page object. However, I don't think I'm setting my data up correctly. One thing you'll see is I'm attempting to set isLoading to false so the 'unselected-item.hbs` will only load everything under the else statement.

-Considering the above, in `unselected-item-test` we test the default state when user task is loaded:
```test('the default state when user task is loaded', function(assert) {
  let task = { userTask: { isLoading: false } };
  this.set('task', task);
  renderPage();
  assert.notOk(page.loadingIcon.isVisible, 'The loading icon does not render.');
  assert.notOk(page.isTooltipTarget, 'There is no tooltip because it lazy renders.');
```
We only test that the loading icon doesn't render, never testing that the unselected icon does. Shouldn't we test something rendering over its predecessors not rendering?

tests/integration/components/task-card-test.js
```diff
@@ -293,3 +300,48 @@ test('assignment dropdown does not render if user has no ability', function(asse
...
+  let task = { id: 'task' };
+  let user = { id: 'user', username: 'testuser', photoThumbUrl: 'test.png' };
+  let users = [user];
+  let taskUser = user;
+
+  setProperties(this, { task, users, taskUser });
+
+  renderPage();
+  assert.equal(page.selectedItem.selectedIcon.src, taskUser.photoThumbUrl, 'the selected user photo icon renders');
+  assert.notOk(page.tooltip, 'there is no tooltip because mouse is not hovering.');
```
begedin: These are valid assertions and I'm completely fine with keeping them in this way. However, a simpler thing you could've also done is simply do

```Javascript
assert.ok(page.selectedItem.isVisible, 'The selected item component is rendered on the card.');
```

The way the component is implemented right now means that both methods work and test the behavior correctly, so as i said, both methods are as valid. 

However, if the `selectedItem` component changes internally at some point the `isVisible` test will still work.

tests/integration/components/task-card-test.js
```diff
@@ -293,3 +300,48 @@ test('assignment dropdown does not render if user has no ability', function(asse
...
+  assertTooltipNotRendered(assert);
+
+  page.selectedItem.mouseenter();
+
+  assertTooltipVisible(assert);
+  assert.equal(page.selectedItem.tooltip.text, `Assigned to ${taskUser.username}`, 'the tooltip renders the correct text.');
+});
+
+test('the unselected-item component is visable when a task has a user assigned ', function(assert) {
```
begedin: you got a typo here -> `visable`

tests/integration/components/task-card-test.js
```diff
@@ -293,3 +300,48 @@ test('assignment dropdown does not render if user has no ability', function(asse
...
+  assert.expect(4);
+  let task = { userTask: { isLoading: false }, id: 'task' };
+  let user = { id: 'user', username: 'testuser' };
+  let users = [user];
+
+  setProperties(this, { task, users });
+
+  renderPage();
+  assert.ok(page.unselectedItem.unselectedIcon.isVisible, 'the unselected user icon renders.');
```
begedin: You're one step closer to what I proposed above, but you could just as easily test for 

```Javascript
assert.ok(page.unselectedItem.isVisible, 'The unselected item component renders on the card.');
```

tests/integration/components/task-card-test.js
```diff
@@ -293,3 +300,48 @@ test('assignment dropdown does not render if user has no ability', function(asse
...
+  page.unselectedItem.mouseenter();
+
+  assertTooltipVisible(assert);
+  assert.equal(page.unselectedItem.tooltip.text, 'Assigned this task', 'the tooltip renders the correct text.');
+
+  page.unselectedItem.mouseleave();
+
+  assertTooltipRendered(assert);
+  assertTooltipNotVisible(assert);
+});
```
begedin: These two tests are actually testing 4 behaviors, 2 per test.

- one test checks if selected item is rendered when appropriate, and if the deferred tooltip rendering behavior works on it.
- another test checks if the unselected item component is rendered when appropriate and if the deferred tooltip rendering behavior works on it

DRY is a great concept, but generally, even at the cost of repeating ourselves, we want tests that each test one aspect of a component's behavior. At least, that's the approach we want to use with integration and unit tests.

With acceptance tests, since they are much slower and test overall end to end behavior, we want one test per discrete user action and outcome, which then asserts all aspects of it behaved correctly.

In this specific case, that means we'd prefer 4 tests

- check if selected item is rendered when appropriate,
- check if deferred tooltip rendering behavior works on selected item
- check if the unselected item component is rendered when appropriate
- check if the the deferred tooltip rendering behavior works on unselected item

However, from that list, it's clear that tests 2 and 4 are already covered in `selected-item-test` and `unselected-item-test`, so we don't actually need them here. All we need is tests 1 and 3

Specifically, that means we just want to leave the `.isVisible` assertion in each of the two tests, and remove everything else.

That probably annoys you, but you did the work correctly, so it has value, and I'd likely go through the same process myself, before deciding on this route.

daveconnis:
Finished writing the tests, but hit a snag in the last test.

```test('the unselected-item component is visible when a task has no user assigned', function(assert) {
  assert.expect(1);

  this.register('ability:task', Ability.extend({ canAssign: true }));

  renderPage();
  assert.ok(page.unselectedItem.isVisible, 'the unselected item component renders.');
});```

Initially, I didn't have ` this.register('ability:task', Ability.extend({ canAssign: true }));`. I didn't include it because it seemed to me that the assign ability doesn't hold anything over whether or not `unselected-item` works. I brought this up to Nikola, and he said I found a bug. This is me telling you about the bug.

begedin:
@daveconnis, @joshsmith 

It looks like the `selected/unselected` icons are "trapped" behind `canAssign`, or at the very least, written in a way that it appears they are.

I think, with these tests in place now, the `task-card.hbs` warrants a bit of a refactor to make it clearer.

PR #1489, opened 2017-11-07
Write tests for services/project user

### LINKED ISSUE #1182: Write tests for services/project-user
# Problem

The `projectUser` service is currently untested. The only part of the interface is the `joinProject` function, so that's what needs to be tested.

## Subtasks

Write the following tests:
- [ ] `joinProject` creates a new projectUser with properties `{project, currentUser, role: 'pending'}`
- [ ] `joinProject` creates a flash notification on success

## Implementation

Probably the simplest approach to this would be to inject stubs into the `store`, `flashMessages` and `currentUser` services used by the `projectUser` service.

This will allow us to test that the correct actions on those 3 have been called.

daveconnis:
# What's in this PR?

Two tests for project-user component.

 joinProject creates a new projectUser with properties {project, currentUser, role: 'pending'}
 joinProject creates a flash notification on success

Fixes #1182

### DIFF
--- a/app/services/project-user.js
+++ b/app/services/project-user.js
@@ -6,6 +6,8 @@ export default Service.extend({
   flashMessages: service(),
   store: service(),
 
+  flashOptions: { fixed: true, sticky: false, timeout: 5000 },
+
   joinProject(project) {
     let user = get(this, 'currentUser.user');
     let store = get(this, 'store');
@@ -14,11 +16,15 @@ export default Service.extend({
 
     return store.createRecord('project-user', projectUser)
       .save()
-      .then(() => this._flashSuccess('Your request has been sent.'));
+      .then(() => this._flashSuccess('Your request has been sent.'))
+      .catch(() => this._flashError('Your request has not been sent.'));
   },
 
   _flashSuccess(message) {
-    let options = { fixed: true, sticky: false, timeout: 5000 };
-    get(this, 'flashMessages').clearMessages().success(message, options);
+    get(this, 'flashMessages').clearMessages().success(message, get(this, 'flashOptions'));
+  },
+
+  _flashError(message) {
+    get(this, 'flashMessages').clearMessages().danger(message, get(this, 'flashOptions'));
   }
 });
--- a/tests/unit/services/project-user-test.js
+++ b/tests/unit/services/project-user-test.js
@@ -1,15 +1,91 @@
+import RSVP from 'rsvp';
+import { set } from '@ember/object';
 import { moduleFor, test } from 'ember-qunit';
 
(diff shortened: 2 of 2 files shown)


app/services/project-user.js
```diff
@@ -13,12 +13,17 @@ export default Service.extend({
     let projectUser = { project, user, role: 'pending' };
 
     return store.createRecord('project-user', projectUser)
-      .save()
-      .then(() => this._flashSuccess('Your request has been sent.'));
-  },
+    .save()
+    .then(() => this._flashSuccess('Your request has been sent.'))
+    .catch(() => this._flashError('Your request hasn not been sent.'));
+
```
JoshSmith: Missing `},` to close the function.

app/services/project-user.js
```diff
@@ -13,12 +13,17 @@ export default Service.extend({
...
   _flashSuccess(message) {
     let options = { fixed: true, sticky: false, timeout: 5000 };
     get(this, 'flashMessages').clearMessages().success(message, options);
-  }
+  },
+
+  _flashError(message) {
+    let options = { fixed: true, sticky: false, timeout: 5000 };
+    get(this, 'flashMessages').clearMessages().danger(message, options);
 });
```
JoshSmith: Missing `}` to close the function.

tests/unit/services/project-user-test.js
```diff
@@ -1,15 +1,85 @@
...
+  set(service, 'currentUser', mockCurrentUser);
+  set(service, 'flashMessages', mockFlashMessage);
+  set(service, 'store', mockStore);
+
+  service.joinProject(project).then(() => {
+    done();
+  });
+
+  test('it sends an error flash message if joinProject fails', function(assert) {
+    let service = this.subject();
```
JoshSmith: The `test` function on line 14 is not closed, and the `test` on line 54 is not indented correctly.

JoshSmith:
Your last commits are both named `--amend`. I'm guessing you were trying to amend your commit but instead committed with the message of "--amend".

tests/unit/services/project-user-test.js
```diff
@@ -1,15 +1,86 @@
...
+  };
+
+  set(service, 'currentUser', mockCurrentUser);
+  set(service, 'flashMessages', mockFlashMessage);
+  set(service, 'store', mockStore);
+
+  service.joinProject(project).then(() => {
+    done();
+  });
+
```
JoshSmith: This `test` function is still not closed by a `});` which means you're running right into the next `test` inside this `test` (which itself is not left indented correctly).

PR #1492, opened 2017-11-07
Improve performance of task card

### LINKED ISSUE #1504: Task assignment user icons have an incorrect border-radius
In the `task-assignment` component's dropdown menu the user icons appear to have a `border-radius` for a larger icon width. The `border-radius` needs to be reduced to actually render a circle.

JoshSmith:
# What's in this PR?

This is a spike and should not (yet) be considered production-ready code.

The changes made here have so far improved render of a given task list from ~1500ms to ~500ms, resulting in a 3x performance improvement.

- Replaced `svg/sprite-icon` component on `task-card` with simple HTML
  - removes ~1ms render multiplied by `n` cards
- Does not render `ember-power-select` in `task-assignment` until hovering the card
  - this may have implications for mobile usage that needs considered
  - removes ~10ms render multiplied by `n` cards
- Does not render the `github/issue-link` component until hovering the card and instead renders simple HTML
  - removes < 1ms render multiplied by `n` cards

The `tooltip-on-component` from `ember-tooltip` has similarly deleterious effects on performance when rendering across multiple cards.

One thing that we could do is to change the behavior from hovering to whether the card is in the viewport (and could possibly unrender the complex elements when leaving the viewport, though this unrendering does not work for `mouseLeave` because of other UX considerations).

Much of the mobile behavior changes may depend on whether we can reliably sniff the mobile device. Ideally the behavior would be the same across devices and screen sizes.

Fixes #1504

### DIFF
--- a/app/templates/components/task-card.hbs
+++ b/app/templates/components/task-card.hbs
@@ -19,15 +19,24 @@
     <span class="task-card__time" data-test-task-time>{{moment-from-now task.createdAt}}</span>
     <span class="task-card__integrations" data-test-task-integrations>
       {{#if task.githubIssue}}
-        {{github/issue-link
-          githubIssue=task.githubIssue
-          githubRepo=task.githubRepo
-          size="small"
-        }}
+        {{#if hasHovered}}
+          {{github/issue-link
+            githubIssue=task.githubIssue
+            githubRepo=task.githubRepo
+            size="small"
+          }}
+        {{else}}
+          <span class="sprite-icon">
+            <svg class="solid-light-gray size-16">
+              <use xlink:href="#github-48"></use>
+            </svg>
+          </span>
+        {{/if}}
       {{/if}}
     </span>
   </p>
   {{task-assignment
+    deferredRendering=(not hasHovered)
     task=task
     taskUser=taskUser
     users=users
--- a/app/components/task-assignment.js
+++ b/app/components/task-assignment.js
@@ -11,6 +11,7 @@ export default Component.extend({
   currentUser: service(),
   taskAssignment: service(),
 
+  deferredRendering: null,
   task: null,
   taskUser: null,
   users: null,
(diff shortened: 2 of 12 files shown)


JoshSmith:
@begedin I could use some thoughts on this.

JoshSmith:
It may be worth also testing removing our use of `ember-composable-helpers` and the like for simple computed properties.

app/templates/components/task-card.hbs
```diff
@@ -19,15 +19,24 @@
...
+            <svg class="solid-light-gray size-16">
+              <use xlink:href="#github-48"></use>
+            </svg>
+          </span>
+        {{/if}}
       {{/if}}
     </span>
   </p>
   {{task-assignment
+    hideDropdown=(not hasHovered)
```
begedin: Would call this something like `renderDropdown` or `dropdownReady` or something to avoid a potentially confusing `not`.

begedin:
> Does not render ember-power-select in task-assignment until hovering the card
> - this may have implications for mobile usage that needs considered

Viewport might work, with probably a moderate amount of complexity. Another option might be to somehow trigger rendering followed by the first open on click. Something like 

```Javascript
click() {
  set(this, 'renderDropdown', true);
  run(this, 'afterRender', () => get(this, 'showDropdown')()
}
```
> Does not render the github/issue-link component until hovering the card and instead renders simple HTML

Similar approach might work here, but not sure.

app/templates/components/task-card.hbs
JoshSmith: Maybe `dropdownVisible`?

JoshSmith:
@begedin this is now ready for a code review and can be squashed/merged if approved.

app/templates/components/task-card.hbs
begedin: actually, I'm thinking leave the `not` and call `hideDropdown` something like `dropdownSupressed`. I think it shows intent a bit more, maybe

Possibly even something like `deferredRendering` would maybe work a bit better.

PR #1499, opened 2017-11-08
Add admin UI for GitHub events

### LINKED ISSUE #1498: Create admin UI for viewing paginated list of GitHub API events
# Problem

- Create an `AdminRouteMixin` that checks whether the `currentUser.user` is `admin: true`
- Add a `/admin` route
- Add a `/admin/github` subroute with a list of events that are:
  - Paginated
  - Show the status, type, failure reason, action
- Add a `/admin/github/events/:id` page
  - Show the status, type, failure reason, action, payload (pretty-printed JSON)

JoshSmith:
# What's in this PR?

WIP

The acceptance tests here need expanded and `paginate` and `admin-route-mixin` could use unit tests.

## References
Fixes #1498

### DIFF
--- a/app/models/github-event.js
+++ b/app/models/github-event.js
@@ -1,0 +1,19 @@
+import Model from 'ember-data/model';
+import attr from 'ember-data/attr';
+import { computed, get } from '@ember/object';
+
+export default Model.extend({
+  action: attr(),
+  eventType: attr(),
+  failureReason: attr(),
+  githubDeliveryId: attr(),
+  insertedAt: attr('date'),
+  payload: attr(),
+  status: attr(),
+  updatedAt: attr('date'),
+
+  prettyPayload: computed('payload', function() {
+    let payload = get(this, 'payload');
+    return JSON.stringify(payload, null, 2);
+  })
+});
--- a/mirage/config.js
+++ b/mirage/config.js
@@ -1,5 +1,6 @@
 import Mirage from 'ember-cli-mirage';
 import Ember from 'ember';
+import { paginate } from 'code-corps-ember/mirage/utils/pagination';
 
 const {
   isEmpty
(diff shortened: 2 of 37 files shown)


mirage/config.js
```diff
@@ -408,6 +419,12 @@ export default function() {
 
     let [sluggedRoute] = schema.sluggedRoutes.where({ 'slug': sluggedRouteSlug }).models;
 
+    if (!sluggedRoute) {
+      // If there's no slugged route, we likely added a new route and didn't
+      // `routes` array above.
+      console.error(`Did you add a new route under /${sluggedRouteSlug} (/${projectSlug}) without adding it to mirage/config.js?`);
+    }
```
begedin: I like this 👍

app/models/github-event.js
```diff
@@ -0,0 +1,19 @@
...
+  githubDeliveryId: attr(),
+  insertedAt: attr('date'),
+  payload: attr(),
+  status: attr(),
+  updatedAt: attr('date'),
+
+  prettyPayload: computed('payload', function() {
+    let payload = get(this, 'payload');
+    return JSON.stringify(payload, null, 2);
+  })
```
begedin: I know our thoughts on this probably diverge, but I dislike anything other than pure attributes and relationships on models, unless it significantly dries up the code. In this specific case, a single controller is the only one using `prettyPayload`, so it would make sense to me to have the property there instead of the model.

It's not a deal breaker for me, but it is my preference. I also don't think it matters internally, performance-wise, since a computed generally does not get evaluated unless used directly or indirectly in a template.

mirage/config.js
JoshSmith: I do too after losing 90 mins yesterday figuring out what was wrong. :facepalm:

PR #1523, opened 2017-11-17
Add Github event retrying

JoshSmith:
# What's in this PR?

Adds the ability to retry Github event processing from the admin UI.


JoshSmith:
I disagree since Ecto specifically has virtual attributes for changesets.

begedin:
@joshsmith It's my understanding those virtual attributes are for cases where you need one value to set another value.

In this case, were using a virtual attribute to declare "this is not actually an update of the record, we just want you to perform a specific action using this record as the entry point". 

If it's not an update, it makes more sense to me not to force it into being an update.

PR #1525, opened 2017-11-17
Fix task assignment filtering and focus bug

### LINKED ISSUE #1443: Bug when clicking assign a task and clicking out of it
# Problem

- Open the console
- Click to assign a task
- Dismiss the form
- Type any string
- You should get the following error:

```
Assertion Failed: {{power-select}} If you want the default filtering to work on options that are not plain strings, you need to provide `searchField`
```

### LINKED ISSUE #1524: Task assignment dropdown filter no longer appears to work
# Problem

I'm not sure what changed or why the tests are still passing, but the task assignment dropdown filter is not working.

JoshSmith:
# What's in this PR?

Fixes issues with task assignment filtering and also blurs the trigger when the dropdown is closed to prevent it from acting like a native `select` with focus.

## References
Fixes #1443 
Fixes #1524

### DIFF
--- a/app/components/power-select/before-task-options.js
+++ b/app/components/power-select/before-task-options.js
@@ -1,7 +1,25 @@
-import { get } from '@ember/object';
+import { get, set } from '@ember/object';
+import { on } from '@ember/object/evented';
 import BeforeOptionsComponent from 'ember-power-select/components/power-select/before-options';
+import {
+  EKMixin as EmberKeyboardMixin,
+  keyDown
+} from 'ember-keyboard';
 
-export default BeforeOptionsComponent.extend({
+export default BeforeOptionsComponent.extend(EmberKeyboardMixin, {
+  init() {
+    this._super(...arguments);
+
+    set(this, 'keyboardActivated', true);
+  },
+
+  keydown: on(keyDown(), function(event, ekEvent) {
+    ekEvent.stopPropagation();
+    ekEvent.stopImmediatePropagation();
+    // Send the action ember-power-select expects
+    this.sendAction('onKeydown', event);
+  }),
+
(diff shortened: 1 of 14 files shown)


JoshSmith:
This needs some tests for:

- preventing the card being archived when pressing <kbd>C</kbd> and the assignment dropdown is open
- the power select trigger should not be focused when it's closed again

app/components/task-assignment.js
```diff
@@ -67,6 +71,18 @@ export default Component.extend({
       }
     },
 
+    closeDropdown() {
+      this.sendAction('onclose');
+      let that = this;
+      run.next(that, function() {
```
snewcomer: could this work as `next(that, () =>` and keep the scope so you don't have to have an intermediate variable here?

app/components/power-select/before-task-options.js
```diff
@@ -1,7 +1,25 @@
...
+export default BeforeOptionsComponent.extend(EmberKeyboardMixin, {
+  init() {
+    this._super(...arguments);
+
+    set(this, 'keyboardActivated', true);
+  },
+
+  keydown: on(keyDown(), function(event, ekEvent) {
+    ekEvent.stopPropagation();
+    ekEvent.stopImmediatePropagation();
```
snewcomer: I think this implicitly calls `stopPropagation`

snewcomer: Is the reason for `stopImmediatePropagation` to prevent other events from firing on power-select?

JoshSmith: It doesn't. They're two separate things and explicated more in their docs.

PR #1532, opened 2017-11-18
Improve fixed alerts

joshsmith:
# What's in this PR?

WIP spike of improving the fixed alerts, and adding them to alert us of behaviors on the task board.


joshsmith:
This needs some work to make it so the flash messages can stack vertically from the bottom left corner of the screen.

app/templates/components/flash-messages/fixed-position.hbs
```diff
@@ -0,0 +1,11 @@
+{{#each flashMessages.queue as |flash|}}
```
begedin: Not critical for this PR to be mergeable, but could we not use a filter here?

joshsmith: You mean to filter-by flash.fixed in this case and reject-by in the other?

begedin: Yes, that's what I had in mind.

PR #1536, opened 2017-11-19
Fix status icon to include PR state and icon

joshsmith:
# What's in this PR?

This still needs tests for the new behavior.

The async nature of the `task` => `issue` => `pull_request` makes this a bit sad, too, as it renders in one state and then changes to another. We may want to bring PR up to the level of task or simplify the states on `Task` to account for the PR state rather than this roundabout method.


joshsmith:
@begedin any thoughts on the above?

PR #1547, opened 2017-11-23
Add ember-power-select for changing project categories

joshsmith:
# What's in this PR?

I'm likely to close this because I don't like how it works at all.

I think this is something where we're ultimately better off writing our own components here.


joshsmith:
Yeah I'm going to close this now @begedin. It makes no sense.

I do think there's a strong case for eventually not using power select in the task assignment dropdown. It feels like more work to customize it vs just writing a component ourselves.

PR #1556, opened 2017-11-27
Add organizations/new

### LINKED ISSUE #1542: Add /organizations/new
Part of #1541

We need a route where a user can create an organization, provided they have an invite to create it with.

> Notes are now in PR.

# Estimate

~4 hours

### LINKED ISSUE #1543: On the organizations/new route, load token and prefil form
Part of #1541, enhances #1542, blocked by #1542

While validating the invite token before creating an organization is not strictly necessary, we could use this step to improve UX a bit.

- add `OrganizationInvite` model
- try to load the `OrganizationInvite` via token in `beforeModel`  of `organizations/new`
- If load fails, redirect to root with a flash
- If load passes, `model` hook prefills new `Organization` with name from invite
- add acceptance test for the fail case
- add acceptance test for the success case

~1 hour

joshsmith:
# What's in this PR?

This still a WIP.

Was doing a lot of styling work that I think is mostly there now. What remains is everything else. 😄 

- [x] Need `code` and `organization_name` as query parameters
- [x] Code is used to verify if we can create an organization (in `beforeModel`)
  - [x] we should validate code initially and redirect somewhere else if it fails
  - [x] we can prefill the form with the `organization_name` #1543
- [x] Submitting form hits create route on API
- [ ] Successful save redirects to `/org-slug/projects/new`

### Additional requirements

- [ ] use `ember-concurrency` for saving
- [x] failed saved due to validation errors shows those errors
- [x] failed save due to invalid token shows flash
- [x] needs some copy to make the onboarding friendlier
- [x] maybe needs copy emphasizing this is not project creation yet
- [x] native confirm dialog when trying to leave the page to prevent losing information (maybe we should add a mixin for this?)

### Required tests

- [x] acceptance test for token redirection 
- [x] acceptance test for successful save 
- [x] acceptance test for failed save due to validations
- [ ] acceptance test for failed save due to token

### Notes

- [ ] will require an API issue to implement the create route - code-corps/code-corps-api#1229

## References
Fixes #1542 
Fixes #1543

### DIFF
--- a/app/routes/organizations/new.js
+++ b/app/routes/organizations/new.js
@@ -1,0 +1,75 @@
+import Route from '@ember/routing/route';
+import { inject as service } from '@ember/service';
+import { get, set } from '@ember/object';
+import { isPresent } from '@ember/utils';
+import AuthenticatedRouteMixin from 'ember-simple-auth/mixins/authenticated-route-mixin';
+import { resolve } from 'rsvp';
+
+export default Route.extend(AuthenticatedRouteMixin, {
+  currentUser: service(),
+  flashMessages: service(),
+
+  queryParams: { code: 'code' },
+
+  async beforeModel(transition) {
+    if (get(this, 'session.isAuthenticated')) {
+      let invite = await this._loadInvite(transition);
+      // Pass data along into the `model` hook
+      return set(transition, 'data', { organizationInvite: invite });
+    } else {
+      // call `beforeModel` in `AuthenticatedRouteMixin`
+      return this._super(transition);
+    }
+  },
+
+  async model({ code }, transition) {
+    let organizationInvite = get(transition, 'data.organizationInvite');
+    if (isPresent(organizationInvite)) {
+      let organization = await this._initOrganization(organizationInvite, code);
+      return { organization, organizationInvite };
+    } else {
+      return { organization: null, organizationInvite: null };
+    }
+  },
+
+  setupController(controller, { organization, organizationInvite }) {
+    controller.set('organization', organization);
+    controller.set('organizationInvite', organizationInvite);
+  },
+
+  async _loadInvite(transition) {
+    let { queryParams: { code } } = transition;
+    if (code) {
+      let invites = await get(this, 'store').query('organization-invite', { code });
+      return get(invites, 'firstObject');
+    } else {
+      return resolve(null);
+    }
+  },
+
+  async _initOrganization(organizationInvite, inviteCode) {
+    let name = get(organizationInvite, 'name');
+    let owner = await get(this, 'currentUser.user');
+    return get(this, 'store').createRecord('organization', { name, owner, inviteCode });
+  },
+
+  actions: {
+    willTransition(transition) {
+      let organization = get(this, 'controller.organization');
+      let inviteError = get(this, 'controller.inviteError');
+      // prompt to confirm if the user did not save
+      if (!isPresent(inviteError) && get(organization, ('isNew'))) {
+        let confirmed = window.confirm('You will lose any unsaved information if you leave this page. Are you sure?');
+        if (confirmed) {
+          organization.destroyRecord();
+        } else {
+          transition.abort();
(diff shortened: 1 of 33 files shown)


joshsmith:
Right now the API's invite email goes to `/invites/organization?code=[CODE]&organization_name=[NAME]`. We'll need to change that to `/organizations/new?code=[CODE]&organization_name=[NAME]`, validate the `code`, and use the `organization_name` (should we dasherize this param? or make it `organization[name]`? idk best JSON API-like standard to abide) to prefill the organization's `name`.

begedin:
@joshsmith Made some progress, but had complications, so not fully done. Some notes:

1. I'm performing validation by checking the API for the organization invite record by code. For that purpose, I'll push a PR to the API which allows filtering by `code`.
2. We don't need an organization_name query param. Since we fetch the invite record, it provides us with that value directly instead.
3. I'm not sure handling of bad token submission requires any additional code. It should be reported as a validation error, other than adding the field.
4. We **do** either need to submit the code with the form, so we can find the associated invite on the API and fulfill it, or we need to fulfill it from the client manually (IMO, unsafe), or we need to associate the organization with the invite and use the association to fulfil it.
5. We're not redirecting to the correct route on save yet. We can do that when we add the route in another PR.

joshsmith:
I’m guessing you’re tracking the eslint errors?

begedin:
# TODO

- [ ] Add a code attribute to organisation (virtual), set it from the invite code, so organization is saved with it
- [ ] Display the validation error for the "hidden" code attribute, in case validation fails with an error on that field.

joshsmith:
The form no longer redirects when the code is missing/invalid. Instead, it gives the user a chance to fix the situation by putting in the correct code. (Perhaps they mangled the URL).

The `organizationInvite` is now also decoupled from the `organization` record that's created, since we need to check the organization relationship in the invite; otherwise, we would unset that organization value. It's also not quite representative of the relationship between the two. We might want to even consider simply passing in the ID rather than the invite itself.

The overall behavior is there now. The templates could maybe be cleaned up slightly or have things extracted out into components where necessary, but we should first fix the tests to validate this behavior, since manually it is verified as accurate.

begedin:
Realized this just as you went to bed, but we can't create the organization by specifying invite id on the client. It's unsafe.

When going through the client, it works. Invite code is checked and validated first, then the user is shown the organization form. User submits form, associations are dealt with and that's it.

However, when hitting the API directly, any user would be able to create an organization simply by guessing an invite id. Since all of our invites are `bigint`,  that would be relatively easy to guess. 

Sure, we can still make it so the guessed invite has to be unfulfilled and we still need to approve an organization created in such a way, but it could still be used against our system in a slightly malicious way.

begedin:
The solution I'm implementing is

- we don't set the invite id client-side. It adds confusion anyway, since we explicitly invert the relationship on the API
- organization gets an `inviteCode` virtual attribute, we set and send that API side
- api authorizes by fetching invita via code and making sure it's not fulfilled (org id not set)
- api creates organization and, as part of a multi, fetches invite via code and associates with organization

begedin:
@joshsmith This should be ready for another review and it matches the API PR code-corps/code-corps-api#1262

app/routes/organizations/new.js
```diff
@@ -0,0 +1,75 @@
...
+          organization.destroyRecord();
+        } else {
+          transition.abort();
+        }
+      }
+    },
+
+    verifyInvite() {
+      this.refresh();
+    }
```
begedin: Tested this out manually, and I'm not seeing the UI difference in the way how this behaves compared to an action which explicitly duplicates the `beforeModel/model` behavior we already have.

I think we should also familiarize ourselves with loading substate routes and `intermediateTransisitionTo`. It may help us in dealing with our loading UI.

joshsmith:
Needs tests for:

- when it's already fulfilled
- image upload
- failed save flash message

PR #1564, opened 2017-11-29
Add organization invites admin ui

### LINKED ISSUE #1563: Add admin UI to create organization invites
# Problem

We need to create an admin UI to:

- view previously sent invites
- create new invites
  - with an email
  - with an (optional) organization name

joshsmith:
# What's in this PR?

Fixes #1563


app/routes/admin/organization-invites/new.js
```diff
@@ -0,0 +1,28 @@
...
+      // prompt to confirm if the user did not save
+      if (get(organizationInvite, ('isNew'))) {
+        let confirmed = window.confirm('You will lose any unsaved information if you leave this page. Are you sure?');
+        if (confirmed) {
+          organizationInvite.destroyRecord();
+        } else {
+          transition.abort();
+        }
+      }
+    }
```
begedin: This seems mixin worthy. Maybe we should create an issue

```Javascript
export default Ember.Mixin.extend({;
  modelName: null,
  message: 'You will lose any unsaved information if you leave this page. Are you sure?',
  
  actions: {
    let modelName = get(this, 'modelName');
    if (isPresent(modelName)) {
       let model = get(this, modelName);
       if (get(modelName, 'isNew') {
         this._confirmTransition(transition, model);
       } 
    } else {
      this._confirmTransition(transition);
    }
  },

  _confirmTransition(transition, model) {
    let message = get(this, 'message');
    if (window.confirm(message)) {
      if isPresent(model) {
        model.destroyRecord();
      }
    } else {
      transition.abort();
    }
  }
})

```

Come to think of it, do we really need to check for `isNew` on a `.new` route? Ok, the `organization.new` has that case, but I'm not sure it's strictly necessary there either. If we're still on the route, we're doing something we don't want to lose. If we aren't doing that, we should've gotten redirected.

joshsmith: The willTransition will fire on that redirect, though. Even if it’s successful.

PR #1565, opened 2017-11-29
initial commit with donation components

ilovethebends:
# What's in this PR?

- Add-on Ember Freestyle
- Added donation components

### DIFF
--- a/app/controllers/freestyle.js
+++ b/app/controllers/freestyle.js
@@ -1,0 +1,18 @@
+import Ember from 'ember';
+import FreestyleController from 'ember-freestyle/controllers/freestyle';
+
+const { inject } = Ember;
+
+export default FreestyleController.extend({
+  emberFreestyle: inject.service(),
+  card: {
+    id: 1, brand: 'Visa', last4: 4242, expMonth: '01', expYear: '2022'
+  },
+  saveAndDonateHandler() {},
+  donateHandler() {},
+  project: {
+    title: 'Test Project',
+    description: 'Test project description',
+    iconThumbUrl: 'icon.png'
+  }
+});
--- a/app/router.js
+++ b/app/router.js
@@ -35,6 +35,8 @@ Router.map(function() {
     });
   });
 
+  this.route('freestyle');
+
   // GitHub OAuth redirection route
   this.route('github', {
     path: '/oauth/github'
--- a/app/routes/freestyle.js
+++ b/app/routes/freestyle.js
@@ -1,0 +1,4 @@
+import Route from '@ember/routing/route';
+
+export default Route.extend({
+});
--- a/app/styles/_freestyle.scss
+++ b/app/styles/_freestyle.scss
@@ -1,0 +1,19 @@
+.card-logo {
+  &__container {
+    top: inherit;
+    left: inherit;
+    position: inherit;
+  }
+
+  &__inner {
+    position: inherit;
+  }
+}
+
+code {
+  background: none;
+}
+
+.donation-container {
+  float: inherit;
+}
--- a/app/styles/app.scss
+++ b/app/styles/app.scss
@@ -78,6 +78,8 @@
 //
 @import "addons/ember-stripe-elements/stripe-element";
 
+
+
 //
 // COMPONENTS
 //
@@ -255,3 +257,5 @@
 // SHAME.CSS
 //
 @import "shame";
+
+@import "freestyle";
--- a/app/templates/freestyle.hbs
+++ b/app/templates/freestyle.hbs
@@ -1,0 +1,113 @@
+{{#freestyle-guide
+  title="Ember Freestyle"
+  subtitle="Living Style Guide"
+  }}
+
+  {{#freestyle-section name="Donation" as |section|}}
+
+    {{#section.subsection name="Card Item"}}
+      {{#freestyle-usage "card-item" name="Card Item"}}
+        {{donation/card-item card=card}}
+      {{/freestyle-usage}}
+    {{/section.subsection}}
+
+    {{#section.subsection name="Card Logos"}}
+      {{#freestyle-collection title="Card Logos " defaultKey="all" inline=true as |collection|}}
+        {{#collection.variant key="American Express"}}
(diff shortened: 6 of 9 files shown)


ilovethebends:
@joshsmith 
- definitely feel like this will make things easier to get the ball rolling. Using freestyle gives us a nice framework for components to be put in. 
- Somethings are a little opinionated, but sure we can work around. 
- Not sure how I feel about everything being inside one template file. Might be hard to collaborate?

joshsmith:
Everything in one template definitely would make things a challenge.

Is there also a way to prevent freestyle from being deployed (including the assets) alongside our application? i.e. I wouldn't want everything that's intended for use for local development to be made available in our public site, nor burden users with downloading a bunch of CSS/JS they don't need.

ilovethebends:
Yeah we can do that. It'll be similar to the previous styleguide.

ilovethebends:
@joshsmith 
- Even though it's all inside one template, each section can be inside it's own separate component. Which makes things more maintainable. `sg/folderName--component-name`
- As for assets, the freestyle addon can be blacklisted, but I'm not sure about removing the components code 

screen shot 2017-12-04 at 10 25 54 pm

ilovethebends:
@snewcomer sorry for long delay and no response. Hopefully you're still interested in helping out. :)
I closed the previous styleguide branch in favor for this one using ember-freestyle instead of starting from scratch. 

@joshsmith i was able to move the styleguides into it's own in-app repo so it can be blacklisted. 
I think we're ready to go?

lib/styleguide/app/components/styleguide/donation/card-item.js
```diff
@@ -0,0 +1,19 @@
...
+```
+import Ember from 'ember';
+
+export default Ember.Component.extend({
+  card: {
+    id: 1, brand: 'Visa', last4: 4242, expMonth: '01', expYear: '2022'
+  }
+});
+```
+END-FREESTYLE-USAGE */
```
begedin: I wasn't really part of the previous conversation. Has it been decided that this is a note we want to render. The way I see it, something like markdown explaining what the component attributes are would be of more help.

In some other component, the markdown could explain which actions are triggered, etc.

In this case, for example, we could have markdown explaining what the internal structure of `card` is.

This is OK to, if that was the decision. I just feel like we could do more.

snewcomer:
@ilovethebends Awesome work!  I love it so far.  A few things...do the tests need to be committed/should there be tests?  Also, is the idea to document every component in the application or just ones that we think are reused more than once (like a button style, link hover, etc)?

lib/styleguide/app/components/styleguide/donation/card-item.js
ilovethebends: @begedin 
Main thing for this PR was a proof of concept it could be added as an addon and be blacklisted to production. 

- Yes, we can just keep the markup to explaining the attributes and actions. 
- If more documentation is need more clarification maybe that could be part the PR?
- I'll keep building more components.

ilovethebends:
@snewcomer 
Unless anyone can think of a reason, I don't see a reason for tests. 

There is no process of which components should or shouldn't be documented. Right now i'm just going through the tests folder to see what I can come up with.

tests/integration/components/github/connected-installation-test.js
```diff
@@ -96,7 +96,7 @@ test('triggers/passes out all actions', function(assert) {
   });
 
   renderPage();
-
+  debugger;
```
joshsmith: Forgotten `debugger`, FYI.

joshsmith:
I'm seeing some "usage" notes in here that don't seem to represent actual usage but things about the test setup.

Is there a better way we can be handling this?

ilovethebends:
@joshsmith I could just copy the actual component.js in the usage?
There was some async stuff i tried to simulate. ie loading repo, syncing repo etc..

joshsmith:
No I mean, isn't the usage supposed to be the handlebars? Or am I confused?

ilovethebends:
Okay, so now i'm confused too. 

I was assuming that usage is both handlebars and data that gets passed in, and how actions might be used? That's why each of the actual components is wrapped inside of a sandbox?

screen shot 2017-12-16 at 10 17 41 am

Sorry if i'm misunderstanding you.
Usage "notes" can come from using either: 

1.  the .hbs `{{#freestyle-note "globally-unique-slug--notes"}}` (not used in this example)
2. or the .js `/* BEGIN-FREESTYLE-USAGE globally-unique-slug--notes`

In the example above "Connected Installation" & "Example Data" code snippets both are coming from the commented section in the .js file. `styleguide/components/github/connected-installation.js`

The bottom half is where the component is rendered and the handlebars snippet is from the .hbs file. 

This is from inside the "usage" helper `{{#freestyle-usage 'globally-unique-slug`

##### styleguide/templates/components/github/connected-installation.hbs
```
{{#freestyle-usage 'github--connected-installation' title='Connected Installation'}}
  {{github/connected-installation
    disconnect=(action disconnect)
    connectRepo=(action connectRepo)
    disconnectRepo=(action disconnectRepo)
    organizationGithubAppInstallation=organizationGithubAppInstallation
    project=project
  }}
{{/freestyle-usage}}
```

I could just remove all the notes and just render the component and code snippet if you think that's more clear?

joshsmith:
@begedin could we have your thoughts here on this?

joshsmith:
To me the bottom code is useful (the `hbs`) to render, whereas the top part (sandboxed data) being rendered feels not as useful and perhaps even confusing to someone reading the docs expecting to see how it's used.

joshsmith:
For me, pending that change, getting tests passing again, and giving me some time to take it for a spin, I would then consider this mergeable and we can create a tracking issue for the components we'd like to prioritize next.

joshsmith:
I'd also love to see a follow-up issue to expand our documentation (perhaps with a link to the wiki for easier doc-writing) to include some thoughts on how we intend to use this and how it can make the development lifecycle easier.

joshsmith:
Looks like test failures are just `eslint`, btw.

PR #1566, opened 2017-11-29
Add project new

### LINKED ISSUE #1544: Add /org-slug/projects/new
We need a route where an organization owner can create a new project.

- add route
- authorization checks if user is owner of organization
- `model` hook inits the project
- we should check that it's possible to redirect to this route by providing a pre-built project model. If I'm not mistaken, this ought to skip the `model` hook
- route is basic, just renders the form with `title`, `description`, `long_description_markdown`,  `website`
- successful save redirects to `/org-slug/project-slug/`
- failed save renders errors
- required acceptance tests
  - requires authentication
  - requires user to be organization owner
  - successful save
  - failed save

# Questions

Presumably, the initially sent invite is intended to allow for one project.

My guess is, we want to use the `approved` field on the project so admins have to explicitly approve each new project. Is the initially created project auto-approved, or is it still required to be explicitly approved by an admin? Either way, that would probably be done on the side of the API, so we should open an issue to discuss.

From the model on the API side, I assume `title`, `description` are required, the rest is not, but maybe it ought to be?

# Estimate

~2 hours

begedin:
Closes #1544

Adds project creation for organizations.


app/controllers/projects/new.js
```diff
@@ -0,0 +1,119 @@
...
+export default Controller.extend({
+  flashMessages: service(),
+  loadingBar: service(),
+  projectSkillsList: service(),
+  store: service(),
+
+  project: null,
+  showCallout: true,
+
+  isFirstProject: lte('organizationProjectsCount', 1),
```
begedin: Was wondering if we'd do something like this, but I was also wondering if se should limit the number of projects per organization initially? Like, add an actual field, `organization.projectLimit`, which defaults to 1 initially and we can increase through the admin  UI?

That would mean server-side validations.

joshsmith: I'm not too concerned since we still have to approve the projects.

app/services/project-skills-list.js
```diff
@@ -25,6 +34,8 @@ export default Service.extend({
 
   remove(skill) {
     let projectSkill = this.find(skill);
+    let project = get(this, 'project');
+    project.get('skills').popObject(skill);
     return projectSkill.destroyRecord();
```
joshsmith: I don't like that this is necessary to do now either.

app/services/project-skills-list.js
```diff
@@ -6,13 +6,22 @@ import recordsList from 'code-corps-ember/utils/records-list';
...
   add(skill) {
     let { store, project } = getProperties(this, 'store', 'project');
     return store.createRecord('project-skill', { project, skill }).save();
   },
 
+  addWithoutSave(skill) {
+    let { store, project } = getProperties(this, 'store', 'project');
+    let projectSkill = store.createRecord('project-skill', { project, skill });
+    project.get('skills').pushObject(skill);
+    return projectSkill;
```
joshsmith: I don't like that this is necessary to do now.

app/services/project-skills-list.js
```diff
@@ -42,7 +53,25 @@ export default Service.extend({
...
     return get(this, 'projectSkills').reload();
+  },
+
+  _toggle(skill, add) {
+    let projectSkills = get(this, 'projectSkills');
+    if (recordsList.includes(projectSkills, skill)) {
+      return this.remove(skill);
+    } else {
+      return add(skill);
+    }
```
joshsmith: Can't figure this out. Want to reduce the duplication of the toggle fns above.

app/templates/projects/new.hbs
```diff
@@ -0,0 +1,164 @@
...
+  <div class="panel">
+    <section>
+      <aside>
+        <h3>{{fa-icon "address-card"}} Your project's details</h3>
+        <p>Projects are where you'll find volunteers, manage tasks, and accept donations.</p>
+      </aside>
+
+      <section data-test-project-form>
+        {{#if showCallout}}
+          {{#if isFirstProject}}
```
joshsmith: Probably could do an `(and showCallout isFirstProject)` here.

app/templates/projects/new.hbs
```diff
@@ -0,0 +1,164 @@
...
+        </div>
+
+        <div class="input-group">
+          <div class="input-explain">
+            <p>Ready to submit your project?</p>
+          </div>
+          <button class="save default" {{action 'save'}} data-test-submit>Create project</button>
+          {{#if project.errors}}
+            <p class="input-error">There were some errors when trying to create the project.</p>
+          {{/if}}
```
joshsmith: Should add this to the other form.

app/utils/records-list.js
```diff
@@ -50,6 +50,12 @@ export default {
         let targetModelName = get(target, 'constructor.modelName')
                            || get(target, 'content.constructor.modelName');
 
+        // If the relationship isn't there yet
+        if (!relationshipId && relationshipModelName && targetId && targetModelName) {
+          let foundId = found.belongsTo(targetModelName.camelize()).id();
+          return foundId == targetId;
+        }
```
joshsmith: Not sure if I'm a fan of having to do this. Definitely needs a test. But only way I could get this to work for hasMany relationships where one model `isNew`.

app/routes/projects/new.js
```diff
@@ -0,0 +1,47 @@
...
+  },
+
+  actions: {
+    willTransition(transition) {
+      let project = get(this, 'controller.project');
+      // prompt to confirm if the user did not save
+      if (!isPresent(project) && get(project, ('isNew'))) {
+        let confirmed = window.confirm('You will lose any unsaved information if you leave this page. Are you sure?');
+        if (confirmed) {
+          project.destroyRecord();
```
joshsmith: This does not appear to be working as expected.

app/controllers/projects/new.js
```diff
@@ -0,0 +1,119 @@
...
+  actions: {
+    copyOrganizationData(organization) {
+      let { cloudinaryPublicId, description, iconLargeUrl, name, slug }
+        = getProperties(organization, 'cloudinaryPublicId', 'description', 'iconLargeUrl', 'name', 'slug');
+      let project = get(this, 'project');
+      let attrs = { cloudinaryPublicId, description, iconLargeUrl, slug, title: name };
+      setProperties(project, attrs);
+      set(this, 'showCallout', false);
+      get(this, 'flashMessages').clearMessages().success('Your project now has the details from your organization.');
+    },
```
joshsmith: This whole thing feels complex enough to warrant extracting the form out into a component. Thoughts here? It will get even worse with client-side validations.

We could also try to reduce some of the acceptance tests with more usage of route/controller unit tests.

begedin: If you feel it's worth the potential speed gain of integration tests instead of acceptance tests to invest the time right now, I'm ok with it. 

I'm hesitant to abandon the base acceptance tests in favor of faster integration tests. Acceptance tests are better at ensuring the overall app behavior works and we're risking not detecting some layer not working as expected when it interfaces with another layer.

As an example, I think we need an acceptance test for validation errors, but we could use integration tests to ensure errors on all supported fields are rendered.

For navigation confirmation tests alone, it's difficult to do integration tests well, without mocking large parts of the app, which sort of defeats the purpose.

app/routes/projects/new.js
begedin: `!isPresent(project)` is the problem, I'm guessing.

app/services/project-skills-list.js
begedin: Considering it's a `many-to-many` now I'm not sure there's a need for this service.

When creating the project, we assign to the `many-to-many` directly.
When updating the project, we create individual `projectSkill` records. We _could_ assign and remove from the `many-to-many` to, but at the moment, we don't and I'm not sure it would be an improvement.

As far as I know, those are our only two scenarios for working with project skills. Considering that, this doesn't really dry up our code. It appears to do so, but under the hood, it's just trying to fit two different chunks of code together.

app/controllers/projects/new.js
joshsmith: Not saying we abandon the acceptance tests, but all the edge cases would be much simpler to test with integration tests.

PR #1572, opened 2017-12-01
Improve donation goal form

joshsmith:
# What's in this PR?

Some minor design improvements to the donation goal form.


begedin:
Fixed the test here. Looks good to go

PR #1576, opened 2017-12-05
Added admin UI for project approval

### LINKED ISSUE #1575: Add Admin UI for project approval
# Problem

We need an Admin UI to view projects and approve them

## Subtasks
- [x] Add projects tab
- [x] Lists all projects, each linking to their show page
- [x] Unapproved projects have an "approve" button

begedin:
# What's in this PR?

Closes #1575 

Still in the process of wrapping up, but ought to be ok for review.

### DIFF
--- a/app/router.js
+++ b/app/router.js
@@ -36,6 +36,7 @@ Router.map(function() {
     this.route('organization-invites', function() {
       this.route('new');
     });
+    this.route('projects', function() {});
   });
 
   // GitHub OAuth redirection route
--- a/app/routes/admin/projects/index.js
+++ b/app/routes/admin/projects/index.js
@@ -1,0 +1,7 @@
+import Route from '@ember/routing/route';
+
+export default Route.extend({
+  model() {
+    return this.store.findAll('project');
+  }
+});
--- a/app/templates/admin/projects/index.hbs
+++ b/app/templates/admin/projects/index.hbs
@@ -1,0 +1,51 @@
+<div class="log-rows">
+  <div data-test-log-row-header class="log-row log-row--header">
+    <span class="log-cell log-cell--shrink"></span>
+    <span class="log-cell log-cell--shrink">Project</span>
+    <span class="log-cell log-cell--shrink">Organization</span>
+    <span class="log-cell">Status</span>
+    <span class="log-cell log-cell--shrink">Approval</span>
+  </div>
+
+  {{#each model as |project|}}
+    <div data-test-log-row class="log-row">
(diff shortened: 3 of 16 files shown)


app/router.js
```diff
@@ -36,6 +36,7 @@ Router.map(function() {
     this.route('organization-invites', function() {
       this.route('new');
     });
+    this.route('projects', function() {});
```
joshsmith: I think the `function` is unnecessary here.

app/routes/admin/projects/index.js
```diff
@@ -0,0 +1,7 @@
+import Route from '@ember/routing/route';
+
+export default Route.extend({
+  model() {
+    return this.store.findAll('project');
```
joshsmith: This is fine for now, though we'll need to eventually paginate it.

app/router.js
begedin: I've been meaning to comment on that.

Without the `function`, there's no `projects.index` route being autogenerated, so we only get the `projects` route, meaning the file structure is not consistent with github events and organization invites.

Really, that would be fine, but my assumption is, we'll soon have other subroutes of projects anyway, so leaving this here might save us some time.

I could

- leave this here
- leave this here and add a comment to explain why it's here
- remove this and 
  - move `routes/admin/projects/index.js` -> `/routes/admin/project.js/`
  - move `controllers/admin/projects/index.js` -> `/controllers/admin/project.js/`
  - move `tests/pages/admin/projects/index.js` -> `tests/pages/admin/projects.js`

app/routes/admin/projects/index.js
begedin: Probably, same as with org invites, albeit probably sooner than that.

app/router.js
joshsmith: I think fine leaving it for now.

app/templates/admin/projects/index.hbs
```diff
@@ -0,0 +1,51 @@
...
+        {{else if project.approvalRequested}}
+          {{fa-icon "question-circle"}} Pending approval
+        {{else}}
+          {{fa-icon "asterisk"}} Newly created
+        {{/if}}
+      </span>
+      <span data-test-actions class="log-cell log-cell--shrink">
+        {{#if project.approvalRequested}}
+          <button {{action approve project}} class="default small">Approve</button>
+        {{/if}}
```
joshsmith: This logic means that you see the approval button simply if `project.approvalRequested` is `true`, but will still show if `project.approved` is `true`.

PR #1597, opened 2017-12-11
Conversations Feature

### LINKED ISSUE #1592: Implement project /messages route
# Problem

From https://github.com/code-corps/code-corps-api/issues/1234#issuecomment-349187243

#### Messages for projects

- Left sidebar list of messages
  - photo
  - name
  - last message time
- Center main area individual message
  - show conversation parts
    - photo
    - body
    - sent at
    - read at / not seen yet
- Right sidebar user info
  - photo
  - name
  - skills

### LINKED ISSUE #1593: Implement user /messages route
# Problem

From https://github.com/code-corps/code-corps-api/issues/1234#issuecomment-349187243

#### Messages for users

- Left sidebar list of messages
  - photo
  - name
  - last message time
  - project icon + title
- Center main area individual message
  - show conversation parts
    - photo
    - body
    - sent at
    - read at / not seen yet
- Right sidebar project info
  - icon
  - title
  - skills

### LINKED ISSUE #1594: Shift focus of project contributors page UI towards messages
# Problem

From https://github.com/code-corps/code-corps-api/issues/1234#issuecomment-349187243

We need to remove the skills UI from the project contributor list page to reduce complexity of the page and shift focus towards messages

We also need to modify the contributor list items to show
- photo
- name
- button to start/view the conversation with

### LINKED ISSUE #1596: Support realtime update of messages/conversations
# Problem

From https://github.com/code-corps/code-corps-api/issues/1234#issuecomment-349187243

This is probably to be moved into a separate milestone (for enhancement), but we would like to have support for real time updates of projects messages.

For that, we would likely need support of realtime communication with the API via phoenix channels.

begedin:
The API changes required for this to work are in PR code-corps/code-corps-api#1301

- [x] rewrite to use modal for creation (see below)
- [x] test conversation-thread component
- [x] test conversation-composer component
- [x] test new-conversation-modal component
- [x] update conversation-part component tests
- [x] test channels
- [x] write (small) acceptance tests
  - [x] /project/people
  - [x] /project/conversations (list, view, reply) - #1626 
  - [x] /conversations (list, view, reply) - #1626
- [x] allow for closing conversations
- [x] update socket connection to change when user becomes authenticated/unauthenticated (probably by disconnecting and reconnecting)
- [ ] allow for reopening conversations
- [ ] add filtering conversations by open/closed

Closes #1592 
Closes #1593 
Closes #1594 
Closes #1596


begedin:
I pushed a big commit, because I'm afraid there was a bit of a snowball effect of things that needed to work to make the PR self-contained. Will add comments to the code where applicable.

app/abilities/project.js
```diff
@@ -12,6 +12,15 @@ import projectMember from 'code-corps-ember/macros/project-member';
...
   currentUser: service(),
 
+  /**
+   * An `ember-can` ability.
+   *
+   * Indicates if the current user can manage a project.
+   * Returns true if the user is the owner of the project.
+   * @type {Boolean}
+   */
+  canAdminister: or('{userIsAdmin,userIsOwner}'),
```
begedin: Until this PR, all of the bigger actions on a project were accessible by owners only. The only thing admins could do was approve pending members.

app/controllers/project/messages/message.js
```diff
@@ -0,0 +1,6 @@
+import Controller from '@ember/controller';
+import { alias } from '@ember/object/computed';
+
+export default Controller.extend({
+  conversation: alias('model')
```
begedin: ember inspector expects a `model` property on the controller, so I started using this alias instead of `setupController` we usually do. If we don't want this, I'm fine with reverting to `setupController`

app/controllers/project/messages/new.js
```diff
@@ -0,0 +1,7 @@
+import Controller from '@ember/controller';
+import { alias } from '@ember/object/computed';
+
+export default Controller.extend({
+  message: alias('model'),
+  user: alias('model.conversations.firstObject.user')
```
begedin: This, like many other places in the code, assumes a single user per message, meaning a single conversation as well.

As we add UI support for multiple users, we will need to update this.

app/router.js
```diff
@@ -65,9 +65,13 @@ Router.map(function() {
 
   this.route('project', { path: '/:slugged_route_slug/:project_slug' }, function() {
     this.route('checkout'); // Where you enter your credit card details
+    this.route('contributors');
```
begedin: `contributors` was under `settings`, meaning, since `settings` is owner only, `contributors` was also owner only. To make it `admin` or `owner`, I had to move it.

If we move it back, each settings subroute will have to implement it's own `CanMixin`, while the base `settings` route should not have it.

If we want it here, but keep the old `/:orgslug/slug/settings/contributors` url, we can just set the path here.

Personally, I prefer the new placement and the new `:orgslug/:slug/contributors` url. I would even move `donations` in the same way

app/routes/project/messages.js
```diff
@@ -0,0 +1,45 @@
...
+  },
+
+  async model() {
+    console.log('model');
+    let projectId = get(this.modelFor('project'), 'id');
+    let store = get(this, 'store');
+    // peekAll returns a live array, so it will alse render any newly created
+    // messages, causing the list to update properly on project.messages.new
+    await store.query('conversation', { project_id: projectId });
+    return store.peekAll('conversation');
```
begedin: We have a `list-detail` UI here

- a list of conversations in the sidebar on `/messages/`
- clicking one of the items navigates to `/messages/:id`, and renders the sidebar in the outlet of the main area, while still displaying the list
- navigating to `/messages/new` does the same

This model hook allows the store to detect a new conversation record and show it in the list (styled differently). 

It's hacky, but it's the only way I could find to do this.

An alternative would be to clone the query result of the `project.messages` model into an array, then render that array in controller, instead of the query result directly. 

That would allow us to push the new conversation into that array during the `messages.new` `afterModel` hook. 

Pushing into the query result directly, which is what I initially tried, does not seem to be possible without resorting to using private functions and properties of `DS.model`

app/routes/project/messages/new.js
```diff
@@ -0,0 +1,66 @@
...
+    });
+
+    let message =  store.createRecord('message', {
+      author: currentUser,
+      initiatedBy: 'admin',
+      project
+    });
+
+    let conversation = store.createRecord('conversation', { user });
+    get(message, 'conversations').pushObject(conversation);
```
begedin: This causes the conversation to appear in the list of conversations in the sidebar.

app/routes/project/messages/new.js
```diff
@@ -0,0 +1,66 @@
...
+    return message;
+  },
+
+  actions: {
+    save(message) {
+      // Since it's created without an ID, the store will not detect the
+      // association was saved and will keep it around.
+      // We need to store a reference to it, so we can unload it on save
+      let unsavedConversation = get(message, 'conversations.firstObject');
```
begedin: Comment explains it. Since we store save the two records (message and conversation) together as one, ember does not detect the conversation child has changed, so we need to unload it manually so it gets replaced with the saved record.

app/routes/project/messages/new.js
```diff
@@ -0,0 +1,66 @@
...
+      let conversation = get(message, 'conversations.firstObject');
+      // prompt to confirm if the user did not save
+      if (window.confirm('You will lose any unsaved information if you leave this page. Are you sure?')) {
+        if (get(message, 'isNew')) {
+          message.destroyRecord();
+        }
+
+        if (get(conversation, 'isNew')) {
+          conversation.destroyRecord();
+        }
```
begedin: This will, unfortunately, mess with our proposed mixin.

app/serializers/message.js
```diff
@@ -0,0 +1,18 @@
...
+    conversations: { serialize: true }
+  },
+
+  serialize(snapshot) {
+    let json = this._super(...arguments);
+
+    // support saving child conversations alongside record
+    if (!json.id) {
+      json.included = snapshot.hasMany('conversations').map((c) => c.serialize());
+    }
```
begedin: This is how we include the conversation alongside the message in our payload when creating a message.

Apparently, ember data does not do this automatically and provides no configuration in order to do it, so I had to provide a manual implementation. It's according to spec, as far as I know, but the sad reality is, this whole process involves custom code both API and client side.

app/models/conversation-part.js
```diff
@@ -0,0 +1,11 @@
+import Model from 'ember-data/model';
+import attr from 'ember-data/attr';
+import { belongsTo } from 'ember-data/relationships';
+
+export default Model.extend({
+  body: attr(),
+  readAt: attr(),
+
+  author: belongsTo('user', { async: true }),
```
snewcomer: I think async: true is default...but I also like the explicitness...

app/routes/project/messages.js
```diff
@@ -0,0 +1,45 @@
...
+    if (isAuthenticated) {
+      return this._ensureUserHasCredentials(...arguments);
+    } else {
+      // call `beforeModel` in `AuthenticatedRouteMixin`
+      return this._super(...arguments);
+    }
+  },
+
+  async model() {
+    console.log('model');
```
snewcomer: console

app/router.js
joshsmith: I agree with this move.

app/models/conversation-part.js
joshsmith: It wasn't always, and I was about to ask if it is now finally the default.

app/serializers/message.js
snewcomer: This looks about right!  Is there an equivalent `conversation_part` on the ember side?

app/router.js
```diff
@@ -65,12 +65,12 @@ Router.map(function() {
...
-    this.route('contributors');
-    this.route('donate'); // Where you choose your donation amount
-    this.route('messages', function() {
-      this.route('message', { path: '/:id' });
+    this.route('conversations', function() {
+      this.route('conversation', { path: '/:id' });
       this.route('new');
     });
+    this.route('donate'); // Where you choose your donation amount
+    this.route('people');
```
joshsmith: I've reworked the conventions to be:

- `/contributors` => `/people`
- `/messages` => `/conversations`

The latter is more appropriate because `/messages` would actually be a different route entirely. These would be the automated/manual messages the project has sent/is actively sending to users. These would be more like `/messages/auto` and `/messages/manual`, anyway.

`/converations` serves more as a unified inbox for the project's conversations with users and replies to messages.

app/templates/project/conversations.hbs
```diff
@@ -0,0 +1,62 @@
+<div class="container">
+  <div class="conversations">
+    <section>
+      <aside>
```
joshsmith: I have yet to add the styling for `.conversations` but it will differ from the `.panel` styling. The left sidebar will be a scrolling list of messages and the right side will contain two panes: left for the message thread, right for the information about the target user or project. In total, then, the UI will appear to have three panes, but the right two panes are basically the `outlet` for the individual conversation.

joshsmith:
Also note that creating a message in the UI prompts you to confirm save of the record before redirecting, even though we're redirecting after saving the record.

joshsmith:
We should have a follow-up issue to change the `includes` serializer behavior introduced here whenever JSON API 1.1 is released and available in ember-data.

joshsmith:
I don't want to spend a bunch of time writing acceptance tests for logic that's going to change (in terms of new message creation), so I think we should instead:

- [ ] write a component (and test) for a new message modal on the `/project/people` page
- [ ] move the logic of `/conversations/new` to either `/project/people` or the new modal, depending on what makes sense
- [ ] remove the `/conversations/new`

app/routes/project/conversations/first.js
```diff
@@ -0,0 +1,11 @@
+import { get } from '@ember/object';
+import Route from '@ember/routing/route';
+
+export default Route.extend({
+  redirect() {
+    let { conversations } = this.modelFor('project.conversations');
+    if (get(conversations, 'length') > 0) {
+      this.transitionTo('project.conversations.conversation', get(conversations, 'firstObject'));
+    }
```
joshsmith: I don't have a better way to handle this. Could use thoughts @begedin. The downside here is not getting the `.active` class on the project menu at all, which we want. We could force it some other way, but it's not ideal.

joshsmith:
I think it's possible we're missing a `message.readAt`. I don't see any way to get the time the first message was read otherwise.

app/routes/project/conversations/first.js
joshsmith: Fixed by using `conversations.index` `redirect`.

joshsmith:
In response to my above comment, I think we might be able to infer whether the conversation was read based on whether the `conversation` has a `readAt`, but I'm not positive about this.

joshsmith:
A lot of changes above:

- redirects to `firstObject` (the first open conversation) from the `conversations.index` route
- sockets fully working
- adds avatars to conversation parts
- adds tooltips to those avatars
- adds conversation-thread component that has same automagical scrolling behavior on first load / posting a new message
- adds new message composer for the conversation

The only things not working about the project messages UI are:

- read status (can be implemented later)
- open/closed filtering
- opening/closing conversation

To sum up what's missing outside the scope of this specific PR, but which will need to be added before anything here is merged into `develop` (using this as the base branch, once things are deemed passing here):

- channel subscriptions to add new conversations into the sidebar list
- notification badge in the project menu for project messages
- user-facing messages UI
- channels subscriptions for the user
- notification badge in the header for user messages
- notifications in the favicon
- audible notifications
- "X is typing" channel events
- online presence for each user in the conversation (can be shown in the `/projects/people` UI, too)
- sending a message when joining a project
- sending a message when accepting/denying a contributor

app/services/conversation-channel.js
```diff
@@ -0,0 +1,38 @@
...
+    let socket = get(this, 'socket');
+    let id = get(conversation, 'id');
+    let channel = socket.channel(`conversation:${id}`);
+    channel.leave();
+  },
+
+  _onNewMessage(conversation) {
+    let store = get(this, 'store');
+    let id = get(conversation, 'id');
+    store.findRecord('conversation', id);
```
begedin: Could we not send the json payload for the conversation part with the event and then simply do `store.pushPayload`? 

This would save us two queries at least.

joshsmith: We cannot do that because the user who sent the message would have the payload pushed twice, and ember-data would rightly reject it.

There also may be data that has changed on the backend since the payload was sent, so best to simply update the parent model, IMO.

tests/integration/components/conversations/conversation-thread-test.js
```diff
@@ -1,25 +1,102 @@
...
+  renderPage();
+
+  assert.equal(page.conversationParts().count, 1, 'Only loaded parts are rendered.');
+});
+
+test('it delegates send action from composer', function(assert) {
+  assert.expect(1);
+
+  set(this, 'onSend', (body) => {
+    assert.equal(body, 'foo', 'Correct value was send with action.');
```
joshsmith: sent

joshsmith:
Wow, 3,333 lines of code exactly. 👏

PR #1598, opened 2017-12-11
rewrite-integration-tests-for-task-comment-list-component-to-use-page-objects

### LINKED ISSUE #760: Rewrite integration tests for task-comment-list component to use page objects
Falls under umbrella issue #697.

## Resources

- What's a page object? [[1](http://martinfowler.com/bliki/PageObject.html),[2](https://wyeworks.com/blog/2015/5/13/using-the-page-object-pattern-with-ember-cli/)]
- [Quickstart guide](http://ember-cli-page-object.js.org/docs/v1.6.x/quickstart)
- See [`tests/pages/components/donations/create-donation.js`](https://github.com/code-corps/code-corps-ember/blob/develop/tests/pages/components/donations/create-donation.js) and [`tests/integration/components/donations/create-donation-test.js`](https://github.com/code-corps/code-corps-ember/blob/develop/tests/pages/components/donations/create-donation.js) for in-house examples of, respectively, a page object and the page object being utilized in an integration test.
- To generate a page object, run `ember generate page-object component/<component-name>` in the project directory.

daveconnis:
# What's in this PR?

Rewriting `task-comment-list` to use page object in acceptance test.

Fixes #760

### DIFF
--- a/tests/acceptance/task-comments-test.js
+++ b/tests/acceptance/task-comments-test.js
@@ -22,7 +22,7 @@ test('Task comments are displayed correctly', function(assert) {
   });
 
   andThen(() => {
-    assert.equal(taskPage.comments().count, 4, 'The correct number of task comments is rendered');
+    assert.equal(taskPage.taskCommentList.comments().count, 4, 'The correct number of task comments is rendered');
   });
 });
 
@@ -50,7 +50,7 @@ test('A comment can be added to a task', function(assert) {
 
   andThen(() => {
     assert.equal(server.schema.comments.all().models.length, 1, 'A new comment was created');
-    assert.equal(taskPage.comments().count, 1, 'The comment is being rendered');
+    assert.equal(taskPage.taskCommentList.comments().count, 1, 'The comment is being rendered');
     let [comment] = server.schema.comments.all().models;
 
     assert.equal(comment.markdown, 'Test markdown', 'New comment has the correct markdown');
--- a/tests/pages/components/task-comment-list.js
+++ b/tests/pages/components/task-comment-list.js
@@ -1,0 +1,10 @@
+import {
+  collection
+} from 'ember-cli-page-object';
+
+export default {
+  comments: collection({
+    scope: '.task-comment-list',
+    itemScope: '.comment-item'
+  })
+};
--- a/tests/integration/components/task-comment-list-test.js
+++ b/tests/integration/components/task-comment-list-test.js
@@ -2,22 +2,36 @@ import RSVP from 'rsvp';
 import { moduleForComponent, test } from 'ember-qunit';
 import hbs from 'htmlbars-inline-precompile';
 import stubService from 'code-corps-ember/tests/helpers/stub-service';
+import pageObject from 'ember-cli-page-object';
+import taskCommentList from 'code-corps-ember/tests/pages/components/task-comment-list';
+
+let page = pageObject.create(taskCommentList);
+
+function renderPage() {
+  page.render(hbs`
+  {{task-comment-list comments=comments}}`
+  );
+}
 
 moduleForComponent('task-comment-list', 'Integration | Component | task comment list', {
   integration: true,
   beforeEach() {
+    page.setContext(this);
     stubService(this, 'store', {
       query() {
         return RSVP.resolve([]);
       }
     });
+  },
+  afterEach() {
+    page.removeContext();
   }
 });
 
 test('it renders', function(assert) {
   assert.expect(1);
-  this.render(hbs`{{task-comment-list}}`);
-  assert.equal(this.$('.task-comment-list').length, 1, 'The component\'s element renders');
+  renderPage();
+  assert.ok(page.isVisible, 'The component\'s element renders');
 });
 
 test('It renders a list of comments if there are comments', function(assert) {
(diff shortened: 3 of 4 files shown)


tests/acceptance/task-comments-test.js
```diff
@@ -3,6 +3,16 @@ import moduleForAcceptance from 'code-corps-ember/tests/helpers/module-for-accep
...
 import taskPage from '../pages/project/tasks/task';
+import PageObject from 'ember-cli-page-object';
+import hbs from 'htmlbars-inline-precompile';
+import taskCommentList from 'code-corps-ember/tests/pages/component/task-comment-list';
+
+let page = PageObject.create(taskCommentList);
+
+function renderPage() {
+  page.render(hbs`
+    {{comment-item comment=comment saveError=onSaveError}}`);
```
joshsmith: As a heads up, I would move the `);` down to the next line generally.

tests/pages/component/task-comment-list.js
```diff
@@ -0,0 +1,32 @@
+import {
+  collection,
+  clickable
+} from 'ember-cli-page-object';
+import taskPage from 'code-corps-ember/tests/pages/project/tasks/task';
+import createCommentForm from 'code-corps-ember/tests/pages/components/create-comment-form';
+export default ({
+
+  ClickSave: clickable('.save'),
```
joshsmith: `clickSave`

tests/pages/component/task-comment-list.js
```diff
@@ -0,0 +1,32 @@
...
+    itemScope: '.comment-item'
+  }),
+
+  commentBody: {
+    scope: '.comment-body p'
+  },
+
+  createCommentForm,
+
+  taskPage,
```
joshsmith: The `taskPage` is the parent here, so you might want to rethink the ordering of your page objects here.

tests/acceptance/task-comments-test.js
```diff
@@ -15,14 +25,15 @@ test('Task comments are displayed correctly', function(assert) {
 
   server.createList('comment', 4, { taskId: task.id });
 
-  taskPage.visit({
+  renderPage();
+  page.visit({
```
joshsmith: Since this is an acceptance test, you still want to test the `taskPage` by doing `.visit()`. Your `renderPage` is superfluous here because it's a) not an integration test, and b) we're testing the comments in the context of the task page (which is the parent, see below).

tests/acceptance/task-comments-test.js
```diff
@@ -15,14 +25,15 @@ test('Task comments are displayed correctly', function(assert) {
...
     organization: organization.slug,
     project: project.slug,
     number: task.number
   });
 
   andThen(() => {
-    assert.equal(taskPage.comments().count, 4, 'The correct number of task comments is rendered');
```
daveconnis: I'm getting an error that says "died on test#1" then shows `Error: Assertion Failed: ":first" matched more than one element. If this is not an error use { multiple: true }` 

This is my first assert and I'm not seeing `first:` in the code. Only thing I can think of is that somehow it's trying to return the first comment in the list but is catching something else with it? In the past, when I've seen this error, it's because I was re-rendering the objects through a `export default create` on the page object. So, I'm assuming that I've got code somewhere that's returning multiple objects, but I'm not sure how and what because I can't even track down where `:first` is defined.

daveconnis:
Done for the day, worked on this all day, lots of slow debugging that worked until it didn't. First bug was that I wasn't importing my page objects correctly into my new `task-comment-list`.  Took me a little bit to figure out what was going on, but I got it. Wrote out what I thought would be the hierarchy/specs for what I was doing but then I got stuck because I was using renderPage() on an acceptance test, which is probably where the error I outlined in the code is originating.

I have tests passing now! I'm still wondering if/how I need to include the new page object into task-card-test. Will consider in the morning. Otherwise, if it looks good, I will start on something else in the morning

tests/acceptance/task-comments-test.js
begedin: @daveconnis 

As I spoke about during our call, integration tests, instead of booting up the whole app just do a `page.render`, to render the component they're testing.

Acceptance tests, however, are end to end and do boot-up the whole app. This is done in the `moduleForAcceptance` function.

That means that, instead of rendering that one component, since the app is booted up, you navigate to the page/route in the app which you are testing, using `page.visit`

tests/acceptance/task-comments-test.js
begedin: To explain why you're not seeing `:first` in the code.

If you look at the docs for `collection()`, there are some examples there:

http://ember-cli-page-object.js.org/docs/v1.12.x/api/collection

If you define a collection as 

```javascript

items: collection({
  scope: 'foo',
  itemScope: 'bar'
})
```

What this means is, you're saying

- i have a bunch of elements of the same type on my page
- they are all contained within an element with the class of 'foo'
- each individual one of those items is carrying the class 'bar'

Since you said that, 
- `page.items().count` internally calls `$('.foo .bar').length`
- `page.items(1)` internally calls `$('.foo .bar:eq(1)')`

Now, all that's fine now that you fixed the page object, but your old one had some weird error, which I can't figure out myself, which caused the `ember-cli-page-object` internals to fallback on their default selector - `:first`, which then also somehow returned multiple objects, which then, by default, unless you manually set it up otherwise, will error out.

So you might not have seen `:first` yourself, but it was in there, same as how 
'.foo .bar:eq(1)' is in there, even though you're not the one writing it.

tests/pages/components/task-comment-list.js
```diff
@@ -0,0 +1,11 @@
+import {
+  collection
+} from 'ember-cli-page-object';
+export default ({
+
+  comments: collection({
+    scope: '.task-comment-list',
+    itemScope: '.comment-item'
+  })
+
```
joshsmith: Extra space here.

tests/pages/components/task-comment-list.js
```diff
@@ -0,0 +1,11 @@
+import {
+  collection
+} from 'ember-cli-page-object';
+export default ({
```
joshsmith: You should add a line between this `export`, and remove the parens wrapping it. Much prefer the `export default {` style.

joshsmith:
🙌

PR #1599, opened 2017-12-12
Add Mixin for confirm dialogs on new routes

### LINKED ISSUE #1569: Create a mixin for confirmation dialogs when working with long forms
# Problem

We have a number of places where we ask a user to confirm before leaving the page. We could write a mixin to help clean up this logic.

```js
export default Ember.Mixin.extend({;
  modelName: null,
  message: 'You will lose any unsaved information if you leave this page. Are you sure?',
  
  actions: {
    willTransition() {
      let modelName = get(this, 'modelName');
      if (isPresent(modelName)) {
         let model = get(this, modelName);
         if (get(modelName, 'isNew') {
           this._confirmTransition(transition, model);
         } 
      } else {
        this._confirmTransition(transition);
      }
    }
  },

  _confirmTransition(transition, model) {
    let message = get(this, 'message');
    if (window.confirm(message)) {
      if isPresent(model) {
        model.destroyRecord();
      }
    } else {
      transition.abort();
    }
  }
})
```

snewcomer:
## References
Fixes #1569


app/controllers/admin/organization-invites/new.js
```diff
@@ -7,6 +7,7 @@ export default Controller.extend({
   flashMessages: service(),
   store: service(),
 
+  modelName: 'organizationInvite',
```
snewcomer: Not sure I'm a huge fan of this pattern.  The mixin needs to determine a dynamic model on the route and we don't have a common identifier such as `model`.  Have any thoughts?

begedin: Not sure if there's another way to go about it.

To me, calling the main model just `model` isn't great either. It lacks explicitness. 

We expected to require this. I just wish there was a way to extend a mixin while explicitly providing an option for it, kind of like elixir's `use X, option: something`. This line lacks explicitness to.

I'm also not sure I like the fact that we have to use the mixin in a route and then specify a model name in the controller.That part, we can change, though, with the way the mixin works, so could you move the model name into route?

@joshsmith What do you think about the other stuff?

Maybe naming can help?

We could call the mixin `RollbackModelOnNavigatingAway`. It's verbose, but also provides clarity for itself, as well as for the `modelName` property we could use on the route.

We could also make the modelName into an alias, as well as call it `rollbackModel` to add further clarity. In this specific case:

```javascript
export default Route.extend(RollbackModelOnNavigatingAway, {
  rollbackModel: alias('controller.organizationInvite')
});
```

snewcomer: @begedin I moved the name to the route.  Lmk what you still think about the naming...I can change it if you would like!

app/mixins/confirm.js
```diff
@@ -0,0 +1,33 @@
...
+  actions: {
+    willTransition(transition) {
+      let modelName = get(this, 'modelName');
+      if (isPresent(modelName)) {
+        let model = get(this, `controller.${modelName}`);
+        if (get(model, 'isNew')) {
+          this._confirmTransition(transition, model);
+        }
+      } else {
+        this._confirmTransition(transition);
```
joshsmith: Sorry I'm just now saying this, but it looks like this mixin has an edge case that's not covered here: https://coveralls.io/builds/14663763/source?filename=app%2Fmixins%2Fconfirm.js

snewcomer:
@joshsmith I think this PR is g2g.

PR #1625, opened 2017-12-18
Update busy model wrapper component to use page object

### LINKED ISSUE #1607: Rewrite integration tests for busy-model-wrapper-test component
# Problem

As of right now, `tests/integration/components/common/busy-model-wrapper-test.js` isn't utilizing ember-cli-page-objects, and instead is using jQuery to select elements.

## References
#1606 

## Resources

- What's a page object? [[1](http://martinfowler.com/bliki/PageObject.html),[2](https://wyeworks.com/blog/2015/5/13/using-the-page-object-pattern-with-ember-cli/)]

- [Quickstart guide](http://ember-cli-page-object.js.org/docs/v1.6.x/quickstart)

- See tests/pages/components/donations/create-donation.js and tests/integration/components/donations/create-donation-test.js for in-house examples of, respectively, a page object and the page object being utilized in an integration test.

- To generate a page object, run `ember generate page-object component/<component-name>` in the project directory.

daveconnis:
# What's in this PR?

Adding page object to `busy-model-wrapper-test` and writing out tests that actually test the component. Originally, there was only an `it renders` test. Planning on testing both `onSaving` and `onDeleting`.

Fixes #1607

### DIFF
--- a/tests/integration/components/common/busy-model-wrapper-test.js
+++ b/tests/integration/components/common/busy-model-wrapper-test.js
@@ -1,25 +1,56 @@
 import { moduleForComponent, test } from 'ember-qunit';
 import hbs from 'htmlbars-inline-precompile';
+import { set } from '@ember/object';
+import PageObject from 'ember-cli-page-object';
+
+// Since this is a tagless component, we cannot effectively use a typical page
+// object.
+//
+// Instead, we render the tagless form inside a test container and dynamically
+// define a page object using the scope of the test container.
+
+let page = PageObject.create({ scope: '.test-container' });
+
+function renderPage() {
+  page.render(hbs`
+    <div class="test-container">
+      {{#common/busy-model-wrapper model=model}}
+        Yielded
+      {{/common/busy-model-wrapper}}
+    </div>
+  `);
+}
 
 moduleForComponent('common/busy-model-wrapper', 'Integration | Component | common/busy model wrapper', {
-  integration: true
+  integration: true,
+  beforeEach() {
+    page.setContext(this);
+  },
+  afterEach() {
+    page.removeContext();
+  }
 });
 
-test('it renders', function(assert) {
+test('it shows "Deleting" if model is saving and is deleted', function(assert) {
+  let model = { isSaving: true, isDeleted: true };
+  set(this, 'model', model);
 
-  // Set any properties with this.set('myProperty', 'value');
-  // Handle any actions with this.on('myAction', function(val) { ... });
+  renderPage();
+  assert.equal(page.text, 'Deleting...', 'A deletion indicator is rendered in place of the yielded block.');
+});
 
-  this.render(hbs`{{common/busy-model-wrapper}}`);
+test('it shows "Saving" if model is saving and not deleted', function(assert) {
+  let model = { isSaving: true, isDeleted: false };
+  set(this, 'model', model);
 
-  assert.equal(this.$().text().trim(), '');
+  renderPage();
+  assert.equal(page.text, 'Saving...', 'A deletion indicator is rendered in place of the yielded block.');
+});
 
-  // Template block usage:
-  this.render(hbs`
-    {{#common/busy-model-wrapper}}
-      template block text
-    {{/common/busy-model-wrapper}}
-  `);
+test('it shows yield if model is neither saving or deleted', function(assert) {
+  let model = { isSaving: false, isDeleted: false };
+  set(this, 'model', model);
 
-  assert.equal(this.$().text().trim(), 'template block text');
+  renderPage();
+  assert.equal(page.text, 'Yielded', 'The yielded block is rendered.');
 });
--- a/app/components/common/busy-model-wrapper.js
+++ b/app/components/common/busy-model-wrapper.js
@@ -1,9 +1,8 @@
 import Component from '@ember/component';
 
 export default Component.extend({
-  tagName:'',
   model: null,
 
-  onSaving: 'Saving...',
-  onDeleting: 'Deleting...'
+  onDeleting: 'Deleting...',
+  onSaving: 'Saving...'
 });


daveconnis:
Unsure what I need in the page object because I'm not sure what `model` is in `/Users/imaginationstation/sites/code-corps-ember/app/templates/components/common/busy-model-wrapper.hbs`. What is this referencing? It isn't the model which is set to `null` in the corresponding .js file is it? if so, where is the action defined for setting model to `model.isSaving` or `model.isDeleting`

begedin:
@daveconnis The idea of that component, is, you give it an ember data model and wrap the component around some sort of UI related to that model. If the model is not being saved or deleted at the moment, it shows that UI. Otherwise, it shows a message indicating it's saving or deleting.

From that, you don't really need to put anything in the page object, although we may at some point

What you want to test is 

- if model is saving, but is not deleted, page.text is "Saving"
- if model is saving and is deleted, page.text is "Deleting"
- if model is neither saving nor deleted, page.text is whatever you wrapped the component around

If you set your render page as

```Javascript
page.render(hbs`{{#budy-model-wrapper model=model}}Foo{{/busy-model-wrapper}}`)
```

Then bullet nr. 3 simply asserts  `page.text` is `'Foo'`

begedin:
So to add to my previous comment, your direction is fine. Just need to cover the 2 other cases.

daveconnis:
Have the first two tests passing, by setting the model, however, I'm unsure how to get the third passing without something to yield. Nikola mentioned earlier that I could set my renderPage to `page.render(hbs`{{#busy-model-wrapper model=model}}Foo{{/busy-model-wrapper}}`)` but I've tried this both in my initial `renderPage()` function and in the test itself and haven't gotten anywhere with it. The error I get is 

```Error: Element not found.

PageObject: 'page.text'
  Selector: ':first'```

I got this on both of my tests prior, but only because I'd messed up the boolean values onSaving and onDeleting. I fixed that, but I'm not sure what this would need to accept that neither onSaving or onDeleting are false.

File: tests/integration/components/common/busy-model-wrapper-test.js [hunk 1]
```diff
@@ -22,15 +23,26 @@ moduleForComponent('common/busy-model-wrapper', 'Integration | Component | commo
...
+test('it shows "Deleting" if model is saving and is deleted', function(assert) {
+  let model = { isSaving: true, isDeleted: true };
+  set(this, 'model', model);
   renderPage();
 
-  assert.ok(page.isVisible);
+  assert.equal(page.text, 'Deleting...', 'the model is deleting');
 });
 
-test('it shows "saving" if model is saving', function(assert) {
```
daveconnis: Have the first two tests passing, by setting the model, however, I'm unsure how to get the this one passing without something to yield. Nikola mentioned earlier that I could set my renderPage to page.render(hbs{{#busy-model-wrapper model=model}}Foo{{/busy-model-wrapper}}) but I've tried this both in my initial renderPage() function and in the test itself and haven't gotten anywhere with it. The error I get is

Error: Element not found.

PageObject: 'page.text'
  Selector: ':first' 

I got this on both of my tests prior, but only because I'd messed up the boolean values onSaving and onDeleting. I fixed that, but I'm not sure what this would need to accept that neither onSaving or onDeleting are false.

joshsmith:
You won't want to use `renderPage` for this since that's just a helper function.

In this case you would do:

```
page.render(hbs`
  {{#busy-model-wrapper model=model}}
    Foo
  {{/busy-model-wrapper}}
`);
```

inside of your test.

tests/integration/components/common/busy-model-wrapper-test.js
```diff
@@ -1,25 +1,48 @@
...
-    {{#common/busy-model-wrapper}}
-      template block text
-    {{/common/busy-model-wrapper}}
-  `);
+test('it shows yeild if model is neither saving or deleted', function(assert) {
+  let model = { isSaving: false, isDeleted: false };
+  set(this, 'model', model);
 
-  assert.equal(this.$().text().trim(), 'template block text');
+  assert.equal(page.text, 'Foo', 'the model returns block');
```
joshsmith: Is this where you were having trouble with the yield block?

tests/integration/components/common/busy-model-wrapper-test.js
```diff
@@ -1,25 +1,48 @@
...
 });
```
daveconnis: Have the first two tests passing, by setting the model, however, I'm unsure how to get the this one passing without something to yield. Nikola mentioned earlier that I could set my renderPage to page.render(hbs{{#busy-model-wrapper model=model}}Foo{{/busy-model-wrapper}}) but I've tried this both in my initial renderPage() function and in the test itself and haven't gotten anywhere with it. The error I get is

Error: Element not found.

PageObject: 'page.text'
Selector: ':first'

I got this on both of my tests prior, but only because I'd messed up the boolean values onSaving and onDeleting. I fixed that, but I'm not sure what this would need to accept that neither onSaving or onDeleting are false.

joshsmith: You won't want to use `renderPage` for this since that's just a helper function.

In this case you would do:

```js
page.render(hbs`
  {{#common/busy-model-wrapper model=model}}
    Foo
  {{/common/busy-model-wrapper}}
`);
```

inside of your test.

daveconnis:
I've deleted the tag, class, and page object for this, haven't switched the page object syntax back to jQuery yet, but when I do, I'll be sure to comment out WHY we haven't switched this over to a page object.

joshsmith:
@daveconnis do you want to finish this one up before continuing on the other?

tests/integration/components/common/busy-model-wrapper-test.js
```diff
@@ -1,53 +1,46 @@
...
-    page.removeContext();
-  }
+  integration: true
 });
 
-test('it shows "saving" if model is saving, but not deleted', function(assert) {
-  let model = { isSaving: true, isDeleted: false };
-  set(this, 'model', model);
-  renderPage();
+// These tests do not make use of a page object. One was written, but it consisted of an empty string and a .busymodelwrapper scope. Set up this way, the page object wasn't actually rendering anything, just empty code placeholders. Switched to using jQuery selectors in order to have cleaner more effectived code.
```
daveconnis: Biggest thing I want attention on is this comment. Want to make sure it makes sense. Remember, we veto'd the page object for this test because it was just, if I remember right:
`tagname: ' ' 
 scope: .busymodelwrapper`

tests/integration/components/common/busy-model-wrapper-test.js
```diff
@@ -1,25 +1,46 @@
 import { moduleForComponent, test } from 'ember-qunit';
 import hbs from 'htmlbars-inline-precompile';
+import { set } from '@ember/object';
 
 moduleForComponent('common/busy-model-wrapper', 'Integration | Component | common/busy model wrapper', {
   integration: true
 });
 
-test('it renders', function(assert) {
+// These tests do not make use of a page object. One was written, but it consisted of an empty string and a .busymodelwrapper scope. Set up this way, the page object wasn't actually rendering anything, just empty code placeholders. Switched to using jQuery selectors in order to have cleaner more effectived code.
```
joshsmith: Comments should be multi-line when over 80 chars.

begedin:
@daveconnis I decided to wrap this up.

In hindsight, we didn't pay enough attention on how the busy model wrapper works and sent you in the wrong direction.

What I ended up doing is I used your findings to setup a dynamically defined page object and a render-page function which wraps the component call in a test container div, so we can still assert `page.text` instead of `this.$().text().trim()`

PR #1626, opened 2017-12-19
Add user conversations

joshsmith:
# What's in this PR?

Adds user conversations.

Still needed:

- [x] acceptance tests for user conversations
- [x] acceptance tests for project conversations (optional, but straightforward once we write the cousin tests)


begedin:
Added what I think are the missing tests.

My vote is against drying up at this point. Those code points have a higher chance of diverging than converging and I don't think dryer code here is worth the potential added complexity.

begedin:
@joshsmith I added a couple more things, since it was easier to do here than extract into separate PRs

- There are 2 more acceptance tests, to ensure routes join the conv. channel and are listening for changes
- There are a couple of route unit tests, really basic, but better than nothing
- There is the close action for the conversation, with required tests here as well. This needs an API PR to enable the update endpoint on conversations. I've used mirage to implement it.

tests/acceptance/project-conversations-test.js
```diff
@@ -83,6 +84,41 @@ test('Project admin can view single conversations', function(assert) {
...
+    assert.equal(page.conversationThread.conversationParts().count, 1, 'Just the message head is rendered.');
+    server.create('conversation-part', { conversation });
+  });
+
+  andThen(() => {
+    assert.equal(page.conversationThread.conversationParts().count, 1, 'No notification yet, so new part was not rendered.');
+    let conversationChannelService = this.application.__container__.lookup('service:conversation-channel');
+    let socket = get(conversationChannelService, 'socket');
+    let [channel] = socket.channels;
+    channel.trigger('new:conversation-part', {});
```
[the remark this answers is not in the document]
joshsmith: I think actually that we're not super up-to-date on the changes in the test helpers.

tests/integration/components/conversations/conversation-list-item-with-user-test.js
```diff
@@ -48,3 +50,45 @@ test('it renders the conversation details', function(assert) {
...
+    user
+  };
+
+  set(this, 'conversation', conversation);
+
+  renderPage();
+
+  assert.notOk(page.closeButton.isVisible, 'Close button is hidden');
+});
+test('it calls bound action when clicking close button', function(assert) {
```
joshsmith: Add line break.

app/controllers/project/conversations.js
```diff
@@ -1,4 +1,11 @@
 import Controller from '@ember/controller';
+import { set } from '@ember/object';
 
 export default Controller.extend({
+  actions: {
+    close(conversation) {
+      set(conversation, 'status', 'closed');
+      conversation.save();
```
joshsmith: I might consider adding a `return` here in the event we ever `pipe` this, so the promise returns/rejects.

PR #1627, opened 2017-12-19
Update socket connection based on session state

begedin:
# What's in this PR?

This binds the conversation channel service to session `authenticationSucceeded` and `invalidationSucceeded` events. 

The former is handled by calling `connect`, while the later is handled by calling disconnect.

I could not find an effective way to test for this, however.

## References

Progress on: #1597

### DIFF
--- a/app/services/conversation-channel.js
+++ b/app/services/conversation-channel.js
@@ -1,39 +1,29 @@
-import PhoenixSocket from 'phoenix/services/phoenix-socket';
 import { get } from '@ember/object';
-import { inject as service } from '@ember/service';
-import ENV from 'code-corps-ember/config/environment';
+import Service, { inject as service } from '@ember/service';
 
-export default PhoenixSocket.extend({
-  session: service(),
+export default Service.extend({
+  socket: service(),
   store: service(),
 
-  connect() {
-    if (!get(this, 'socket')) { // Only create one socket at a time
-      let token = get(this, 'session.session.authenticated.token');
-      this._super(ENV.SOCKETS_BASE_URL, {
-        params: { token }
-      });
-    }
-  },
-
   async join(conversation) {
-    await this.connect();
+    let socket = get(this, 'socket');
+
+    await socket.connect();
+
     let id = get(conversation, 'id');
-    let channel = this.joinChannel(`conversation:${id}`);
+    let channel = socket.joinChannel(`conversation:${id}`);
 
     channel.on('new:conversation-part', () => this._onNewMessage(conversation));
   },
 
   leave(conversation) {
-    let socket = get(this, 'socket');
     let id = get(conversation, 'id');
-    let channel = socket.channel(`conversation:${id}`);
-    channel.leave();
+    return get(this, 'socket').leaveChannel(`conversation:${id}`);
   },
 
   _onNewMessage(conversation) {
     let store = get(this, 'store');
     let id = get(conversation, 'id');
-    store.findRecord('conversation', id);
+    return store.findRecord('conversation', id);
   }
 });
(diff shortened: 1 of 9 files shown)


app/services/conversation-channel.js
```diff
@@ -7,6 +7,11 @@ export default PhoenixSocket.extend({
   session: service(),
   store: service(),
 
+  init() {
+    get(this, 'session').on('authenticationSucceeded', () => this.connect());
```
snewcomer: @begedin is this service instantiated when a user enters the ember app and before the authenticationSucceeded event is triggered?

snewcomer: Ah nvmd.  This is just passively listening.  `join` will call connect at the right time.

app/services/socket.js
```diff
@@ -0,0 +1,43 @@
...
+  init() {
+    let session = get(this, 'session');
+    session.on('authenticationSucceeded', () => this.toggleConnection());
+    session.on('invalidationSucceeded', () => this.toggleConnection());
+  },
+
+  connect() {
+    if (!get(this, 'socket')) { // Only create one socket at a time
+      let token = get(this, 'session.session.authenticated.token');
+      this._super(ENV.SOCKETS_BASE_URL, {
```
snewcomer: Where is this environmental variable defined?  @joshsmith

joshsmith: In the config/environment.js

snewcomer: Where?  I don't see it on develop branch

joshsmith: The base branch is `1592-project-messages`.

app/services/socket.js
```diff
@@ -0,0 +1,52 @@
...
+    let token = get(this, 'session.session.authenticated.token');
+    let params = token ? { token } : {};
+
+    // Only create a socket when:
+    // - no socket exists
+    // - we're exchanging an unauthenticated socket for an authenticated one
+    if (!socket) {
+      this._super(ENV.SOCKETS_BASE_URL, { params });
+    } else if (socket && token) {
+      socket.disconnect(() => this._super(ENV.SOCKETS_BASE_URL, { params }));
```
snewcomer: Is this the same `disconnect` method a few lines below?

app/services/socket.js
```diff
@@ -0,0 +1,52 @@
...
+  disconnect(callback) {
+    let socket = get(this, 'socket');
+    if (socket) {
+      socket.disconnect(callback);
+    } else {
+      return callback && callback();
+    }
+  },
+
+  joinChannel(channelName) {
```
snewcomer: I wonder if some doc blocks for a few of these methods might be helpful...

app/services/conversation-channel.js
```diff
@@ -1,39 +1,29 @@
...
-        params: { token }
-      });
-    }
-  },
-
   async join(conversation) {
-    await this.connect();
+    let socket = get(this, 'socket');
+
+    await socket.connect();
```
snewcomer: Any failure state we need to be aware of?

app/services/socket.js
joshsmith: These are really just wrappers around `connect` and `disconnect` for `PhoenixSocket` from `ember-phoenix`.

app/services/conversation-channel.js
joshsmith: There are some defined in Phoenix's JS but none I think we _need_ to handle here right now.

joshsmith:
@begedin fyi I had to amend this to account for the fact that when the application loads and a session is restored, `authenticationSucceeded` does not fire. We have to use the application route's `model` hook _in addition to_ observing state, and "upgrade" a socket to its authenticated form if a token is present.

I spent way too much time on this, but it now works in a verifiable way.

PR #1643, opened 2017-12-21
Sort conversations by updatedAt timestamp

### LINKED ISSUE #1634: Sort conversations by updated_at timestamp
# Problem

Right now conversations are sorted by `id`. We should change this to sort by their `updated_at` timestamp instead.

This should happen in both user and project conversations.

joshsmith:
# What's in this PR?

No tests yet. I would love to come up with a way to test this that does not involve us writing an acceptance test.

I think perhaps we should consider introducing a component here as a result.

Thoughts?

## References
Fixes #1634

### DIFF
--- a/tests/acceptance/project-conversations-test.js
+++ b/tests/acceptance/project-conversations-test.js
@@ -4,7 +4,7 @@ import moduleForAcceptance from 'code-corps-ember/tests/helpers/module-for-accep
 import { authenticateSession } from 'code-corps-ember/tests/helpers/ember-simple-auth';
 import page from 'code-corps-ember/tests/pages/project/conversations';
 
-const hourOfMiliseconds = 3600 * 1000;
+const hour = 3600 * 1000;
 
 function createConversations(count, project, user) {
   return [...Array(count)].map(() => {
@@ -56,8 +56,8 @@ test('Project admin can view list of conversations', function(assert) {
 
   let [date1, date2, date3] = [
     Date.now(),
-    Date.now() - 5 * hourOfMiliseconds,
-    Date.now() - 1 * hourOfMiliseconds
+    Date.now() - 5 * hour,
+    Date.now() - 1 * hour
   ];
 
   server.create('conversation', { message:  message1, user, updatedAt: date1 });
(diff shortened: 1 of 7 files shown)


begedin:
@joshsmith I tried to make it work with a component.

However, due to the way the two routes are rendered, we would have the following choises

- add a single, extremely coupled component with a decent amount of `#if` switching
  - it needs to know which route it's rendered on in order to show the `blank` fallback
  - it needs to use a link-to helper for the project route
  - it needs to know which route it's on in order to know which item component to render
  - advantage: we can integration test
  - disadvantage: lots of logic to test, to the point where a single acceptance test might not be that much slower
  - disadvantage: lots of dependencies we have to mock

- add a single, extremely decoupled component
  - it uses the template block approach
  - it yields to a given block with 2 arguments - type of yield ('empty' or 'item') and the item itself
  - the given block then checks the first argument and renders either the proper item component, or the fallback for no items
  - advantage: almost no coupling, extremely easy and quick to test the component itself
  - disadvantage: not that much of the logic has been moved away from the route
  - disadvantage: the component is effectively a glorified `sort-by` helper

- some sort of mix between the two
  - difficult to do because, as I said, a different item component is rendered based on the route, different actions are bound to the item component, different fallback in case of a blank list. We can't even use the `component` helper due to different actions

Due to all that, I decided to go with a third option.

I did not write a new acceptance test.

Each of our acceptance test modules __already has a _"user can view a list of conversations"_ test___.

I simply expanded those tests by creating records with different timestamps and making sure those timestamps are rendered in the correct order.

As an added bonus, __we also check the timestamps are properly formatted__ with moment.

Since this just expands two existing tests, it increases our test time far less than a new integration test module would.

begedin:
Sorry, was supposed to be removed before commit. Will clean up

tests/acceptance/project-conversations-test.js
```diff
@@ -4,6 +4,8 @@ import moduleForAcceptance from 'code-corps-ember/tests/helpers/module-for-accep
 import { authenticateSession } from 'code-corps-ember/tests/helpers/ember-simple-auth';
 import page from 'code-corps-ember/tests/pages/project/conversations';
 
+const hourOfMiliseconds = 3600 * 1000;
```
joshsmith: Can you just write this as hour? Means effectively the same thing and is easier to read quickly.

Issue #1632, opened 2017-12-20
Add filtering conversations by open/closed to the project conversation list


joshsmith:
# Problem

The conversation list for projects needs to have an `open/closed` filter placed on it.

We should show a (styled) `select` that has two items:

- Open (3)
- Closed (18)

Where the number in parens is the number of currently open or closed conversations.

This is blocked by implementing a counter cache for open and closed conversations on the API.

PR #1647, opened 2017-12-21
Add conversation status selector

joshsmith:
# What's in this PR?

Add a selector to change the conversation status.

Needs tests.

## References
Progress on: #1632

### DIFF
--- a/app/controllers/project/conversations.js
+++ b/app/controllers/project/conversations.js
@@ -2,6 +2,11 @@ import Controller from '@ember/controller';
 import { set } from '@ember/object';
 
 export default Controller.extend({
+  queryParams: ['status'],
+  status: 'open',
+
+  currentlyLoading: null,
+
   actions: {
     close(conversation) {
       set(conversation, 'status', 'closed');
--- a/app/routes/project/conversations.js
+++ b/app/routes/project/conversations.js
@@ -21,15 +21,21 @@ export default Route.extend(AuthenticatedRouteMixin, CanMixin, {
     }
   },
 
-  async model() {
+  async model(params) {
     let project = this.modelFor('project');
     let projectId = get(project, 'id');
     let store = get(this, 'store');
-    // peekAll returns a live array, so it will alse render any newly created
-    // messages, causing the list to update properly on project.messages.new
-    await store.query('conversation', { project_id: projectId });
-    let conversations = store.peekAll('conversation');
+    let conversations = await store.query('conversation', {
+      project_id: projectId,
+      status: params.status
+    });
     return { conversations, project };
+  },
+
+  queryParams: {
+    status: {
+      refreshModel: true
+    }
   },
 
   renderTemplate() {
@@ -64,6 +70,15 @@ export default Route.extend(AuthenticatedRouteMixin, CanMixin, {
       this._super(...arguments);
       get(this, 'conversations').activate();
       return true;
+    },
+
+    loading(transition) {
+      let controller = this.controllerFor('project.conversations');
+      controller.set('currentlyLoading', true);
+      transition.promise.finally(() => {
+        controller.set('currentlyLoading', false);
+        this.transitionTo('project.conversations.index');
+      });
     }
   }
 });
--- a/app/routes/project/conversations/index.js
+++ b/app/routes/project/conversations/index.js
@@ -3,10 +3,17 @@ import Route from '@ember/routing/route';
 import AuthenticatedRouteMixin from 'ember-simple-auth/mixins/authenticated-route-mixin';
 
 export default Route.extend(AuthenticatedRouteMixin, {
-  redirect() {
+  redirect(model, transition) {
     let { conversations } = this.modelFor('project.conversations');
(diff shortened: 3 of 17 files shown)


app/routes/project/conversations.js
```diff
@@ -64,6 +70,16 @@ export default Route.extend(AuthenticatedRouteMixin, CanMixin, {
       this._super(...arguments);
       get(this, 'conversations').activate();
       return true;
+    },
+
+    loading(transition) {
+      let controller = this.controllerFor('project.conversations');
+      controller.set('currentlyLoading', true);
+      let that = this;
+      transition.promise.finally(function() {
```
snewcomer: can a hash rocket be used here?

tests/integration/components/conversations/status-select-test.js
```diff
@@ -0,0 +1,52 @@
...
+    page.removeContext();
+  }
+});
+
+test('it opens, closes, and renders the correct states', async function(assert) {
+  assert.expect(8);
+
+  set(this, 'status', 'open');
+
+  renderPage();
```
snewcomer: does `renderPage` need an away?  I'm guessing not since this passes, but have seen `await` with render in the new qunit api

tests/acceptance/project-conversations-test.js
```diff
@@ -214,11 +212,28 @@ test('Project admin can reopen a conversation', function(assert) {
     project: project.slug
   });
 
+  andThen(() => {
+    page.statusSelect.openButton.click();
```
snewcomer: The new async await would be rid of the andThens'.  I'm sure you already know that but I found it was useful to start moving acceptance tests over 1 test file at a time (maybe a codemod) available now???

app/controllers/project/conversations.js
```diff
@@ -2,6 +2,11 @@ import Controller from '@ember/controller';
 import { set } from '@ember/object';
 
 export default Controller.extend({
+  queryParams: ['status'],
+  status: 'open',
+
+  currentlyLoading: null,
```
snewcomer: do you want the default to be false?

app/routes/project/conversations.js
```diff
@@ -64,6 +70,16 @@ export default Route.extend(AuthenticatedRouteMixin, CanMixin, {
...
+        controller.set('currentlyLoading', false);
+        that.transitionTo('project.conversations.index');
+      });
```
begedin: using an arrow function should eliminate the need for `that`

app/routes/project/conversations/index.js
```diff
@@ -3,10 +3,14 @@ import Route from '@ember/routing/route';
...
 
 export default Route.extend(AuthenticatedRouteMixin, {
-  redirect() {
+  redirect(model, transition) {
+    let status = transition.queryParams.status || 'open';
     let { conversations } = this.modelFor('project.conversations');
-    if (get(conversations, 'length') > 0) {
-      this.transitionTo('project.conversations.conversation', get(conversations, 'firstObject'));
+    let sortedConversations
+      = conversations.filterBy('status', status).sortBy('updatedAt');
```
begedin: I don't think you need to filter here, right? The model should already be filtered. If it's not, then we should add a comment as to why that is.

You do need to sort, but not until you know the array isn't empty. That shouldn't matter much, though. Sorting a blank array is not expensive.

tests/integration/components/conversations/status-select-test.js
begedin: It might need it of we're setting a component property without wrapping it in a `run`. Someone ought to try that out.

app/controllers/project/conversations.js
joshsmith: I don't think so, since it could result in some weird state.

app/routes/project/conversations/index.js
joshsmith: The filtering does need to happen since the model does not necessarily update when an individual conversation changes.

tests/acceptance/project-conversations-test.js
```diff
@@ -64,112 +62,96 @@ test('Project admin can view list of conversations', function(assert) {
...
-  page.visit({
+  await page.visit({
     organization: project.organization.slug,
     project: project.slug
   });
 
-  andThen(() => {
-    page.conversations(1).click();
-  });
+  page.conversations(1).click();
```
snewcomer: `await` here as well?  I thik it may b/c `click` is an async test helper.

https://github.com/san650/ember-cli-page-object/blob/master/addon/-private/properties/clickable.js#L68

tests/acceptance/project-conversations-test.js
```diff
@@ -64,112 +62,96 @@ test('Project admin can view list of conversations', function(assert) {
...
-  page.visit({
+  await page.visit({
     organization: project.organization.slug,
     project: project.slug
   });
 
-  andThen(() => {
-    page.conversations(0).click();
-  });
+  page.conversations(0).click();
```
snewcomer: same here

PR #1652, opened 2017-12-24
Add conversation mirage setup and remove un-needed mirage models

snewcomer:
Getting something showing in 4200 land.  

Another note/question is if we want mirage to auto-detect the model, thus removing the need to define a mirage model (got in version 0.3.2 I think).  Lmk if that is something you want me to remove if the mirage model isn't needed.

### DIFF
--- a/config/environment.js
+++ b/config/environment.js
@@ -173,7 +173,6 @@ module.exports = function(environment) {
     ENV['simple-auth'] = {
       store: 'simple-auth-session-store:ephemeral'
     };
-    ENV.SOCKETS_BASE_URL = '';
   }
 
   if (environment === 'staging') {
--- a/config/environment.js
+++ b/config/environment.js
@@ -173,6 +173,7 @@ module.exports = function(environment) {
     ENV['simple-auth'] = {
       store: 'simple-auth-session-store:ephemeral'
     };
+    ENV.SOCKETS_BASE_URL = '';
   }
 
   if (environment === 'staging') {
--- a/mirage/factories/conversation-part.js
+++ b/mirage/factories/conversation-part.js
@@ -13,6 +13,4 @@ export default Factory.extend({
   updatedAt(i) {
     return moment().subtract(i, 'days');
   }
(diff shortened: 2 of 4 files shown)


config/environment.js
```diff
@@ -173,6 +173,7 @@ module.exports = function(environment) {
     ENV['simple-auth'] = {
       store: 'simple-auth-session-store:ephemeral'
     };
+    ENV.SOCKETS_BASE_URL = '';
```
joshsmith: Could we add this between API/WEB?

mirage/factories/conversation-part.js
```diff
@@ -0,0 +1,18 @@
...
+    return moment().subtract(i, 'days');
+  },
+  readAt(i) {
+    return moment().subtract(i, 'days');
+  },
+  updatedAt(i) {
+    return moment().subtract(i, 'days');
+  }
+  // author: belongsTo('user', { async: true }),
+  // conversation: belongsTo('conversation', { async: true })
```
joshsmith: You can remove these two lines.

mirage/models/conversation.js
```diff
@@ -1,7 +0,0 @@
-import { Model, belongsTo, hasMany } from 'ember-cli-mirage';
-
-export default Model.extend({
-  message: belongsTo(),
-  user: belongsTo(),
-  conversationParts: hasMany()
-});
```
joshsmith: Why is this being removed?

snewcomer: https://github.com/samselikoff/ember-cli-mirage/pull/1056

So mirage can auto generate mirage models based off of the ember-data model, avoiding mismatches b/w the two.

Lmk what you think?  We can keep slowly removing these as needed or all of them in a separate PR.

joshsmith: Ah okay thanks. Maybe open a separate issue so others are tracking that’s the case as a point of reference?

joshsmith:
@snewcomer want to squash and merge when you're ready?

PR #1657, opened 2017-12-26
add project to conversation

snewcomer:
This will send up the project_id to the API in order to help with the below issue on the api side.

## References
https://github.com/code-corps/code-corps-api/issues/1313

### DIFF
--- a/app/models/conversation.js
+++ b/app/models/conversation.js
@@ -10,5 +10,6 @@ export default Model.extend({
 
   conversationParts: hasMany('conversation-part', { async: true }),
   message: belongsTo('message', { async: true }),
+  project: belongsTo('project', { async: true }),
   user: belongsTo('user', { async: true })
 });
--- a/tests/unit/models/conversation-test.js
+++ b/tests/unit/models/conversation-test.js
@@ -17,4 +17,5 @@ test('it exists', function(assert) {
 
 testForAttributes('conversation', ['insertedAt', 'readAt', 'status', 'updatedAt']);
 testForBelongsTo('conversation', 'message');
+testForBelongsTo('conversation', 'project');
 testForBelongsTo('conversation', 'user');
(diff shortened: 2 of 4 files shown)


app/models/conversation.js
```diff
@@ -10,5 +10,6 @@ export default Model.extend({
 
   conversationParts: hasMany('conversation-part', { async: true }),
   message: belongsTo('message', { async: true }),
-  user: belongsTo('user', { async: true })
+  user: belongsTo('user', { async: true }),
+  project: belongsTo('project', { async: true })
```
snewcomer: @joshsmith since a project already exists on the `message`, I wasn't sure if there was another way to pass up the project_id to the backend than this.

joshsmith: I would just make sure to alpha order here.

joshsmith: I would just make sure to alpha order here.

tests/unit/models/conversation-test.js
```diff
@@ -18,3 +18,4 @@ test('it exists', function(assert) {
 testForAttributes('conversation', ['insertedAt', 'readAt', 'status', 'updatedAt']);
 testForBelongsTo('conversation', 'message');
 testForBelongsTo('conversation', 'user');
+testForBelongsTo('conversation', 'project');
```
joshsmith: Also alpha here.

snewcomer:
@joshsmith I think this PR is g2g...

joshsmith:
It is, save a rebase, which is why I approved and marked as `ready to merge`.

joshsmith:
Wait, we don't have `project` on the API yet, though, do we?

snewcomer:
Nope.  I thought this was the precursor to this PR I believe - https://github.com/code-corps/code-corps-api/pull/1343

PR #1660, opened 2017-12-27
Remove wait helpers and replace with async/await where needed

joshsmith:
# What's in this PR?

Removes `wait` helpers and replaces with `async`/`await` where they were needed. Should help resolve an issue I've seen where `window.wait` was not available as tests were running in a background tab in Chrome.

### DIFF
--- a/app/components/signup-email-input.js
+++ b/app/components/signup-email-input.js
@@ -2,7 +2,14 @@ import Component from '@ember/component';
 import { not, empty, and, alias } from '@ember/object/computed';
 import { inject as service } from '@ember/service';
 import { once, debounce, cancel } from '@ember/runloop';
-import { set, observer, get, computed } from '@ember/object';
+import { set, get, computed } from '@ember/object';
+import Ember from 'ember';
+
+const {
+  testing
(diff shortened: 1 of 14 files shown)


app/components/signup-email-input.js
```diff
@@ -2,7 +2,14 @@ import Component from '@ember/component';
...
 import { inject as service } from '@ember/service';
 import { once, debounce, cancel } from '@ember/runloop';
-import { set, observer, get, computed } from '@ember/object';
+import { set, get, computed } from '@ember/object';
+import Ember from 'ember';
+
+const {
+  testing
+} = Ember;
+
```
snewcomer: One thought is to make this a variable in the `environment.js` file.  DEBOUNCE_EMAIL_TIMER = 0 in `testing` and = 500 in all other cases.  Then just import `DEBOUNCE_EMAIL_TIMER`

joshsmith: Sounds like a good thing to add later for sure. For now, I'm not waiting on another 7 mins to build this 😛

PR #1661, opened 2017-12-27
Sort converation parts by insertedAt:asc

snewcomer:
Fixes sorting issues with conversation parts.  Pairs well with: https://github.com/code-corps/code-corps-api/pull/1349


joshsmith:
Good catch on this!

PR #1665, opened 2017-12-28
Upgrade mirage with always include linkages flag

snewcomer:
Still figuring a few things out.

Fixes #1461


snewcomer:
The only problem is this PR now.  

https://github.com/samselikoff/ember-cli-mirage/pull/1078

joshsmith:
@snewcomer is there more progress that we need to make here? I'm not sure where to go from here.

PR #1669, opened 2017-12-29
Add status filtering to GitHub events

joshsmith:
WIP.


joshsmith:
I would really rather not add a ton of acceptance tests here. The downside of the added test run time is not really worth any guarantees seen by very few users.

begedin:
Added the missing assertions for the clear button

joshsmith:
I don’t think we’re terribly disagreed on acceptance tests, so curious what you think of this post: https://dockyard.com/blog/2016/03/28/prefer-integration-tests

PR #1673, opened 2018-01-01
Improve footer layout

joshsmith:
# What's in this PR?

Improves the footer layout by having a reduced version on most pages, and then another on marketing-oriented pages.

Introduces the `MarketingRouteMixin` to accomplish some of this, with the thought that if there are other things we'd want specifically on those types of pages, then we would explicitly add that behavior.


app/mixins/marketing-route-mixin.js
```diff
@@ -0,0 +1,19 @@
...
+  siteFooter: service(),
+
+  actions: {
+    didTransition() {
+      get(this, 'siteFooter').enlarge();
+      return true;
+    },
+
+    willTransition() {
+      get(this, 'siteFooter').reduce();
```
begedin: Just on the basis of `Array.prototype.reduce()` being a thing, how would you feel about renaming this to `.shrink()`?

joshsmith: I would agree. Can you make this change and merge?

PR #1675, opened 2018-01-01
Fix project users to render three full rows

joshsmith:
# What's in this PR?

Fix project users to render three full rows of users.


begedin:
This is a simple enough change, so I'll just merge to reduce clutter

PR #1686, opened 2018-01-10
Fix sign up form validation

### LINKED ISSUE #1685: Sign up form validation breaks on autocompletion
# Problem

- Go to `/signup`
- Type the first letter of an email you know will autocomplete.
- The error renders
- Click (don't hit enter) on that email
- The error shows

joshsmith:
# What's in this PR?

Removes `key-down` actions in signup inputs and replaces them with observers. Not ideal, but fixes the bug with minimal effort.

This is a hard one to test, so no regression test for it.

## References
Fixes #1685

### DIFF
--- a/app/components/signup-email-input.js
+++ b/app/components/signup-email-input.js
@@ -66,14 +66,12 @@ export default Component.extend({
     });
   },
 
-  actions: {
-    keyDown() {
-      once(this, '_check');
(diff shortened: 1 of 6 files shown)


joshsmith:
There are tests, but there no regression tests to satisfy the bug's conditions (since it's so specific to the browser and not standard DOM).

PR #1690, opened 2018-01-11
Ensure conversation-thread scrolls to bottom after socket update close #1637

### LINKED ISSUE #1637: Scroll to bottom after new conversation part loaded if scrolled to bottom
# Problem

If a new conversation part is loaded in the background due to a websocket update _and_ you're already scrolled to the bottom, we should scroll to the bottom of the conversation thread again to keep the messages view fully scrolled for you.

snewcomer:
Fixes #1637

### DIFF
--- a/app/models/conversation.js
+++ b/app/models/conversation.js
@@ -1,5 +1,6 @@
 import Model from 'ember-data/model';
 import attr from 'ember-data/attr';
+import { sort } from '@ember/object/computed';
 import { belongsTo, hasMany } from 'ember-data/relationships';
 
 export default Model.extend({
@@ -11,5 +12,8 @@ export default Model.extend({
   conversationParts: hasMany('conversation-part', { async: true }),
   message: belongsTo('message', { async: true }),
   project: belongsTo('project', { async: true }),
-  user: belongsTo('user', { async: true })
+  user: belongsTo('user', { async: true }),
+
+  partsAsc: ['insertedAt:asc'],
+  sortedConversationParts: sort('conversationParts', 'partsAsc')
 });
--- a/app/templates/conversations/conversation.hbs
+++ b/app/templates/conversations/conversation.hbs
@@ -1,6 +1,6 @@
 {{conversations/conversation-thread
   conversation=conversation
-  conversationParts=conversation.sortedConversationParts
+  sortedConversationParts=conversation.sortedConversationParts
   send=(action send)
 }}
 {{conversations/project-details
--- a/app/templates/conversations/conversation.hbs
+++ b/app/templates/conversations/conversation.hbs
@@ -1,5 +1,6 @@
 {{conversations/conversation-thread
   conversation=conversation
+  conversationParts=conversation.sortedConversationParts
   send=(action send)
 }}
 {{conversations/project-details
--- a/tests/integration/components/conversations/conversation-thread-test.js
+++ b/tests/integration/components/conversations/conversation-thread-test.js
@@ -10,6 +10,7 @@ let page = PageObject.create(component);
 function renderPage() {
   page.render(hbs`
     {{conversations/conversation-thread
+      conversationParts=conversation.conversationParts
       conversation=conversation
       send=send
     }}
@@ -60,7 +61,7 @@ test('it delays rendering head until loaded', function(assert) {
   assert.equal(page.conversationParts().count, 0, 'No part is rendered.');
 });
 
-test('it renders each conversation part, sorted by insertedAt', function(assert) {
+test('it renders each conversation part', function(assert) {
   assert.expect(4);
 
   let conversationParts = [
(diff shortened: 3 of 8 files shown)


app/templates/conversations/conversation.hbs
```diff
@@ -1,5 +1,6 @@
 {{conversations/conversation-thread
   conversation=conversation
+  conversationParts=conversation.sortedConversationParts
```
snewcomer: This ensures `didUpdateAttrs` is ran again when conversationParts has been updated.

joshsmith:
This has been a particularly nasty bug, so I'd like to see if there's a way we could add a regression test here. Can you think of a way that we could test this?

Also, why is it that this worked here but the `sort-by` composable helper did not?

app/models/conversation.js
```diff
@@ -10,5 +11,8 @@ export default Model.extend({
 
   conversationParts: hasMany('conversation-part', { async: true }),
   message: belongsTo('message', { async: true }),
-  user: belongsTo('user', { async: true })
+  user: belongsTo('user', { async: true }),
+
+  partsAsc: ['insertedAt:asc'],
+  sortedConversationParts: sort('conversationParts', 'partsAsc')
```
joshsmith: A model test for this would be cheap.

joshsmith:
This breaks the conversation part thread component test.

tests/unit/models/conversation-test.js
```diff
@@ -1,18 +1,27 @@
...
 import { testForAttributes } from 'code-corps-ember/tests/helpers/attributes';
 import { testForBelongsTo } from 'code-corps-ember/tests/helpers/relationship';
+import { run } from '@ember/runloop';
 
 moduleForModel('conversation', 'Unit | Model | conversation', {
-  // Specify the other units that are required for this test.
   needs: [
     'model:message',
+    'model:conversation',
+    'model:conversation-part',
```
joshsmith: Can we alpha-order these?

tests/unit/models/conversation-test.js
```diff
@@ -1,18 +1,27 @@
...
+test('sortedConversationParts sorts insertedAt by asc', function(assert) {
+  let firstPart, secondPart;
+  run(() => {
+    firstPart = this.store().createRecord('conversation-part', { body: 'wat', insertedAt: new Date('2017-10-07') });
+    secondPart = this.store().createRecord('conversation-part', { body: 'foo', insertedAt: new Date('2017-10-06') });
+  });
+  let model = this.subject({ conversationParts: [firstPart, secondPart] });
+  assert.equal(model.get('sortedConversationParts.length'), 2);
+  assert.equal(model.get('sortedConversationParts').get('firstObject').get('body'), 'foo');
+  assert.equal(model.get('sortedConversationParts').get('lastObject').get('body'), 'wat');
```
joshsmith: I would maybe consider renaming `firstPart` to `laterPart` and `secondPart` to `earlierPart`, then in the assertions you can use those variables to test against (rather than the explicit string). I think that could make it more obvious what's under test.

tests/integration/components/conversations/conversation-thread-test.js
```diff
@@ -60,7 +61,7 @@ test('it delays rendering head until loaded', function(assert) {
   assert.equal(page.conversationParts().count, 0, 'No part is rendered.');
 });
 
-test('it renders each conversation part, sorted by insertedAt', function(assert) {
+test('it renders each conversation part', function(assert) {
```
joshsmith: I'm not sure we wanted to change this test, but change `conversationParts` to be `conversation.sortedConversationParts`, no?

joshsmith: In fact, it makes me wonder if we shouldn't pass `conversationParts` in at all, and instead the `conversation-thread` component should take the `conversation` only and make use of its `sortedConversationParts` internally.

snewcomer: The problem is we are using plain objects and not ember data models.  So we don't have access to `sortedConversationParts`.  lmk if you see something different.

Also if we don't pass in `sortedConversationParts`, `didUpdateAttrs` will never fire b/c none of it's direct attrs changed when something comes through a web socket.

tests/unit/models/conversation-test.js
joshsmith: It looks like you did the first part I suggested previously but not the second. Is there a reason for that?

tests/integration/components/conversations/conversation-thread-test.js
joshsmith: You can mock the `conversation` by having a key of `sortedConversationParts`.

But the latter point seems like the key. I'm curious to hear what @begedin thinks about this, as well.

tests/unit/models/conversation-test.js
snewcomer: Ah.  I didn't see that.  

So What we are testing is that the `firstObject` is 'foo'.  I could add a test failure description to those.

app/models/conversation.js
snewcomer: dunzo

tests/integration/components/conversations/conversation-thread-test.js
snewcomer: Yeah lmk what you you would like me to do.  By mocking out sortedConversationParts, we are def closer to the real implementation.  But not actually testing the real implementation.

joshsmith:
So what I would test is this, actually:

```
assert.equal(model.get('sortedConversationParts').get('firstObject').get('body'), earlierPart.get('body'));
assert.equal(model.get('sortedConversationParts').get('lastObject').get('body'), laterPart.get('body'));
```

tests/integration/components/conversations/conversation-thread-test.js
joshsmith: I think in that case, we should pass in `sortedConversationParts` (as we do in the actual uses of this component) and maybe even change the component to use a `sortedConversationParts` (vs naming it just `conversationParts`), that way it's clear that we're always passing in something that needs sorted. Right now this test leads us to think the component is used differently than it actually is _and_ that there isn't any date sorting. We need to clarify that the sorting happens outside the component's scope.

joshsmith:
Squash and merge whenever tests pass. Looks good to go.

PR #1724, opened 2018-06-04
chore(package): update ember-source to version 3.2.0

begedin:
Closes #1699 and a bunch of other tasks I have yet to identify

This PR updates `ember-cli` and `ember-source` to the latest version. With that, comes a bunch of other updates. However, I have underestimated the amount of time needed, so I'm stopping at this point and will complete it in my spare time.

# TODO 

- [ ] Perform an ember-cli-mirage migration from 0.3.x to 0.4.x
- [ ] See why some actions in tests seem to be failing? Did actions change at some point in ember?
- [ ] Double-check if ember-can migration was OK. Seems we were unnecessarily using the mixin (which is now deprecated) in several places.
- [ ] Update any other packages that might be causing issues, but keep it minimal in this PR. Other upgrades should be separate


app/routes/conversations/conversation.js
```diff
@@ -2,9 +2,8 @@ import { get } from '@ember/object';
 import { inject as service } from '@ember/service';
 import Route from '@ember/routing/route';
 import AuthenticatedRouteMixin from 'ember-simple-auth/mixins/authenticated-route-mixin';
-import { CanMixin } from 'ember-can';
 
-export default Route.extend(AuthenticatedRouteMixin, CanMixin, {
```
begedin, line 7: Seems `can` is just a template helper now and we now longer need the mixin for the route, but I'm not 100% sure.

app/routes/organizations/slugged-route/settings.js
```diff
@@ -1,14 +1,16 @@
...
 
-export default Route.extend(AuthenticatedRouteMixin, CanMixin, {
+export default Route.extend(AuthenticatedRouteMixin, {
+  can: service(),
   session: service(),
 
   beforeModel() {
     let organization = this.modelFor('organizations.slugged-route');
-    if (this.cannot('manage organization', organization)) {
+    if (get(this, 'can').cannot('manage organization', organization)) {
```
begedin, line 13: We inject the service now and use it, instead of the mixin

ember-cli-build.js
```diff
@@ -36,6 +36,12 @@ module.exports = function(defaults) {
       // runOnPostBuild: (env === 'development') ? 'development-postbuild' : false,
       // shouldActivate: true
     },
+    sassOptions: {
+      includePaths: [
+        'node_modules/bourbon/core',
+        'node_modules/bourbon-neat/app/assets/stylesheets'
```
begedin, line 42: This is how a sass npm package is added to the app. The files become available for `import` same as they are by using the wrapper addon `ember-cli-bourbon`. The addon has an issue which causes the app to fail on build, and this is straightforward enough to warrant removing the addon completely.

ember-cli-build.js
```diff
@@ -49,18 +55,7 @@ module.exports = function(defaults) {
...
-  // If you need to use different assets in different
-  // environments, specify an object as the first parameter. That
-  // object's keys should be the environment name and the values
-  // should be the asset to use in that environment.
-  //
-  // If the library that you are including contains AMD or ES6
-  // modules that you would like to import into your application
-  // please specify an object with the list of modules as keys
-  // along with the exports of each module as its value.
+  app.import('node_modules/normalize.css/normalize.css');
```
begedin, line 58: This is how css is imported into `vendor.css`. If we have a need for a sass import, we can use `normalize-sass`

package.json
```diff
@@ -107,9 +109,10 @@
     "eslint-plugin-ember-suave": "^1.0.0",
     "loader.js": "^4.2.3",
     "normalize.css": "^7.0.0",
+    "qunit-dom": "^0.6.2",
     "yuidoc-ember-theme": "^1.3.0"
   },
   "engines": {
-    "node": "^4.5 || 6.* || >= 7.*"
+    "node": "6.* || >= 8.*"
```
begedin, line 116: The official version range supported by `ember-cli`

app/styles/app.scss
```diff
@@ -2,11 +2,6 @@
 // VENDORS
 //
 
-// Normalize.css
-// A modern, HTML5-ready alternative to CSS resets
-// https://necolas.github.io/normalize.css/
-@import "normalize";
-
```
begedin, line 9: normalize is imported straight into `vendor.css` from the `normalize.css` vendor package

